PR #145, opened 2015-02-08
[Fix #144] Fix NPE when :session is nil

### LINKED ISSUE #144: Broken tests
@cichli @gtrak Please, take a look [here](https://travis-ci.org/clojure-emacs/cider-nrepl/jobs/49959038). Is this caused by some upstream change in something cljs related?

cichli:


bbatsov:
:+1: Thanks!

PR #149, opened 2015-02-14
Add pprint middleware

cichli:
This adds a `wrap-pprint` middleware so that CIDER's pprinting functionality will also be ClojureScript-compatible. By using `interruptible-eval`'s optional `:eval` slot we can ensure that `clojure.core/pprint` is not evaluated in a CLJS context, and also keep the evaluation inside `interruptible-eval`'s machinery, so the user can interrupt the printing of the result at the REPL.

See also clojure-emacs/cider#972.

### DIFF
--- a/src/cider/nrepl/middleware/pprint.clj
+++ b/src/cider/nrepl/middleware/pprint.clj
@@ -1,0 +1,33 @@
+(ns cider.nrepl.middleware.pprint
+  (:require [clojure.pprint :refer [pprint *print-right-margin*]]
+            [clojure.tools.nrepl.middleware :refer [set-descriptor!]]
+            [cider.nrepl.middleware.util.cljs :as cljs]))
+
+(defn pprint-eval
+  [form]
+  (let [result (eval form)]
+    (pprint result)
+    result))
+
+(defn wrap-pprint-reply
+  [handler {:keys [right-margin session] :as msg}]
+  (when right-margin
+    (swap! session assoc #'*print-right-margin* right-margin))
+  (handler (assoc msg :eval 'cider.nrepl.middleware.pprint/pprint-eval)))
+
+(defn wrap-pprint
+  [handler]
+  "Middleware that adds a pretty printing option to the eval op.
+  Passing a non-nil value in the `:pprint` slot will cause eval to call
+  clojure.pprint/pprint on its result. The `:right-margin` slot can be used to
+  bind `*clojure.pprint/*print-right-margin*` during the evaluation."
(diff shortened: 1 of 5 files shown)


cichli:
I think it would also be useful to add a `pprint` op to this that will use `clojure.tools.reader/read-string` to read the input and then use `clojure.core/pprint` to print the result - the idea being that this could be used by the client as the EDN counterpart to `format-code`. Should I fold that into this or make a separate PR?

bbatsov:
> This adds a wrap-pprint middleware so that CIDER's pprinting functionality will also be ClojureScript-compatible. By using interruptible-eval's optional :eval slot we can ensure that clojure.core/pprint is not evaluated in a CLJS context, and also keep the evaluation inside interruptible-eval's machinery, so the user can interrupt the printing of the result at the REPL.

Nice!

> I think it would also be useful to add a pprint op to this that will use clojure.tools.reader/read-string (we already depend on this via rewrite-clj) to read the input and then use clojure.core/pprint to print the result - the idea being that this could be used by the client as the EDN counterpart to format-code. Should I fold that into this or make a separate PR?

Yes, it makes a lot of sense to leverage `tools.reader` more. I'd suggest doing so in a separate PR indeed.

Add a few tests and we're good to merge.

src/cider/nrepl/middleware/pprint.clj
```diff
@@ -0,0 +1,30 @@
...
+  (comp pprint eval))
+
+(defn wrap-pprint-reply
+  [h {:keys [right-margin session] :as msg}]
+  (when right-margin
+    (swap! session assoc #'*print-right-margin* right-margin))
+  (h (assoc msg :eval 'cider.nrepl.middleware.pprint/pprint-eval)))
+
+(defn wrap-pprint
+  [h]
```
bbatsov: I'd suggest naming this `handler` for clarity's sake. We know what `h` stands for here, but others might be confused.

cichli:
Done. Noticed a slight deviation in behaviour from the way CIDER does pprinting - I was returning the return value of `pprint` rather than `eval` in the `:value` slot. Not sure we were ever actually using the `:value` when pretty printing but it's worth having it there. Fixed w/ test added.

I added the `send-message` fn in the pprint test file itself but it's worth moving out to a generic test utils place if we ever add other middlewares that need this kind of integration testing.

bbatsov:
> I added the send-message fn in the pprint test file itself but it's worth moving out to a generic test utils place if we ever add other middlewares that need this kind of integration testing.

Indeed.

:+1: Everything looks good. Thanks!

PR #150, opened 2015-02-15
Fix as-sym to not throw

cichli:
It seems like `[]` was being passed in the `:ns` slot from the backtrace posted in clojure-emacs/cider#879. No idea why the client was doing that, but we shouldn't throw in that case.

The only valid types to pass to `symbol` are strings and symbols - since it either returns the symbol you passed, or calls `clojure.lang.Symbol/intern`, which takes a string.

### DIFF
--- a/src/cider/nrepl/middleware/util/misc.clj
+++ b/src/cider/nrepl/middleware/util/misc.clj
@@ -17,7 +17,8 @@
 
 (defn as-sym
   [x]
-  (if x (symbol x)))
+  (when (or (symbol? x) (string? x))
+    (symbol x)))
 
 (defmulti transform-value "Transform a value for output" type)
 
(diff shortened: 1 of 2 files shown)


cichli:
Seems the Travis failures are caused by the cljfmt version bump.

bbatsov:
> Seems the Travis failures are caused by the cljfmt version bump.

Yep. I reported an issue upstream - https://github.com/weavejester/cljfmt/issues/28

bbatsov:
P.S. Nice catch. You might want to add a couple of tests for this.

cichli:
Done.

cichli:
@bbatsov ping :)

bbatsov:
:+1: Thanks! Sorry about the delay - for some reason I thought I had actually merged this. :-)

cichli:
:+1: No worries!

PR #153, opened 2015-02-20
Handle classpath-related errors in info middleware

cichli:
Not sure the best way to add tests for these without adding `funnyqt` to the test profile deps (since I'm not sure what it's doing to the classpath, and how else to replicate it). Will go ahead with that if you think it's needed.


bbatsov:
:+1: I think this will be enough. Thanks!

PR #154, opened 2015-02-22
Fix misplaced docstring

cichli:
Whoops :blush:.

Found using Eastwood btw, I'm going through the [rest of the errors](https://gist.github.com/cichli/13bccd2cb3fcd89b0124) now. Might be useful to add this to Travis once they're all fixed so we don't introduce any more?

### DIFF
--- a/src/cider/nrepl/middleware/pprint.clj
+++ b/src/cider/nrepl/middleware/pprint.clj
@@ -16,11 +16,11 @@
   (handler (assoc msg :eval 'cider.nrepl.middleware.pprint/pprint-eval)))
 
 (defn wrap-pprint
-  [handler]
   "Middleware that adds a pretty printing option to the eval op.
   Passing a non-nil value in the `:pprint` slot will cause eval to call
   clojure.pprint/pprint on its result. The `:right-margin` slot can be used to
   bind `*clojure.pprint/*print-right-margin*` during the evaluation."
+  [handler]
   (fn [{:keys [op pprint right-margin] :as msg}]
     (if (and pprint (= op "eval"))
       (wrap-pprint-reply handler msg)


bbatsov:
Guess you didn't notice the different colours. :-)

I run eastwood and kibit infrequently. Don't think we can add them to TravisCI as some of their warnings are clearly false positives. On the other hand you might know something on the subject that I don't. :-)

cichli:
Yeah, seems the `constant-test` linter has some issues with quoted collection literals - will investigate and report an issue upstream if necessary.

P.S. which different colours? It all looks the same to me; I might have some config clobbering it.

bbatsov:
Not sure what colour theme you're using, but strings and docstring are font-locked differently.

image

cichli:
Solarized (seems the issue occurs with both dark and light variants).

cichli:
Must be something in my config preventing the `:slant` from being applied [here](https://github.com/bbatsov/solarized-emacs/blob/674520d15f98ffc316e56a7df9e7890cdc823037/solarized.el#L454).

cichli:
Ah, this will trigger a `constant-test` warning:

```clojure
(is (= (:status val) #{:done}))
```

but this will not:

```clojure
(is (= #{:done} (:status val)))
```

We can configure Eastwood to suppress the warning specifically in the expansion of `is` (see [here](https://github.com/jonase/eastwood#eastwood-config-files)) but I'm not sure if this would possibly suppress any genuine warnings.

Maybe it makes sense to prefer the second form for style/readability/consistency reasons anyway though. For example when testing that multiple expressions evaluate to the same constant value, like [here](https://github.com/gtrak/cljs-tooling/blob/master/test/cljs_tooling/test_complete.clj#L20), I think makes things more readable. Perhaps a section on testing in the style guide would be useful? :-)

bbatsov:
Regarding solarized - I think the current config is suboptimal and you should probably file a ticket regarding this. Using the same face for two different semantic categories is bad with or without the slant. I don't use it much these days and I've totally missed this.

PR #156, opened 2015-02-22
Fix a redef'd var in info tests

cichli:
`test-info` is already defined earlier in the file, which triggers a warning in Eastwood. We could also just extract the eldoc tests into their own file.

### DIFF
--- a/test/cider/nrepl/middleware/info_test.clj
+++ b/test/cider/nrepl/middleware/info_test.clj
@@ -104,21 +104,21 @@
             "resource" "clojure/core.clj"}))))
 
 ;;;; eldoc
-(def test-info {:arglists '([x] [x y])})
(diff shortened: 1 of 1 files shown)


test/cider/nrepl/middleware/info_test.clj
```diff
@@ -104,21 +104,21 @@
             "resource" "clojure/core.clj"}))))
 
 ;;;; eldoc
-(def test-info {:arglists '([x] [x y])})
+(def example-info {:arglists '([x] [x y])})
```
bbatsov: Maybe `test-eldoc-info`?

cichli: Done.

bbatsov:
:+1:

PR #157, opened 2015-02-22
Prefer :require and :refer :all to :use

cichli:
As per Eastwood warnings and the style guide :-).


bbatsov:
:+1: I always meant to change this, but was too lazy to do the extra typing. :-)

PR #158, opened 2015-02-22
Fix an eastwood warning

cichli:
Because `chars` is the name of a var in core, binding something to the name `chars` and then calling it directly as a function triggers an Eastwood warning. We could also rename `chars` to `char-map` or somesuch.

### DIFF
--- a/src/cider/nrepl/middleware/util/java/parser.clj
+++ b/src/cider/nrepl/middleware/util/java/parser.clj
@@ -89,29 +89,29 @@
 ;; Use GFM extensions for multiline code blocks and tables.
 (def markdown
   "Syntax map from html tag to a tuple of tag type key, start, and end chars"
-  (let [chars {:p     ["\n\n"]     :code  ["`" "`"]
-               :br    ["\n"]       :code* ["\n\n```\n" "```\n\n"]
-               :em    ["*" "*"]    :table ["\n|--" "\n|--"]
-               :str   ["**" "**"]  :thead ["" "|--\n"]
-               :list  ["\n"]       :tr    ["\n" "|"]
-               :li    ["- "]       :td    ["|"]
-               :dd    [": "]       :th    ["|"]}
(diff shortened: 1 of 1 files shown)


bbatsov:
Guess we should rename the local to avoid the shadowing altogether.

cichli:
Done.

src/cider/nrepl/middleware/util/java/parser.clj
```diff
@@ -89,27 +89,27 @@
...
+                  HTML$Tag/OL :list        HTML$Tag/TR    :tr
+                  HTML$Tag/DL :list        HTML$Tag/TD    :td
+                  HTML$Tag/LI :li          HTML$Tag/TH    :th
+                  HTML$Tag/DT :li
+                  HTML$Tag/DD :dd}]
     (-> (reduce (fn [tags [tag k]]
-                  (assoc tags tag (cons k (chars k))))
+                  (assoc tags tag (cons k (char-map k))))
                 {} tags)
         (with-meta chars))))
```
bbatsov: You missed it here.

cichli: :blush:

bbatsov:
Hmm, the build is broken in some pretty strange way. Not sure if this isn't some travis problem...

cichli:
Force pushed to test again and a different build in the matrix failed in the same way this time! Maybe we should cache `.m2` to reduce the chance of this kind of error when downloading JARs (see http://docs.travis-ci.com/user/caching/#Arbitrary-directories).

bbatsov:
> Force pushed to test again and a different build in the matrix failed in the same way this time! Maybe we should cache .m2 to reduce the chance of this kind of error when downloading JARs (see http://docs.travis-ci.com/user/caching/#Arbitrary-directories).

Might be a good idea. Haven't really thought about this until you brought it up.

PR #160, opened 2015-02-27
Update complete op to return annotated candidates

cichli:
Required by clojure-emacs/cider#993.

The 1.7.0-SNAPSHOT builds still seem to be unavailable but I can confirm locally that it builds and tests pass against 1.7.0-alpha5.


bbatsov:
:+1:

PR #161, opened 2015-02-28
Bump cljs-tooling dep to 0.1.5-SNAPSHOT

cichli:
This fixes cljs completion on 0.9.0-SNAPSHOT / latest CIDER. I still haven't sorted the Travis issue but received confirmation on the mailing list that it should be working - I think we should just use alpha5 instead of master until I get this figured out.


bbatsov:
:+1: I'm fine with using the alpha for now.

PR #162, opened 2015-03-01
Fix formatting and run `lein cljfmt check` in Travis builds

cichli:
Guess we can add other lint tools in the same way if you're interested.

btw, I've not heard back from the Eastwood maintainers regarding the erroneous warnings being raised in the tests, but we can work around it by tweaking the tests in question slightly. We just need to ensure that constant expressions are the first argument to `=` in `is` forms, for example:

```clojure
(is (= (:javadoc (class-info 'java.lang.String))
       "java/lang/String.html"))
```

becomes

```clojure
(is (= "java/lang/String.html"
       (:javadoc (class-info 'java.lang.String))))
```

Would you be fine with such a PR?


bbatsov:
:+1: 



> btw, I've not heard back from the Eastwood maintainers regarding the erroneous warnings being raised in the tests, but we can work around it by tweaking the tests in question slightly.

What was the exact issue? They want tests to follow a "expected" followed by "actual" pattern or something else?

cichli:
A form like this:

```clojure
(is (= (:status val) #{:done}))
```

gets expanded into something like this:

```clojure
(if
 temp__4090__auto__
 (clojure.core/let
  [more__35__auto__ temp__4090__auto__]
  (clojure.core/let
   [result__36__auto__
    (clojure.core/apply clojure.core/= a__34__auto__ more__35__auto__)]
   (if
    result__36__auto__
    (clojure.test/do-report
     {:actual more__35__auto__,
      :type :pass,
      :message nil,
      :expected a__34__auto__})
    (clojure.test/do-report
     {:diffs
      (clojure.core/map
       clojure.core/vector
       more__35__auto__
       (clojure.core/map
        (fn*
         [p1__33__37__auto__]
         (clojure.core/take
          2
          (clojure.data/diff a__34__auto__ p1__33__37__auto__)))
        more__35__auto__)),
      :actual more__35__auto__,
      :type :fail,
      :message nil,
      :expected a__34__auto__}))
   result__36__auto__))
 (throw (java.lang.Exception. "= expects more than one argument")))
```

where `temp__4090__auto__` is equal to `(clojure.core/seq (clojure.core/list #{:done}))`, so the outermost `if` always goes down the `then` branch, and Eastwood warns:

`test/cider/nrepl/middleware/macroexpand_test.clj:105:7: constant-test: Test expression is always logical true or always logical false`

I've actually looked into this a bit more and it's easy to fix by configuring Eastwood. We just need a config file containing:

```clojure
(disable-warning
 {:linter :constant-test
  :if-inside-macroexpansion-of #{'clojure.test/is}
  :within-depth 1
  :reason "The `is` macro commonly expands to contain an `if` with a condition that is a constant."})
```

This will cause Eastwood to ignore the above case but still warn for something like:

```clojure`
(is (= (if true 1 2) 1))
```

I'll open a PR.

cichli:
I've proposed making this a default in Eastwood in jonase/eastwood#154.

PR #164, opened 2015-03-07
Portable inspector

cichli:
see clojure-emacs/cider#1009

* The magic around inspecting vars is now gone. It's possible to inspect either `map` or `#'map` depending on if you want to inspect the function itself or the var.

* Inspector state now lives in the session bindings map rather than a global atom.

* I've removed the reset op for now as the client was not using it. Easy to add it back in, which I will do in a separate PR when I add support to the client.

* We no longer serialise the result, as `pr-values` is now doing it for us.


bbatsov:
:+1: Great work! You should mention this in cider's changelog.

PR #165, opened 2015-03-07
Use default handlers in session fixture

cichli:
Both `wrap-inspect` and `wrap-pprint` rely on the eval op, and I think it would be nice to have the tests catch any unexpected interactions between them (and likewise for any other handlers, existing or future).


bbatsov:
:+1:

Issue #13, opened 2014-02-20
Implement reloading middleware


bbatsov:
I'm thinking of something pretty basic in the lines of `clojure.tools.namespace.repl/refresh`. More on the subject is available [here](clojure.tools.namespace.repl/refresh).

Issue #103, opened 2014-08-03
Support xref or caller functionality


dgtized:
I took a brief look at the old swank code, and found this: https://github.com/technomancy/swank-clojure/blob/master/src/swank/commands/xref.clj. If I read correctly it's naively grepping the source of all functions to see if they call a particular function. Should we revive support for this in cider, and can we do better than simply grepping?

bbatsov:
Closing in favour of merging @benedekfazekas & @expez's work instead.

PR #166, opened 2015-03-07
Bump tools.namespace dep to 0.2.10

cichli:
Changelog is [here](https://github.com/clojure/tools.namespace#change-log).

AFAICT this is only used by inlined code in the client, but moving to middleware would be useful. A potential `refresh` op could have the optional args:

* `before` - fully-qualified name of a function to run before refreshing
* `after` - fully-qualified name of a function to pass as the `:after` option to `refresh`

edit: ah just noticed #13 :-).

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -15,7 +15,7 @@
                  [org.tcrawley/dynapath "0.2.3"]
                  [org.clojure/tools.nrepl "0.2.7"]
                  [org.clojure/java.classpath "0.2.0"]
-                 [org.clojure/tools.namespace "0.2.5"]
+                 [org.clojure/tools.namespace "0.2.10"]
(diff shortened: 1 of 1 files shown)


bbatsov:
> AFAICT this is only used by inlined code in the client, but moving to middleware would be useful. A potential refresh op could have the optional args:
> 
> before - fully-qualified name of a function to run before refreshing
> after - fully-qualified name of a function to pass as the :after option to refresh

Feel free to implement it. :-) As you noticed this has been on my backburner for quite a while (although I imagine it will be pretty easy to implement this middleware).

bbatsov:
#103 is the other big missing middleware that should not be very hard to implement (at least as described there). There's a smarter version of this in `refactor-nrepl` but we cannot use it until we figure out the dependency isolation issue...

PR #168, opened 2015-03-09
Make undef test idempotent and add extra test cases

cichli:
This makes the original test not fail on repeated invocations when using [lein-test-refresh](https://github.com/jakemcc/lein-test-refresh) and adds a few other cases.


bbatsov:
:+1:

PR #175, opened 2015-03-24
Piggieback 0.2.x support

cichli:
**Please do not merge yet.**

This PR is for adding support to the [0.2.0-SNAPSHOT branch](https://github.com/cemerick/piggieback/compare/new-cljs-repl) of piggieback, in a backwards-compatible way.

- [x] The compiler env is now in `*cljs-compiler-env*`, rather than assoc'd into `*cljs-repl-env*`.
- [x] It's not delegating to the `eval` op in the same way anymore, so the pprint middleware needs updating to not depend on this.
- [x] The `:value` slot used to be the result of calling `read-string`, but it now just contains the (generally readable) printed output from `cljs.repl/repl*` for the eval. We have to do the reading ourselves now in the pprint and inspector middlewares.
- [x] `:eval-error` is now returned when an exception is thrown, but the stacktrace middleware relies on `*e` being set in the session to work - so the error buffer just displays the last thrown Clojure exception when an exception is thrown at the ClojureScript REPL.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -14,7 +14,7 @@
                  [debugger "0.1.6"]
                  [cljfmt "0.1.10"]
                  [org.tcrawley/dynapath "0.2.3"]
-                 [org.clojure/tools.nrepl "0.2.7"]
+                 [org.clojure/tools.nrepl "0.2.10"]
                  [org.clojure/java.classpath "0.2.0"]
                  [org.clojure/tools.namespace "0.2.10"]
                  [org.clojure/tools.trace "0.7.8"]
--- a/src/cider/nrepl/middleware/complete.clj
+++ b/src/cider/nrepl/middleware/complete.clj
@@ -50,7 +50,7 @@
 
 (set-descriptor!
  #'wrap-complete
- (cljs/maybe-piggieback
+ (cljs/requires-piggieback
   {:handles
    {"complete"
     {:doc "Return a list of symbols matching the specified (partial) symbol."
--- a/src/cider/nrepl/middleware/info.clj
+++ b/src/cider/nrepl/middleware/info.clj
@@ -266,7 +266,7 @@
 
 (set-descriptor!
  #'wrap-info
- (cljs/maybe-piggieback
+ (cljs/requires-piggieback
   {:handles
    {"info"
     {:doc "Return a map of information about the specified symbol."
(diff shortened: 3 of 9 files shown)


bbatsov:
> piggieback now requires nREPL 0.2.10 to work properly, maybe we should have the plugin automatically add this dep to the :dev profile?

The less work users have to do, the better. Ideally one day CIDER will properly modify profiles.clj is something is missing or outdated. That'd be a huge usability win IMO.

cichli:
Getting closer but I still need to go back and re-test with 0.1.5 and a suitable ClojureScript version. The backwards-compatibility stuff isn't too much of a burden, not sure how long it will need to stick around considering the moving target that ClojureScript is (hopefully not for much longer).

For now we just let piggieback print stacktraces to the REPL - we'll need upstream changes and possibly some new stuff added to cljs-tooling for proper stacktrace support.

cichli:
Okay, piggieback 0.2.0 is out now. I'll have time to test all the common CLJS REPL implementations tomorrow with both that and 0.1.5. Then I'll fix any issues and probably just squash the lot into one commit if you're happy with that.

I'll do the nREPL dep injection in a separate PR if you want?

cichli:
Okay, this is good to merge now. I have:

* Isolated all the piggieback backwards-compatibility stuff into the `util.cljs` namespace
* Tested with both piggieback 0.1.5 and 0.2.0 for browser/figwheel/nashorn/node/rhino/weasel REPLs, ensuring that normal eval, printing, pretty printing, stacktraces and the inspector all behave well
* Fixed an issue in the `pprint` middleware (if the `:value` slot is a string we now use `println` so control characters aren't escaped, see screenshot below)

screenshot 2015-04-03 13 07 26

Enhancements to come that I'll log issues for first, as they may warrant some discussion:

* Have the plugin inject the latest nREPL into the `:dev` profile automatically
* Check piggieback version along with other version checks at REPL connect, and if piggieback is present but not 0.2.x, display a warning
* New test profile with tests for piggieback 0.2.0

bbatsov:
:+1: 

> Have the plugin inject the latest nREPL into the :dev profile automatically
> Check piggieback version along with other version checks at REPL connect, and if piggieback is present but not 0.2.x, display a warning
> New test profile with tests for piggieback 0.2.0

Looking forward to those tickets. Mention the piggieback changes in CIDER's changelog.

PR #181, opened 2015-04-05
[cider#1052] Don't debug synchronous code

Malabarba:
Relies on https://github.com/clojure-emacs/cider/pull/1070

### DIFF
--- a/src/cider/nrepl/middleware/debug.clj
+++ b/src/cider/nrepl/middleware/debug.clj
@@ -3,6 +3,8 @@
             [clojure.tools.nrepl.middleware :refer [set-descriptor!]]
             [clojure.tools.nrepl.misc :refer [response-for]]
             [cider.nrepl.middleware.util.instrument :refer [instrument]]
+            [debugger.time :as t]
+            [debugger.config :as c]
             [debugger.core :refer [break]]))
 
 ;;;; ## Interaction with the client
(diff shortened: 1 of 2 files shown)


src/cider/nrepl/middleware/debug.clj
```diff
@@ -28,18 +28,22 @@
...
-   (response-for @debugger-message
-                 (assoc extras
-                        :debug-value (pr-str value)
-                        :breakfunction nil)))
-  ;; Send the actual break.
-  (binding [*out* (new java.io.StringWriter)]
-    (break value)))
+  ;; Don't debug code being run synchronously. It is probably
+  ;; unintended and will unexpectedly ask for input.
+  (if (:sync clojure.tools.nrepl.middleware.interruptible-eval/*msg*)
```
bbatsov: I wonder if this will always behave as we expect it to. @cichli does it look legit to you?

Malabarba: Perhaps I should go with a more verbose name for the key, like `:cider-sync`?

Malabarba: Ah, I think I get your question now. I suppose there's nothing preventing other middleware from `eval`ing code as well.
In fact, I think pprint might be one.

cichli: It seems `breakpoint` is only ever called within the context of an nREPL eval, so `*msg*` should always be bound here. If we want to call this function from other middleware handlers we could just bind `*msg*` to the message being handled, assuming we include the `:sync` key for all requests (i.e. not just `eval` requests).

Not been following the issue too closely, so forgive my ignorance, but shouldn't this really be fixed client-side (i.e. properly handle `need-input` during synchronous requests) so we don't need this kind of hack?

Malabarba: > Not been following the issue too closely, so forgive my ignorance, but shouldn't this really be fixed client-side (i.e. properly handle need-input during synchronous requests) so we don't need this kind of hack?

Yes, this should be fixed client-side, but I'm under the impression that even with the handle-input fix in sync-request the breakpoint will still bug out during a sync eval (because its workings are more complicated than a simple need-input event). Still, I haven't had a chance to test that conjecture yet (busy busy busy :-P).

PR #186, opened 2015-04-08
Bump tools.reader dep to 0.8.16

cichli:


bbatsov:
15 minutes build time. :-)

cichli:
That's the total for all jobs - five run concurrently at any one time so it's around 3 minutes actual :-).

PR #187, opened 2015-04-10
Don't send debugger message if the break won't prompt

Malabarba:
While debugging something else I noticed a bug in the debugger.

If the user hits `c` to continue without stopping, clj-debugger's `break` statements don't trigger anymore (as expected) but we were still sending the debugger message for cider to "prepare" for a break.

This fixes that.


Malabarba:
Connected to https://github.com/razum2um/clj-debugger/issues/12

bbatsov:
:+1:

PR #190, opened 2015-04-11
Test against latest ClojureScript version

cichli:
CLJS is still a somewhat fast moving target, I think for now we should test against latest rather than relying on the version piggieback provides. Please merge after #189, or the tests will not pass :-).


cichli:
Pushed again for Travis.

bbatsov:
:+1:

PR #183, opened 2015-04-08
[Fix #143] Inline dependencies

benedekfazekas:
- use mranderson to inline dependencies at package time
- simple build script to help working with mranderson
- bit of explanation in readme
- travis file modified accordingly

tests pass, not yet tested the built middleware tho (will do) but ready for a review I suppose.

PR #192, opened 2015-04-14
Update 'With Immutant/WildFly' section

tobias:
This fixes the section about working with a container that uses JBoss
modules - the issue actually *doesn't* affect Immutant users.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,14 +66,19 @@ server with CIDER's own nREPL handler.
   (nrepl-server/start-server :port 7888 :handler cider-nrepl-handler))
 ```
 
-### With Immutant
-Using the advanced features of the `info` middleware with WildFly/Immutant
-requires a tweek, since JBoss modules prevent modifications to `AppClassLoader`
-(usually the highest modifiable classloader) from being seen by application
-code. To work around this, run the following code from within your
-WildFly/Immutant container to mark that classloader as unmodifiable, and cause
-the lower level `clojure.lang.DynamicClassLoader` to be used instead. This code
-must execute prior to loading the `cider-nrepl` middleware.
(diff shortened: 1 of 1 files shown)


bbatsov:
Btw, will this still be an issue after #183?

tobias:
That's a good question. `cider.nrepl.middleware.util.java/add-classpath!` will still try to modify the `AppClassLoader` even with #183, but the code the user needs to use in-container needs to change. Since mranderson inlines dynapath, the require statement needs to change to:

    (require '[deps.dynapath.v0v2v3.dynapath.dynamic-classpath :as cp])

since the the standard dynapath dep won't be available, and even if it is, we have to use the same one that cider-nrepl is using, since `deps.dynapath.v0v2v3.dynapath.dynamic-classpath.DynamicClasspath` and `dynapath.dynamic-classpath.DynamicClasspath` will be considered different protocols. In fact, this protocol difference will break Immutant, and cause `AppClassLoader` to be used again there as well.

Everything that uses dynapath now should be using v0.2.3 - would it be possible to not inline it? Have you seen conflicts with dynapath not inlined?

bbatsov:
> Everything that uses dynapath now should be using v0.2.3 - would it be possible to not inline it? Have you seen conflicts with dynapath not inlined?

Don't recall such a case. Guess we can remove it from the inlined deps. //cc @benedekfazekas 

That said - maybe it doesn't make sense to inline the debugger as well, not that it doesn't have any third-party deps.

benedekfazekas:
of course. you just don't mark it with `^:source-deps` meta. in fact not inlining debugger might "fix" the current problem with #183 too

PR #193, opened 2015-04-16
Add test for session fixture

cichli:
The unsupported op test will hopefully stabilise the test coverage reports by ensuring the `(handler msg)` branch is always called for all middlewares (currently it seems to depend on how nREPL orders them which is not deterministic).


bbatsov:
:+1:

PR #201, opened 2015-04-30
Update pprint middleware to not hijack *out*

### LINKED ISSUE #200: Use a different slot than `out` in pprint middleware
Currently, we bind `*out*` to the nREPL session's `*out*` when printing - see [here](https://github.com/clojure-emacs/cider-nrepl/blob/master/src/cider/nrepl/middleware/pprint.clj#L15). This unfortunately means it's not possible for a client to differentiate between actual printed output and the result of evaluations - for example to send them to different buffers, or to syntax highlight one but not the other.

We can instead bind `*out*` to a new `PrintWriter` created by [`clojure.tools.nrepl.middleware.session/session-out`](https://github.com/clojure/tools.nrepl/blob/master/src/main/clojure/clojure/tools/nrepl/middleware/session.clj#L26) that uses a different slot. This will mean the result is still streamed back, and the client can decide whether to render the result as it arrives or accumulate it - we should also send some sentinel message after printing each value in case multiple forms are being evaluated.

Client-side we can modify `nrepl-make-response-handler` to treat the new slot as if it were `out` for now, but this will let us solve clojure-emacs/cider#943 eventually.

Thoughts? @laurentpetit will this suffice for differentiating between printed output and evaluation results in CCW?

cichli:
* The printed result is now streamed via the `:pprint-out` slot
* After each result is printed, a response is sent with a non-nil value in the `:pprint-sentinel` slot
* The `:value` slot is now elided from responses

Fixes #200.

Not including the `:value` slot means the client won't choke trying to decode or print it all at once. The client changes aren't ready yet, but I'm posting this for review + so @laurentpetit can be ready for if/when this merges.

Thoughts? Eventually on the client we'll be able to use this to prevent the interleaving of evaluation results and things written to `*out*` - either by sending them to different buffers, or queuing the `*out*` responses while an evaluation result is being printed.

### DIFF
--- a/src/cider/nrepl/middleware/pprint.clj
+++ b/src/cider/nrepl/middleware/pprint.clj
@@ -3,22 +3,31 @@
             [clojure.pprint :refer [pprint *print-right-margin*]]
             [clojure.tools.nrepl.middleware :refer [set-descriptor!]]
             [clojure.tools.nrepl.middleware.interruptible-eval :refer [*msg*]]
-            [clojure.tools.nrepl.middleware.pr-values :refer [pr-values]])
+            [clojure.tools.nrepl.middleware.pr-values :refer [pr-values]]
+            [clojure.tools.nrepl.middleware.session :as session]
+            [clojure.tools.nrepl.misc :refer [response-for]]
+            [clojure.tools.nrepl.transport :as transport])
   (:import clojure.tools.nrepl.transport.Transport))
 
+(defn- pprint-writer
+  [{:keys [session transport] :as msg}]
+  (#'session/session-out :pprint-out (:id (meta session)) transport))
+
 (defn pprint-reply
-  [{:keys [right-margin session] :as msg} response]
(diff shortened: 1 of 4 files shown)


laurentpetit:
I like the fact that it is somehow making nrepl more stream-like. This is also goes in the direction of socket based REPLs which may come in Clojure 1.7.
I will have to check how I need to consume `:pprint-out` but that's a CCW problem.

laurentpetit:
I'm not sure I understand correctly what
```
After each value, a response is sent with a non-nil value in the :pprint-sentinel slot
```

means

laurentpetit:
:+1:

bbatsov:
Looks good to me.

bbatsov:
Guess we should move forward with the client changes.

laurentpetit:
FWIW, I have updated the CCW REPL client to work with this PR (built and installed on my local `~/.m2/` repository and it works like a charm.

cichli:
@bbatsov sure, sorry for the delay - will get it done this evening.

PR #204, opened 2015-05-02
Add ns-path op

expez:
This op returns the file path to the file containing some ns on classpath.

The new op is meant to be the basis for cider-find-ns

### DIFF
--- a/src/cider/nrepl/middleware/ns.clj
+++ b/src/cider/nrepl/middleware/ns.clj
@@ -1,9 +1,16 @@
 (ns cider.nrepl.middleware.ns
-  (:require [clojure.tools.nrepl.transport :as transport]
-            [clojure.tools.nrepl.middleware :refer [set-descriptor!]]
-            [clojure.tools.nrepl.misc :refer [response-for]]
-            [cider.nrepl.middleware.util.cljs :as cljs]
-            [cljs-tooling.util.analysis :as cljs-analysis]))
+  (:require [cider.nrepl.middleware.util.cljs :as cljs]
+            [cljs-tooling.util.analysis :as cljs-analysis]
+            [clojure.java.classpath :as cp]
+            [clojure.java.io :as io]
+            [clojure.tools.namespace
+             [file :refer [read-file-ns-decl]]
+             [find :refer [clojure-sources-in-jar find-clojure-sources-in-dir]]]
+            [clojure.tools.nrepl
+             [middleware :refer [set-descriptor!]]
+             [misc :refer [response-for]]
+             [transport :as transport]])
+  (:import java.util.jar.JarFile))
 
 (defn ns-list-clj []
   (->> (all-ns)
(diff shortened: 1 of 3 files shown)


src/cider/nrepl/middleware/ns.clj
```diff
@@ -51,6 +58,43 @@
   (transport/send transport (response-for msg :ns-vars (ns-vars msg)))
   (transport/send transport (response-for msg :status :done)))
 
+(defn- jar-file?
+  "Returns true if file is a normal file with a .jar or .JAR extension."
```
bbatsov: Is someone really using .JAR?

src/cider/nrepl/middleware/ns.clj
```diff
@@ -51,6 +58,43 @@
...
+  [f]
+  (let [file (io/file f)]
+    (and (.isFile file)
```
bbatsov: The fact there's nothing like this in the standard lib is killing me.

src/cider/nrepl/middleware/ns.clj
expez: I stole this function from `tools.namespace`, where it's private, so I assume so :(

src/cider/nrepl/middleware/ns.clj
```diff
@@ -51,6 +58,43 @@
...
+         (or (.endsWith (.getName file) ".jar")
```
bbatsov: Guess you can downcase the filename and drop the `or`.

src/cider/nrepl/middleware/ns.clj
```diff
@@ -70,4 +115,8 @@
     "ns-vars"
     {:doc "Returns a sorted list of all vars in a namespace."
      :requires {"ns" "The namespace to browse"}
-     :returns {"status" "done"}}}}))
+     :returns {"status" "done"}}
+    "ns-path"
+    {:doc "Returns the path to the file containing ns"
```
bbatsov: Add a `.` at the end of the sentence.

expez:
Refactored.  

It's really annoying that the tests take like 20 mins.  Any thoughts about dropping support for all these old clojure versions?  The clojure team takes regression really seriously so I can't think of any reasons why someone would stay at say clojure 1.5 instead of upgrading.

bbatsov:
> Any thoughts about dropping support for all these old clojure versions? The clojure team takes regression really seriously so I can't think of any reasons why someone would stay at say clojure 1.5 instead of upgrading.

That's out of question. We got some negative feedback when I decided to drop support for 1.3 and 1.4 when many projects were still supporting them. There's also the fact that 1.5 is [widely used yet](https://cognitect.wufoo.com/reports/state-of-clojure-2014-results/). 

Unless there's some strong incentive to do so, I'd like us to support as many versions as possible. We'll probably drop 1.5 once 1.8 is out.

bbatsov:
:+1:

PR #205, opened 2015-05-11
don't attempt to read file paths unless they are actually files.

pleasetrythisathome:
Fixes instances where cljx files have the same name as directories in the classpath.

### DIFF
--- a/src/cider/nrepl/middleware/info.clj
+++ b/src/cider/nrepl/middleware/info.clj
@@ -87,7 +87,7 @@
                 (when (= "file" (.getProtocol rsrc))
                   (io/as-file rsrc))
                 (io/as-file filename))]
-    (when (and file (.exists file))
+    (when (and file (.exists file) (.isFile file))
       (with-open [in (io/input-stream file)]
         (.skip in (max 0 (- (.length file) 1024)))
         (->> (slurp in)


cichli:
LGTM, but please will you capitalise the commit message and remove the period at the end? :-)

pleasetrythisathome:
np

src/cider/nrepl/middleware/info.clj
```diff
@@ -87,7 +87,7 @@
                 (when (= "file" (.getProtocol rsrc))
                   (io/as-file rsrc))
                 (io/as-file filename))]
-    (when (and file (.exists file))
+    (when (and file (.exists file) (.isFile file))
```
expez: What problem did you run in to with the existing code?

If `file` isn't a `File` then the order of the statements in the `and` should be re-ordered or you risk running `.exists` on a String or whatever else `file` might be.

cichli: @expez a `File` can represent either a file or a directory, and if it's a directory, the following call to `io/input-stream` will throw.

@pleasetrythisathome just to clarify, you have a directory on the classpath whose name is something like `core.cljs`, and `find-cljx-source` is trying to open that instead of the actual `core.cljs`?

pleasetrythisathome:
@cichli yes that's exactly what i have. I often structure application where i have a fix like src/project/user.cljx and a directory src/project/user where i put implementation code for user.cljx. 

@expez I believe file in the when predicate can only ever be nil (there is no resource matching filename), or a java.io.File, which could either be a file or a directory, and in the directory case, io/input-stream will throw as @cichli said

bbatsov:
:+1:

PR #209, opened 2015-05-28
* debug.clj: Send *locals* with debug input requests

Malabarba:
This is the first part of displaying local variables in a debugging session. Emacs side of this will only come later (no more time right now) but this is very clean and backwards compatible, so I figured I'd make the PR before I forget.

Locals are sent as a list of pairs of strings.
Affects https://github.com/clojure-emacs/cider/issues/1106


Malabarba:
If I understand the travis report correctly, there's a single build failing because of a deprecation warning in using `Thread.stop()`. This usage isn't new, but in the current code I added a `^Thread` type hint, which is probably why the warning is now detected.

I found a description of the deprecation [here](https://www.securecoding.cert.org/confluence/display/java/THI05-J.+Do+not+use+Thread.stop%28%29+to+terminate+threads), but none of the offered alternatives are viable. Besides, `.stop()` is what the nrepl itself uses for interrupting evals last time I checked, so there's little reason for us to fix it before them.

expez:
Why did you add in the typehint?

Malabarba:
@expez To fix a reflection warning. :) From pan to fire.
Would it be preferable to undo the type hint?

Malabarba:
Commit reworded.

expez:
@bbatsov What do you mean to do about the eastwood failure?  I opened [this issue](https://github.com/jonase/eastwood/issues/165), but that's not an immediate solution.

Malabarba:
For an immediate workaround, I can just remove the type hint. I added it because I got tired of seeing the warning when compiling for android, but its performance impact is completely irrelevant.

bbatsov:
@Malabarba OK, let's do this.

Malabarba:
Done.

expez:
Looking forward to seeing this put to use! :)

PR #210, opened 2015-06-01
Use defaultLocale rather than "en"

cursork:
Setting the locale on a DocEnv object *sets the global default locale*. My symptom was that my locale was being set to `en_MT`, overwriting the existing default.

DocEnv defers the locale to DocLocale and the DocLocale constructor calls Locale.setDefault() with whatever it finds from Locale.getAvailableLocales(). It just so happened on my system that `en_MT` was the first match.

I'm not 100% sure on the best fix, as in an ideal world DocLocale wouldn't be setting the global. This change just passes through whatever the current default is when instantiating the DocEnv.

Example REPL session for reproduction:

```clojure
user=> (require 'cider.nrepl.middleware.util.java.parser)
nil
user=> (in-ns 'cider.nrepl.middleware.util.java.parser)
#<Namespace cider.nrepl.middleware.util.java.parser>
cider.nrepl.middleware.util.java.parser=> (def docenv (let [access   (ModifierFilter. ModifierFilter/ALL_ACCESS)
                                             #_=> context  (doto (Context.) (Messager/preRegister "cider-nrepl-javadoc"))
                                             #_=> options  (doto (Options/instance context) (.put "ignore.symbol.file" "y"))
                                             #_=> compiler (JavadocTool/make0 context)
                                             #_=> enter    (JavadocEnter/instance0 context)
                                             #_=> ]
                                             #_=> (DocEnv/instance context))
                                             #_=> )
#'cider.nrepl.middleware.util.java.parser/docenv
cider.nrepl.middleware.util.java.parser=> docenv
#<DocEnv com.sun.tools.javadoc.DocEnv@5e9081c3>
cider.nrepl.middleware.util.java.parser=> (java.util.Locale/getDefault)
#<Locale en_MT>
cider.nrepl.middleware.util.java.parser=> (.setLocale docenv "en_GB")
nil
cider.nrepl.middleware.util.java.parser=> (java.util.Locale/getDefault)
#<Locale en_GB>
```


expez:
That test failure is funny.  You'd think `setLocale` would have at least one version which accepts an actual `Locale`!

Do you know if the `Locale` affects the docs in some way?  Presumably the original author didn't write an extra line just to pad his resume :)

bbatsov:
Guess we can ping @jeffvalk to join this discussion.

cursork:
Well that test failure was just clumsy - sorry. Was trying to get out of the office and obviously had different code in my REPL to on disk. Locale has an appropriate toString so fixed now.

I was tempted to remove the setLocale (I'm not sure what effect it'd have), but then noticed a comment in the OpenJDK source for the DocEnv constructor:

    // Default.  Should normally be reset with setLocale.
    this.doclocale = new DocLocale(this, "", breakiterator);

So assumed that setLocale was expected to be invoked. I haven't had time to look in depth yet though.

jeffvalk:
I'm not aware of any specific need to have locale set to "en"; it just needed to be set. If this doesn't effect how javadoc text is parsed, it seems reasonable.

expez:
@cursork could you please squash the two commits?  Do you have time to figure out if this change in locale is going to have unintended side-effects?

cursork:
Squashed the commits.

Re. unintended side effects - I can investigate further, but at this point I'm unsure what I'm looking for. A basic test of changing Locale to Locale/GERMAN / "de" (which, as it happens, ends up being re-set to de_AT due to the same issue..) and a few other locales shows no difference in response to `(source-info ...)`

I only detected this because I have an app that specifically checks that the default locale hasn't been messed with on startup. Not due to any issues in the library.

expez:
We need this change in any event, because messing up the locale isn't acceptable, so I'm just going to merge it now.  If this has adverse effects on the javadoc parsing, for certain locales, we'll likely get bug reports pretty quickly and it won't be a big mystery to us what the problem is.

bbatsov:
Fair enough. Did you push to clojars as well?

expez:
Noet yet.  It would actually be great if you could take care of that.

bbatsov:
Done.

On 2 June 2015 at 10:24, Lars Andersen <notifications@github.com> wrote:

> Noet yet. It would actually be great if you could take care of that.
>
> —
> <https://github.com/clojure-emacs/cider-nrepl/pull/210#issuecomment-107857190>
> .
>



-- 
Best Regards,
Bozhidar Batsov

http://www.batsov.com

pesterhazy:
I was bitten by this hilarious bug as well: a pristine `lein repl` with just the cider plugin enabled sets my locale to `en_MT`.

PR #213, opened 2015-06-11
[Fix #196] Rewritten dependencies polluting ns-list

### LINKED ISSUE #196: mranderson deps appear on classpath and need to be filtered out
A similar bug to: https://github.com/clojure-emacs/refactor-nrepl/issues/66 might be affecting some of the cider ops.

completion of namespaces might be one.

expez:
Our own inlined deps, created by mranderson, as well as the inlined deps
from eastwood were polluting the result of the `ns-list` op.  These are
now filtered out.

This change improves all cider commands relying on a completing read of
some namespace e.g. `cider-browse-ns` and `cider-find-ns`

### DIFF
--- a/src/cider/nrepl/middleware/ns.clj
+++ b/src/cider/nrepl/middleware/ns.clj
@@ -12,8 +12,20 @@
              [transport :as transport]])
   (:import java.util.jar.JarFile))
 
+(defn inlined-dependency?
+  "Returns true if the namespace matches one of our, or eastwood's,
+  inlined dependencies."
+  [namespace]
+  (let [ns-name (str (ns-name namespace))]
+    (or
+     ;; rewritten by mranderson
+     (.startsWith ns-name "deps.")
+     ;; rewritten by dolly
+     (.startsWith ns-name "eastwood.copieddeps"))))
+
 (defn ns-list-clj []
   (->> (all-ns)
+       (remove inlined-dependency?)
        (map ns-name)
        (map name)
        (sort)))
--- a/test/clj/cider/nrepl/middleware/ns_test.clj
+++ b/test/clj/cider/nrepl/middleware/ns_test.clj
@@ -1,11 +1,13 @@
 (ns cider.nrepl.middleware.ns-test
-  (:require [cider.nrepl.middleware.ns :refer [ns-list-clj ns-vars-clj]]
+  (:require [cider.nrepl.middleware.ns
+             :refer [ns-list-clj ns-vars-clj inlined-dependency?]]
             [cider.nrepl.test-session :as session]
             [cider.nrepl.test-transport :refer [messages test-transport]]
             [clojure.test :refer :all]))
 
-(deftest test-toogle-ns-list
-  (is (= (count (all-ns)) (count (ns-list-clj)))))
+(deftest test-toggle-ns-list
+  (is (= (count (remove inlined-dependency? (all-ns)))
+         (count (ns-list-clj)))))
 
 (deftest test-toogle-ns-vars
   (let [ns "clojure.core"]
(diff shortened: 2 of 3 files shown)


expez:
Before I merge this, can you think of any other ops that might be adversely affected by our inlined deps, @bbatsov?

expez:
Ok, this was made a lot harder by the fact that `lein test` doesn't run any tests.  When I copied a line from `.travis.yml` I copied one that ran the tests only for cljs.

Is there a reason why `lein test` doesn't run all the tests?  

Also, wouldn't the travis builds run a lot faster if we just asked it to run both clj and cljs tests at the same time for each version?

bbatsov:
I'll take a look at this later.

On Thursday, June 11, 2015, Lars Andersen <notifications@github.com> wrote:

> Ok, this was made a lot harder by the fact that lein test doesn't run any
> tests. When I copied a line from .travis.yml I copied one that ran the
> tests only for cljs.
>
> Is there a reason why lein test doesn't run all the tests?
>
> Also, wouldn't the travis builds run a lot faster if we just asked it to
> run both clj and cljs tests at the same time for each version?
>
> —
> <https://github.com/clojure-emacs/cider-nrepl/pull/213#issuecomment-111195418>
> .
>


-- 
Best Regards,
Bozhidar Batsov

http://www.batsov.com

src/cider/nrepl/middleware/ns.clj
```diff
@@ -12,10 +12,19 @@
              [transport :as transport]])
   (:import java.util.jar.JarFile))
 
+(defn inlined-dependency?
```
bbatsov: You might add a docstring for this.

test/clj/cider/nrepl/middleware/ns_test.clj
```diff
@@ -18,7 +23,11 @@
 (deftest ns-list-integration-test
   (let [ns-list (:ns-list (session/message {:op "ns-list"}))]
     (is (sequential? ns-list))
-    (is (every? string? ns-list))))
+    (is (every? string? ns-list))
+    (testing "Removal of namespaces created by source rewriting"
+      (is (every? #(not (or (.startsWith % "deps.")
```
bbatsov: There's also `not-any?`.

bbatsov:
Seems to me that this change should be safe.

expez:
> Seems to me that this change should be safe.

What I was wondering was if you could think of any other ops that might benefit from similar filtering of inlined dependencies.

We might consider filtering compliment's completion candidates. //cc @alexander-yakushev

bbatsov:
Oh, sorry. Filtering the completion candidates makes some sense, although I doubt it's worth it - who'd be using using `cider`'s namespaces directly? Filtering the classpath and apropos matches might be more sensible.

alexander-yakushev:
In a marginal case where user completes `de█` or `eas█`, yes, this will help. Nevertheless, Compliment will never depend on CIDER, so it is easier just to copy the filter into compliment, looks simple enough.

expez:
> who'd be using using cider's namespaces directly?

Entering `(dep` in the repl and you'll see, among other candidates: `deps.alembic.v0v3v2.deps.classlojure.v0v7v0.classlojure.core`. 

 I agree that this isn't a big deal.

bbatsov:
@expez Well, guess this is best hardcoded in `compliment` then. Can this prefix be customized btw? Maybe the default should be something less common?

expez:
> Can this prefix be customized btw? Maybe the default should be something less common?

Yes, this is something I've mentioned to @benedekfazekas myself.

bbatsov:
At the very least it should be something that starts with the name of the project in question (e.g. `cider.deps` or `cider.nrepl.deps`.

benedekfazekas:
prefix can't be customized atm, needs a mranderson change.
prefixing with project name: good shout. again needs a mranderson change.

benedekfazekas:
is this a showstopper for 0.9 btw? i anyway want to rewrite mranderson (almost from scratch, using @expez's move directory/package magic). so i might postpone and do it together when the big overhaul is done there...

src/cider/nrepl/middleware/ns.clj
expez: done

expez:
> Filtering the classpath and apropos matches [...]

I've filtered the apropos matches.  The classpath isn't an issue because only `cider-nrepl` and `refactor-nrepl` appear on the classpath.  The mess created by our bundled dependencies are online visible inside the jars :)

arrdem:
:+1:

PR #216, opened 2015-06-16
[Fix clojure-emacs/cider#1134] Don't breakpoint maps in arglists

Malabarba:
We already didn't instrument inside maps, but we still wrapped breakpoints around them. This is bad sometimes.

The fix is this [one line change](https://github.com/clojure-emacs/cider-nrepl/compare/master...Malabarba:master#diff-156621f66b9a5c16a8d5eabf6bc46cd7L285). The rest of the diff is cleanup related to this map thing.


bbatsov:
Thanks!

PR #220, opened 2015-06-20
Rewrite the debugger

Malabarba:
This is the clojure counterpart to https://github.com/clojure-emacs/cider/pull/1149
This completely rewrites the debugger (again) to be simpler and (I believe) more reliable.

The big complication with the previous debugger were macros. 
- You can't just instrument everything inside a macro, that will surely break stuff.
- You can't expand the macro before instrumenting it, because then you completely lose track of the original code, you won't be able to know how to find a given sexp on the original source file, and you will end up wrapping brakpoints around code that the user didn't write.

The previous debugger took a lesson from Edebug and tried to fix the first option by instrumenting the macro arguments according to a specification. This worked surprisingly well, but:
- It's complicated as hell
- If the user defines their own macros, they're unlikely to follow the specification we expect, which means it won't be debuggable.
- Even within the clojure.core namespace it was difficult to test everything (for instance, in writing this, I just realised we still don't support `letfn`).

The version proposed here goes with the second option instead, by using something that's not available in emacs-lisp, metadata.

1. First, we walk through the code and attach metadata to every object inside it, recording the location of this object in the code.
2. Then we macroexpand everything, so we don't have to worry about macros. We still need to provide special hadling for special-forms, but they are limited in number (see `instrument-special-form`).
3. Then we walk through the code again, looking for the metadata we had written. Wherever we find it, we know this correponds to user code and we know its original location, so we can use it to make a breakpoint.

Looking at the diffs, it should go without saying that this is much simpler. About half the original code was dedicated to parsing macros.

### Notable differences ###

- This also introduces a new way of instrumenting. Before, we had to call `instrument` on a top-level sexp, and this would instrument everything. Now, we call `instrument-tagged-code` on a top-level sexp, and it will *only* instrument objects with a `:cider-breakfunction` metadata. In particular, this means that if you call `instrument-tagged-code` on a clean sexp, it won't do anything. You have to call `debug-reader` or `breakpoint-reader` on the sexp first, but this is stomething that already happened if the sexp was read with a `#dbg` or `#bp` reader macro.
- Instead of the hacky way we did previously (where cider.el sent us an “eval” message with the code manually wrapped in a `instrument-and-eval`), the `wrap-debug` middleware now hooks cleanly into the “eval” op. If neither one of the reader macros (`#dbg` or `#bp`) are present in the received code, the message just falls through for regular evaluation. If one of the macros *was* present, `wrap-debug` just adds its own evaluator to the message before letting it fall through.
- For now, “eval” messages need a file and a point to be properly debuggable. But I've been thinking of getting rid of that. We could instead just popup a temp buffer displaying the code being debugged (this would be easier and wouldn't break if the user edited the source).


src/cider/nrepl/middleware/debug.clj
```diff
@@ -177,47 +179,77 @@
...
-  (eval
-   (instrument (merge {:coor [], :breakfunction #'breakpoint} ex)
-               form)))
+     (let [val# ~value]
+       (cond
+         (skip-breaks? ~coor) val#
+         (not (seq @debugger-message)) (do (println "Debugger not initialized")
+                                           (skip-breaks! true)
+                                           val#)
+         true (read-debug-command
```
expez: It's customary to use `:else` to denote the default branch in clojure.

expez:
`macrostep` extends `edebug` in an interesting way, letting you expand a macro and then step through the expanded form in the debugger.  It reads to me like this functionality is within reach now, if we want it.

I've just given this a cursory glance because I was curious but this looks like a sound approach!

src/cider/nrepl/middleware/debug.clj
Malabarba: :+1:

Malabarba:
@expez I think it would be possible. We would essentially have to use debugger's "eval in lexical scope" functionality where the code would be like `(instrument-code (debug-reader (macroexpand 'CODE-AT-POINT)))`. If nothing goes wrong, that would start a new debugging session inside the current one.

Malabarba:
One of the erros is elluding me. https://travis-ci.org/clojure-emacs/cider-nrepl/jobs/68627068#L151

I can't figure out which specific test is causing this error, simply because I can't reproduce it. =/


```
lein test cider.nrepl.middleware.inspect-test
ERROR: Unhandled REPL handler exception processing message {:code (sorted-map :a {:b 1} :c "a" :d 'e :f [2 3]), :id 8d1ae6f9-b6f3-426d-a613-52760c59229a, :inspect true, :op eval, :session 48f91a00-c2eb-4ac2-a7f6-81ebc7e444cb}
java.lang.ClassCastException: java.lang.String cannot be cast to clojure.lang.Atom
	at clojure.core$swap_BANG_.doInvoke(core.clj:2163)
	at clojure.lang.RestFn.invoke(RestFn.java:652)
	at cider.nrepl.middleware.debug$prepare.invoke(debug.clj:222)
	at cider.nrepl.middleware.debug$wrap_debug$fn__12510.invoke(debug.clj:251)
	at clojure.tools.nrepl.middleware$wrap_conj_descriptor$fn__4894.invoke(middleware.clj:24)
	at clojure.tools.nrepl.server$handle_STAR_.invoke(server.clj:19)
	at clojure.tools.nrepl.server$handle$fn__7518.invoke(server.clj:28)
	at clojure.core$binding_conveyor_fn$fn__8778.invoke(core.clj:1836)
	at clojure.lang.AFn.call(AFn.java:18)
	at java.util.concurrent.FutureTask.run(FutureTask.java:262)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1145)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:615)
	at java.lang.Thread.run(Thread.java:745)
```

Malabarba:
Tests passing.

src/cider/nrepl/middleware/debug.clj
```diff
@@ -5,9 +5,18 @@
...
+            [cider.nrepl.middleware.util.cljs :as cljs]
+            [clojure.walk :as walk])
   (:import [clojure.lang Compiler$LocalBinding]))
 
+(defn random-uuid-str []
+  (letfn [(hex [] (format "%x" (rand-int 15)))
+          (nhex [n] (apply str (repeatedly n hex)))]
+    (let [rhex (format "%x" (bit-or 0x8 (bit-and 0x3 (rand-int 14))))]
+      (str (nhex 8) "-" (nhex 4) "-4" (nhex 3)
+        "-" rhex (nhex 3) "-" (nhex 12)))))
```
Malabarba: `cljfmt` indented this line like this, but it looks like a bug to me.

bbatsov: Definitely a bug. You should probably report it upstream.

Malabarba:
FIled https://github.com/weavejester/cljfmt/issues/42

PR #211, opened 2015-06-04
Avoid overwriting default locale

rkworks:
This PR is related to #210.
 
That change causes weird behavior in my environment.
If "ja_JP" (my default locale)  is passed to DocEnv/setLocale, the default locale will be overwritten to "ja_JP_JP_#u-ca-japanese".
When new locale is passed to DocEnv/setLocale again, it becomes the error. 

I've read source code of com.sun.tools.javadoc.DocLocale and found that `getLocale` method returns the default locale itself when passed locale name is empty string.

```clj
user=> (require 'cider.nrepl.middleware.util.java.parser)
nil
user=> (in-ns 'cider.nrepl.middleware.util.java.parser)
#<Namespace cider.nrepl.middleware.util.java.parser>
cider.nrepl.middleware.util.java.parser=> (def docenv (let [access   (ModifierFilter. ModifierFilter/ALL_ACCESS)
                                     #_=>                   context  (doto (Context.) (Messager/preRegister "cider-nrepl-javadoc"))
                                     #_=>                   options  (doto (Options/instance context) (.put "ignore.symbol.file" "y"))
                                     #_=>                   compiler (JavadocTool/make0 context)
                                     #_=>                   enter    (JavadocEnter/instance0 context)]
                                     #_=>               (DocEnv/instance context)))
#'cider.nrepl.middleware.util.java.parser/docenv
cider.nrepl.middleware.util.java.parser=> (Locale/setDefault (Locale. "ja" "JP"))
nil
cider.nrepl.middleware.util.java.parser=> (Locale/getDefault)
#<Locale ja_JP>
cider.nrepl.middleware.util.java.parser=> (.setLocale docenv (str (Locale/getDefault)))
nil
cider.nrepl.middleware.util.java.parser=> (Locale/getDefault)
#<Locale ja_JP_JP_#u-ca-japanese>
cider.nrepl.middleware.util.java.parser=> (.setLocale docenv (str (Locale/getDefault)))
cider-nrepl-javadoc: エラー - ロケールja_JP_JP_#u-ca-japaneseが無効です

ExitJavadoc   com.sun.tools.javadoc.Messager.exit (Messager.java:300)
cider.nrepl.middleware.util.java.parser=> (Locale/setDefault (Locale. "ja" "JP"))
nil
cider.nrepl.middleware.util.java.parser=> (.setLocale docenv "")
nil
cider.nrepl.middleware.util.java.parser=> (Locale/getDefault)
#<Locale ja_JP>
```

PR #222, opened 2015-06-30
More debugger

Malabarba:
- Debugger's eval and inject also accept code non-interactively. Instead of sending :eval on the message's "input" entry, send something like `{:response :eval, :code 123}`. This allows cider.el to integrate regular eval commands with the debugger.  
  *In short*, commands like `C-x C-e` or `C-c M-:` will evaluate code in the lexical context while debug-mode is active.
- Since the order inside maps is unreliable, try to use map keys as a coordinate. Affects #211.  
  cider-debug-map
- There are now commands for inspecting values in the lexical context. `:inspect` and `:locals`, in particular.  
  cider-debug-inspect-locals
- When a form has no :cider-coor metadata, but it is a list and its `first` *does* have that metadata, this is probably a proper code form built by a macro (like the threading macros). The debugger will now use the meta of the `first` to guess the coordinates of the parent form.  
  In other words: added support for `->`, `->>`, etc.  
  cider-debug-threading-macro

### DIFF
--- a/src/cider/nrepl/middleware/debug.clj
+++ b/src/cider/nrepl/middleware/debug.clj
@@ -6,7 +6,9 @@
             [clojure.tools.nrepl.misc :refer [response-for]]
             [clojure.tools.nrepl.middleware.interruptible-eval :refer [*msg*]]
             [cider.nrepl.middleware.util.instrument :as ins]
+            [cider.nrepl.middleware.inspect :refer [swap-inspector!]]
             [cider.nrepl.middleware.util.cljs :as cljs]
+            [cider.nrepl.middleware.util.inspect :as inspect]
             [clojure.walk :as walk])
   (:import [clojure.lang Compiler$LocalBinding]))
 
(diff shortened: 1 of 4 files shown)


Malabarba:
Screenshots to come.

Malabarba:
Screenshots are up, and the tests should be silenced now.

bbatsov:
The build is still failing for some profiles, though.

Malabarba:
Yeah, I'm looking into that.

src/cider/nrepl/middleware/debug.clj
```diff
@@ -160,22 +171,31 @@
...
     out: Skip breakpoints in the current sexp.
+    inspect: Evaluate an expression and inspect it.
+    locals: Inspect local variables.
     inject: Evaluate an expression and return it.
     eval: Evaluate an expression, display result, and prompt again.
-    quit: Abort current eval session."
+    quit: Abort current eval session.
+
+  Response received can be any one of these values, or a map
+  whole :response entry is one of these values."
```
expez: This sentence is slightly mangled

Malabarba: Indeed. Fixed.

Malabarba:
At last, I have defeated cloveral.

bbatsov:
:+1:

test/clj/cider/nrepl/middleware/debug_test.clj
```diff
@@ -103,3 +103,44 @@
...
+
+(deftest breakpoint-reader
+  (is (:cider-breakfunction (meta (d/breakpoint-reader '[a b c]))))
+  (is (= '[a :b 10 "ok"]
+         (remove #(:cider-breakfunction (meta %)) (d/breakpoint-reader '[a :b 10 "ok"]))))
+  ;; Just don't error
+  (is (map d/breakpoint-reader '[a :b 10 "ok"])))
+
+(deftest reader-macros
+  (binding [*data-readers* {'dbg d/debug-reader}]
```
expez: @Malabarba If user code is evaluated in this context then the correct thing to do is to merge these two maps, not doing a replacement.

PR #223, opened 2015-07-01
Two things all around

Malabarba:
- If a var's file is not set, info derives it from the namespace and dissocs the line (this one is related to https://github.com/clojure-emacs/cider/pull/1180).
- [Fix clojure-emacs/cider#1178] Handle exceptions in debugger eval.


bbatsov:
:+1:

PR #227, opened 2015-07-02
List instrumented defs

Malabarba:
This PR (along with the [cider side](https://github.com/clojure-emacs/cider/pull/1183)) does two things:

1. A new op to list all currently instrumented definitions. This is pretty reliable as far as `def`s go, but it doesn't do protocols. The cider side uses this to offer a `cider-browse-instrumented-defs` command.
2. A system for notifying cider whenever a new def has been instrumented or uninstrumented. So that cider can use that information to highlight instrumented defs in the source. [This video](https://twitter.com/AMalabarba/status/616192459933130752) shows that in action (the box around the name).

Item 1. is pretty straightforward, but I'd like some input on item 2.  
What it needs to do is wait for the `eval` op to finish, and then send a message to the client listing all the instrumented defs in that namespace. The best way I found of running something after the `eval` op was to add a watch to the session atom, because this atom is reset by the nrepl after the eval op is done.  
This has been working fine for me, but if anyone has thoughts on a better way to do this I'm all ears.

### DIFF
--- a/src/cider/nrepl/middleware/debug.clj
+++ b/src/cider/nrepl/middleware/debug.clj
@@ -256,10 +256,24 @@
 ;;; Middleware setup
 (defn- maybe-debug
   "Return msg, prepared for debugging if code contains debugging macros."
-  [{:keys [code session] :as msg}]
+  [{:keys [code session ns] :as msg}]
   (when (instance? clojure.lang.Atom session)
     (swap! session update-in [#'*data-readers*] assoc
-           'dbg #'debug-reader 'break  #'breakpoint-reader))
+           'dbg #'debug-reader 'break  #'breakpoint-reader)
+    ;; The session atom is reset! after eval, so it's the best way of
+    ;; running some code after eval is done. Since nrepl evals are async
+    ;; we can't just run this code after propagating the message.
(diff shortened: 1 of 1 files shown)


bbatsov:
> What it needs to do is wait for the eval op to finish, and then send a message to the client listing all the instrumented defs in that namespace. The best way I found of running something after the eval op was to add a watch to the session atom, because this atom is reset by the nrepl after the eval op is done.
> 
> This has been working fine for me, but if anyone has thoughts on a better way to do this I'm all ears.

What's the request id that those responses will have?

Malabarba:
@bbatsov 

> What's the request id that those responses will have?

They use the `id` and `transposrt` of the `debugger-message`. So it goes straight to cider-debug ([here](https://github.com/clojure-emacs/cider/pull/1183/files#diff-d17bdfe71e28776dfbca0b5de5d3bb7fR126)), and it's pretty invisible to the command that sent the eval request.

bbatsov:
> They use the id and transposrt of the debugger-message. So it goes straight to cider-debug (here), and it's pretty invisible to the command that sent the eval request.

Excellent, that's what I assumed by looking at the code, but I wanted to double check it.

PR #228, opened 2015-07-07
Fix debugger inspect

Malabarba:
Unrelated to that cider PR I posted just now.

### DIFF
--- a/src/cider/nrepl/middleware/debug.clj
+++ b/src/cider/nrepl/middleware/debug.clj
@@ -173,9 +173,11 @@
 (declare read-debug-command)
 
 (defn inspect-then-read-command
-  "Inspect `value` and send it as part of a new `read-debug-command`."
-  [extras value]
-  (let [i (pr-str (:rendered (swap-inspector! *msg* inspect/start value)))]
+  "Inspect `inspect-value` and send it as part of a new `read-debug-command`.
+  This `read-debug-command` is passed `value` and the `extras` map
+  with the result of the inspection `assoc`ed in."
+  [value extras inspect-value]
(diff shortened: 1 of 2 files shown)


bbatsov:
It's kind of unclear to me what the actual problem is. Can you elaborate a bit on it?

src/cider/nrepl/middleware/debug.clj
```diff
@@ -173,9 +173,9 @@
 (declare read-debug-command)
 
 (defn inspect-then-read-command
-  "Inspect `value` and send it as part of a new `read-debug-command`."
-  [extras value]
-  (let [i (pr-str (:rendered (swap-inspector! *msg* inspect/start value)))]
+  "Inspect `inspect-value` and send it as part of a new `read-debug-command`."
```
bbatsov: inspect "inspect-value" - that sounds a bit weird. Maybe if the other 2 params made an appearance in the docstring it would have been better.

Malabarba:
Yes. The first argument passed to read-debug-command is the value that it'll return if the user types n. In other words, it's the value that the instrumented sexp will return. 
I was inadvertently passing there the value we were supposed to inspect. 
Inspecting worked anyway (because that's done in parallel) but, instead of *just* Inspecting, this also changed the return value of the instrumented sexp as if the user had used :inject. 

Is that clearer? 

I'll add all that to the commit message tomorrow.

src/cider/nrepl/middleware/debug.clj
Malabarba: Good point. I'll do that.

bbatsov:
Yeah, much clearer.

Malabarba:
Tests fixed here too.

bbatsov:
:+1:

PR #230, opened 2015-07-08
Fix instrumenting try forms

Malabarba:
We don't want breakpoints around `catch` or `finally`.


bbatsov:
Probably `comment` as well.

Malabarba:
It's ok, the `comment` form doesn't get this far. It just becomes `nil` when we macroexpand everything, and wrapping in breakpoints is done after macroexpansion.

bbatsov:
:+1:

PR #231, opened 2015-07-09
Limit the size of data printed by the debugger

Malabarba:
Goes with https://github.com/clojure-emacs/cider/pull/1191


src/cider/nrepl/middleware/debug.clj
```diff
@@ -117,12 +117,19 @@
   "Bound by the `breakpoint` macro to the local &env."
   {})
 
+(defn pr-short
+  "Like `pr-str` but limited in length and depth."
+  [x]
+  (binding [*print-length* 6
```
bbatsov: Isn't it better to give some control to the users to configure this? E.g. we can pass some data from cider. Or maybe this is just going to be overkill...

Malabarba: Pushed an updated that lets us specify these values when the debugger is initialized. I'll make the cider.el counterpart soon.

Malabarba:
Finally, tests are passing.

bbatsov:
:+1:

PR #233, opened 2015-07-12
Add `before` and `after` options to refresh middleware

cichli:
I'm still refining the elisp bits for this, but am posting the middleware changes for review. The options are passed in as namespace-qualified function names, which must have zero arity. `before` is invoked before refreshing, and refreshing does not proceed if it throws an exception. `after` is invoked after refreshing.

`refresh-tracker` is now an agent, and the logic for refreshing dispatched using `send-off` - this means that only one `refresh` request will be being processed at any given time, so you don't have to worry about your `before` and `after` functions running concurrently or being retried. This is also means they won't lock up an nREPL thread if they block for a long time.

`*out*` and `*err*` are both captured and streamed back to the client while invoking both `before` and `after`, and if an exception is thrown either during their execution or while reloading, it is printed to `*out*` as it would be at the REPL (using `clojure.main/repl-caught`) and sent back to the client in stacktrace format for rendering in the error buffer.

Here's the client UI so far (a log buffer that prints the output of `before` and `after`, the list of namespaces being refreshed, and whether or not it was successful):

screenshot 2015-07-12 20 02 04

### DIFF
--- a/src/cider/nrepl/middleware/refresh.clj
+++ b/src/cider/nrepl/middleware/refresh.clj
@@ -8,7 +8,7 @@
             [clojure.tools.namespace.repl :as repl]
             [clojure.tools.namespace.track :as track]))
 
-(defonce ^:private refresh-tracker (atom (track/tracker)))
+(defonce ^:private refresh-tracker (agent (track/tracker)))
 
 (defn- scan
   [tracker scan-fn dirs]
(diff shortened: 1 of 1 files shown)


bbatsov:
The changes look good to me. Btw, this log buffer reminds me that it might be nice to add some generic "server-side" logging to `cider-nrepl`.

cichli:
Pushed a slight tweak to send a message both prior and subsequent to calling `before`/`after` - otherwise, if they don't emit any output, the client has no indication of them being called.

> Btw, this log buffer reminds me that it might be nice to add some generic "server-side" logging to cider-nrepl.

Do you mean something that user programs hook into, or some logging performed by `cider-nrepl` itself (enabled in the client by some debug mode flag, presumably)?

bbatsov:
> Do you mean something that user programs hook into, or some logging performed by cider-nrepl itself (enabled in the client by some debug mode flag, presumably)?

The latter. This would be valuable info when trying to pin point where a problem originates (or so I think).

src/cider/nrepl/middleware/refresh.clj
```diff
@@ -18,31 +21,116 @@
...
+              (format "%s is not a single-arity fn" sym))))
+
+    (binding [*msg* msg
+              *out* (get @session #'*out*)
+              *err* (get @session #'*err*)]
+      (@var))))
+
 (defn- reloading-reply
-  [tracker {:keys [transport] :as msg}]
+  [{reloading ::track/load}
```
Malabarba: Nice, I didn't know about this syntax.

PR #235, opened 2015-07-14
[Fix #229] Reset *skip-breaks* before stopping thread

### LINKED ISSUE #229: Quitting the debugger doesn't reset session bindings
A couple of debugger features rely a little on the fact that session bindings are reset after the session is finished. However, that doesn't happen if the user `:quit`s the debugger.

Malabarba:


bbatsov:
:+1:

PR #236, opened 2015-07-15
Rewrite the debugger a bit to depend less on *msg*

Malabarba:
Goes with https://github.com/clojure-emacs/cider/pull/1206

This it so we don't rely on the session bindings map, which was unreliable. In addition it should allow the debugger to work on cases where `*msg*` is not bound.


bbatsov:
:+1:

PR #237, opened 2015-07-17
Include source/line/column in location info

j-mckitrick:
@bbatsov @Malabarba @jeffvalk 

Here's the first round of changes to return location info.

### DIFF
--- a/src/cider/nrepl/middleware/stacktrace.clj
+++ b/src/cider/nrepl/middleware/stacktrace.clj
@@ -104,8 +104,13 @@
   from its message and discard the string representation of its inner cause."
   [{:keys [class] :as cause}]
   (if (= class "clojure.lang.Compiler$CompilerException")
-    (update-in cause [:message] str/replace
-               #".* (compiling:)\((.*)\)" "Error $1 $2")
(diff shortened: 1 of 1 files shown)


Malabarba:
Looks fine from here.

(for reference: clojure-emacs/cider#715)

bbatsov:
:+1: Looks fine to me as well.

j-mckitrick:
So the next step will be pulling those bits up into emacs. I've done a bit
of elisp but very little navigation. What steps would you suggest next?

Malabarba:
I think the next might be changing `nrepl-make-response-handler` to take this `line` and `column` information (with `nrepl-dbind-response`) and pass it to the error handling function.

bbatsov:
What's the state here? I seem to recall @j-mckitrick something along the lines that those changes were redundant, but forgot the details.

j-mckitrick:
Yes, it looks like things should be handled purely on the emacs side.  I think @Malabarba might have had an idea where to do it.

Malabarba:
I remember discussing this at some point, but my memory is not my forte. :P 
Hopefully I wrote my thoughts on a comment somewhere.

PR #242, opened 2015-08-05
Refresh improvements - clear op and support for disabling reloading of namespaces

cichli:
* Sometimes the refresh tracker can get in a state that makes reloading impossible - this can happen if a file has been deleted, or if a circular dependency error occured. The `clear` op does the same thing as `clojure.tools.namespace.repl/clear`, resetting the refresh tracker to its default value. The trade-off of doing this is that stale code from any deleted files may not be completely unloaded - I'll document all of this in the client README when I add support for the op there.

* The problem with using `clojure.tools.namespace.repl/remove-disabled` directly is that it looks up namespaced keywords in each namespace's metadata, and these namespaced keywords are rewritten by mranderson, so will not match those in user code. We workaround this by constructing the correct keywords at runtime, and inlining `remove-disabled`.

### DIFF
--- a/src/cider/nrepl/middleware/refresh.clj
+++ b/src/cider/nrepl/middleware/refresh.clj
@@ -152,6 +152,13 @@
                   (error-reply {:error e} msg)
                   tracker)))))
 
+(defn- clear-reply
+  [{:keys [transport] :as msg}]
+  (send-off refresh-tracker (constantly (track/tracker)))
+  (transport/send
+   transport
+   (response-for msg {:status :done})))
+
 (defn wrap-refresh
   "Middleware that provides code reloading."
   [handler]
@@ -159,6 +166,7 @@
     (case op
       "refresh" (refresh-reply (assoc msg :scan-fn dir/scan))
       "refresh-all" (refresh-reply (assoc msg :scan-fn dir/scan-all))
+      "refresh-clear" (clear-reply msg)
       (handler msg))))
 
 (set-descriptor!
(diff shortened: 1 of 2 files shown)


src/cider/nrepl/middleware/refresh.clj
```diff
@@ -1,10 +1,10 @@
-(ns cider.nrepl.middleware.refresh
+(ns ^{:clojure.tools.namespace.repl/load false
```
expez: This was added for testing purposes or?

cichli: Yeah, I was previously just using this locally when developing the middleware, since otherwise the refresh tracker state gets blown away every time I reload.

But since tools.namespace doesn't reload source in JARs, it shouldn't affect user code that depends on cider-nrepl (except for the case when cider-nrepl is a checkout dependency, but in that case you do want reloading of this namespace to be disabled - otherwise you only get full reloads, not incremental reloads, because the refresh tracker state is being blown away again).

src/cider/nrepl/middleware/refresh.clj
```diff
@@ -132,6 +148,14 @@
                   (error-reply {:error e} msg)
                   tracker)))))
 
+(defn- clear-reply
+  [{:keys [transport] :as msg}]
+  (send-off refresh-tracker (constantly (track/tracker)))
+
```
bbatsov: This newline seems redundant to me.

src/cider/nrepl/middleware/refresh.clj
bbatsov: Maybe this is worth some comment or something.

bbatsov:
> The problem with using clojure.tools.namespace.repl/remove-disabled directly is that it looks up namespaced keywords in each namespace's metadata, and these namespaced keywords are rewritten by mranderson, so will not match those in user code. We workaround this by constructing the correct keywords at runtime, and inlining remove-disabled.

Nice catch.

src/cider/nrepl/middleware/refresh.clj
cichli: Yep, don't know how that snuck in :-).

cichli:
I've addressed your comments - thanks for reviewing!

bbatsov:
:+1:

PR #244, opened 2015-08-16
Instrument def with docstring for debugging

grammati:
Previously, instrumentation would drop the value expression, and the
docstring docstring would become the value.


Malabarba:
Good catch. The code looks good! :+1: 

Why coveralls reports a coverage decrease when you literally changed a `second` to a `rest` **and** actually added tests is beyond me.

bbatsov:
Looks good. Just remove the `.` from the commit message's title (as titles are not sentences as far as I'm concerned).

grammati:
Commit message updated.

bbatsov:
:+1: 10x!

PR #245, opened 2015-08-18
Instrument deftest for debugging

grammati:
This change allows deftest forms to be debugged, by instrumenting the
test function that is attached to the var as metadata.

### DIFF
--- a/src/cider/nrepl/middleware/util/instrument.clj
+++ b/src/cider/nrepl/middleware/util/instrument.clj
@@ -99,8 +99,14 @@
             '#{.} `(~(instrument (first args))
                     ~(second args)
                     ~@(instrument-coll (rest (rest args))))
-            '#{def set!} (list* (with-meta-safe (first args) {:cider-instrumented true})
-                                (map instrument (rest args)))
+            '#{def} (let [sym (first args)]
+                      (list* (if (meta sym)
+                               (with-meta-safe sym (assoc (instrument (meta sym))
+                                                          :cider-instrumented true))
+                               sym)
+                             (map instrument (rest args))))
+            '#{set!} (list (with-meta-safe (first args) {:cider-instrumented true})
+                           (instrument (second args)))
             '#{loop* let* letfn*} (cons (vec (map-indexed (fn [i x] (if (odd? i) (instrument x) x))
                                                           (first args)))
                                         (instrument-coll (rest args)))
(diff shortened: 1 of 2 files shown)


grammati:
This is the minimal change I could find that lets me debug tests.

I have another branch (https://github.com/grammati/cider-nrepl/tree/instrument-meta) where I attempt to more generally instrument all metadata, but I have not managed to get it fully working - or at least, I have been unable to come up with test cases to demonstrate that it's useful in any practical way. So I decided to submit this smaller-scoped change for now.

src/cider/nrepl/middleware/util/instrument.clj
```diff
@@ -99,8 +99,13 @@
             '#{.} `(~(instrument (first args))
                     ~(second args)
                     ~@(instrument-coll (rest (rest args))))
-            '#{def set!} (list* (with-meta-safe (first args) {:cider-instrumented true})
-                                (map instrument (rest args)))
+            '#{def} (let [sym (first args)]
+                      (list* (if (meta sym)
+                               (with-meta-safe sym (instrument (meta sym)))
```
Malabarba: Also assoc `:cider-instrumented true` in the return value of `(instrument (meta sym))`.

Malabarba:
Besides the missing `assoc`, this looks good. Thanks a lot for taking the time!

> I have another branch where I attempt to more generally instrument all metadata, but I have not managed to get it fully working 

I had a look at your branch and I can't immediately tell why it didn't work.

>- or at least, I have been unable to come up with test cases to demonstrate that it's useful in any practical way. So I decided to submit this smaller-scoped change for now.

Agreed. 
“useful” is the important word here. Even if that other branch worked, I'm not sure I'd want to add that much complexity to the code just to support a use case we don't know if anyone wants.

test/clj/cider/nrepl/middleware/util/instrument_test.clj
```diff
@@ -170,3 +170,7 @@
          '#{[(set! (. inst field) (bp (bar) [2])) []] [(bar) [2]]}))
   (is (= (breakpoint-tester '(set! (.field inst) (bar)))
          '#{[(set! (. inst field) (bp (bar) [2])) []] [(bar) [2]]})))
+
+(deftest instrument-deftest
+  (is (= (breakpoint-tester '(deftest foo (bar)))
+         '#{[(bar) [2]]})))
```
grammati: @Malabarba, do you by any chance have a guess why this test behaves differently depending on how it is run? If I run from cider (C-c C-,), it passes as-is. If I run "lein with-profile +test-clj test", it fails, expecting a different result (https://travis-ci.org/clojure-emacs/cider-nrepl/jobs/76097543#L224). I have no clue why.

Malabarba:
Also, please change the commit title to say “[Fix #...] Instrument var metadata in def forms”, where `...` is the number of the original issue.
Also add a changelog entry saying something like “Add debugger support for `deftest`.”.

test/clj/cider/nrepl/middleware/util/instrument_test.clj
Malabarba: @grammati Looks like it's not expanding the deftest. Maybe you need to add a `(use 'clojure.test)` inside the test.

grammati:
@Malabarba, I made the changes you asked for, and found a way to make the tests pass consistently. Should be good to go now, I think.

Malabarba:
Awesome. Thanks a ton!

bbatsov:
@Malabarba What's your clojar's username? We need to add you there, so you can update the snapshot build (after merging something it has to be updated manually, at least for the time being).

Malabarba:
I believe it's malabarba

bbatsov:
You've got the power to push now. :-) (I've pushed this particular update already)

Malabarba:
https://youtu.be/_BRv9wGf5pk?t=2m19s

PR #246, opened 2015-08-30
Add track-state middleware

Malabarba:
This middleware keeps the client notified of what the current repl type is.
This makes it easier for cider.el to chose the right conneciton buffer, since we're now aware of which rpel buffers are cljs even if the user didn't create them with our own commands.

### DIFF
--- a/src/cider/nrepl/middleware/track_state.clj
+++ b/src/cider/nrepl/middleware/track_state.clj
@@ -1,0 +1,41 @@
+(ns cider.nrepl.middleware.track-state
+  "State tracker for client sessions."
+  {:author "Artur Malabarba"}
+  (:require [cider.nrepl.middleware.util.cljs :as cljs]
+            [clojure.tools.nrepl.middleware :refer [set-descriptor!]]
+            [clojure.tools.nrepl.middleware.interruptible-eval :refer [*msg*]]
+            [clojure.tools.nrepl.misc :refer [response-for]]
+            [clojure.tools.nrepl.transport :as transport])
+  (:import clojure.tools.nrepl.transport.Transport))
+
+(defn assoc-state
+  "Return response with a :state entry assoc'ed."
+  [response msg]
+  (assoc response :state {:repl-type (if (cljs/grab-cljs-env msg) :cljs :clj)}))
+
+(defn make-transport
+  "Return a Transport that defers to `transport` and possibly notifies
+  about the state."
+  [{:keys [^Transport transport] :as msg}]
+  (reify Transport
+    (recv [this] (.recv transport))
+    (recv [this timeout] (.recv transport timeout))
+    (send [this {:keys [status] :as response}]
+      (.send transport (cond-> response
+                         (contains? status :done) (assoc-state msg))))))
+
+(defn wrap-tracker
+  "Middleware that tracks relevant server info and notifies the client."
+  [handler]
+  (fn [{:keys [op] :as msg}]
+    (cond
+      (#{"eval" "load-file"} op) (handler (assoc msg :transport (make-transport msg)))
+      :else (handler msg))))
+
+(set-descriptor!
+ #'wrap-tracker
+ (cljs/requires-piggieback
+  {:expects #{"eval"}
+   :handles
+   {"track-state-middleware"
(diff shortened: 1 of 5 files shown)


src/cider/nrepl/middleware/track_state.clj
```diff
@@ -0,0 +1,55 @@
...
+  "State tracker for client sessions."
+  {:author "Artur Malabarba"}
+  (:require [cider.nrepl.middleware.util.cljs :as cljs]
+            [clojure.tools.nrepl.middleware :refer [set-descriptor!]]
+            [clojure.tools.nrepl.middleware.interruptible-eval :refer [*msg*]]
+            [clojure.tools.nrepl.misc :refer [response-for]]
+            [clojure.tools.nrepl.transport :as transport])
+  (:import clojure.tools.nrepl.transport.Transport))
+
+
```
bbatsov: Kill this extra blank line.

src/cider/nrepl/middleware/track_state.clj
```diff
@@ -0,0 +1,55 @@
...

+(def messages (atom {}))
```
bbatsov: Add some docstring for `messages`.

src/cider/nrepl/middleware/track_state.clj
```diff
@@ -0,0 +1,55 @@
...
+      "init-tracker" (initialize msg)
+      (handler msg))))
+
+(set-descriptor!
+ #'wrap-tracker
+ (cljs/requires-piggieback
+  {:expects #{"eval"}
+   :handles
+   {"track-state-middleware"
+    {:doc "Enhances the `eval` op by notifying the client of the current repl state. Currently, only the repl type (Clojure or ClojureScript) is informed."}}}))
```
bbatsov: In documentation I prefer to capitalise acronyms like REPL.

src/cider/nrepl/middleware/track_state.clj
```diff
@@ -0,0 +1,55 @@
...
+        (notify-state session)))))
+
+(defn initialize
+  "Setup `session` for tracking.
+  Changes will be notified to `msg`."
+  [{:keys [session] :as msg}]
+  (swap! messages assoc session msg)
+  (notify-state session))
+
+
```
bbatsov: Kill this extra blank line.

bbatsov:
A pretty interesting idea! Needs a few tests, but otherwise looks good to me.

src/cider/nrepl/middleware/track_state.clj
```diff
@@ -0,0 +1,55 @@
...
+         (transport/send (:transport msg)))))
+
+(defn make-transport
+  [{:keys [^Transport transport session]}]
+  (reify Transport
+    (recv [this] (.recv transport))
+    (recv [this timeout] (.recv transport timeout))
+    (send [this response]
+      (.send transport response)
+      (when (:done (:status response))
```
cichli: Minor style nit - this could be `(get-in response [:status :done])`, but I would go so far as to do `(contains? (:status response) :done)` to make it clear that we're checking for the presence of `:done` in a set.

cichli:
This is a good solution to the problem, but do we need to force the client to subscribe to get the data back? If we change the state tracking transport to look like this:

```clojure
(defn make-transport
  [{:keys [^Transport transport session] :as msg}]
  (reify Transport
    (recv [this] (.recv transport))
    (recv [this timeout] (.recv transport timeout))
    (send [this {:keys [status] :as response}]
      (.send transport (cond-> response
                         (contains? status :done) (assoc :repl-type (if (cljs/grab-cljs-env msg) :cljs :clj)))))))
```

Then we lose the need to track who's subscribed to what - we just unconditionally return the state data with the `:done` response of any eval, and it's up to the client to decide whether or not to use it. Thoughts?

Malabarba:
@cichli Yes, we can do that too. 

What prompted me to do manual subscription is that I'm planning on using this to also keep track of larger amounts of data. But I should cross that bridge when I get to it.

Malabarba:
Changes done. I'll open a PR for the Emacs side now, but this is safe to merge on its own.

PR #247, opened 2015-08-31
Custom pretty printing

cichli:
Context: clojure-emacs/cider#1179

This PR adds a new `wrap-pprint-fn` middleware, that accepts the set of all printing-related options that we currently use for the eval/stacktrace/test/refresh middlewares, in addition to a new `pprint-fn` option (which must be the name of the function to be used for printing, e.g. `"clojure.pprint/pprint"`). It then constructs a closure that wraps up all the options and the provided printing function, placing it into the request map (again under the `pprint-fn` key) so that subsequent middlewares can just call the function without having to care about all the different options.

Eventually I'd like to provide some alternative printers to `clojure.pprint/pprint` out-of-the-box, but ATM the main candidates ([fipp](https://github.com/brandonbloom/fipp); [puget](https://github.com/greglook/puget); [aprint](https://github.com/razum2um/aprint)) all require at least Clojure 1.6, so I'm holding off on adding them as dependencies for now.

Two places where we could be using this, but aren't yet:

* In the debugger, we call `stacktrace/analyze-causes` in `eval-with-locals` and it would be a bit awkward to pass `pprint-fn` all the way down the stack... @Malabarba thoughts? Just use a dynamic var maybe?

* In the macroexpansion middleware, we call `clojure.pprint/write` directly - I still need to investigate if it makes sense to allow use of a custom printing fn here.

Note that there's a breaking change here - the `right-margin` option has been renamed to `print-right-margin`, for consistency. cc @laurentpetit @tpope

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -8,6 +8,8 @@
 
   :dependencies [[org.clojure/tools.nrepl "0.2.12"]
                  [org.tcrawley/dynapath "0.2.3"]
+                 ^:source-dep [mvxcvi/puget "1.0.0"]
+                 ^:source-dep [fipp "0.6.3"]
                  ^:source-dep [compliment "0.2.5"]
                  ^:source-dep [cljs-tooling "0.1.9"]
                  ^:source-dep [cljfmt "0.3.0"]
--- a/project.clj
+++ b/project.clj
@@ -17,7 +17,7 @@
                  ^:source-dep [org.clojure/tools.namespace "0.2.11"]
                  ^:source-dep [org.clojure/tools.trace "0.7.9"]
                  ^:source-dep [org.clojure/tools.reader "0.10.0"]]
-  :plugins [[thomasa/mranderson "0.4.5"]]
+  :plugins [[thomasa/mranderson "0.4.6"]]
   :exclusions [org.clojure/clojure]
 
   :filespecs [{:type :bytes :path "cider/cider-nrepl/project.clj" :bytes ~(slurp "project.clj")}]
--- a/src/cider/nrepl/middleware/debug.clj
+++ b/src/cider/nrepl/middleware/debug.clj
@@ -2,6 +2,7 @@
   "Expression-based debugger for clojure code"
   {:author "Artur Malabarba"}
   (:require [cider.nrepl.middleware.inspect :refer [swap-inspector!]]
+            [cider.nrepl.middleware.pprint :as pprint]
             [cider.nrepl.middleware.stacktrace :as stacktrace]
             [cider.nrepl.middleware.util.cljs :as cljs]
             [cider.nrepl.middleware.util.inspect :as inspect]
@@ -123,6 +124,11 @@
   "Bound by the `breakpoint` macro to the local &env."
   {})
 
+(def ^:dynamic *pprint-fn*
+  "Bound by the `breakpoint` macro to the pretty-printing function determined by
+  the `wrap-pprint-fn` middleware."
+  nil)
+
 (def print-length (atom nil))
 (def print-level (atom nil))
 
@@ -177,7 +183,7 @@
         (when-not (instance? ThreadDeath root-ex)
           (debugger-send
            {:status :eval-error
-            :causes [(let [causes (stacktrace/analyze-causes e 50 50)]
+            :causes [(let [causes (stacktrace/analyze-causes e *pprint-fn*)]
                        (when (coll? causes) (last causes)))]})))
       nil)))
 
@@ -245,7 +251,8 @@
   Sends a response to the message stored in debugger-message."
   [value coor]
   `(binding [*skip-breaks* (or *skip-breaks* (atom nil))
-             *locals* ~(sanitize-env &env)]
+             *locals* ~(sanitize-env &env)
+             *pprint-fn* (:pprint-fn *msg*)]
      (let [val# ~value]
        (cond
          (skip-breaks? ~coor) val#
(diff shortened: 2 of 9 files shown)


arrdem:
Awesome!

Malabarba:
> This PR adds a new wrap-pprint-fn middleware, [...] so that subsequent middlewares can just call the function without having to care about all the different options.

Nice! I like the idea.

Will this obsolete the current pprint middleware?

> Eventually I'd like to provide some alternative printers to clojure.pprint/pprint out-of-the-box, [...] (fipp; puget; aprint)

One thing you could do currently is check for the presence of any of these libs, and use it if it's available.

This way the user wouldn't have to manually configure anything complicated, they would only need to add the lib to their `:dev` profile `:dependencies`.

> In the debugger, we call stacktrace/analyze-causes in eval-with-locals and it would be a bit awkward to pass pprint-fn all the way down the stack... @Malabarba thoughts? Just use a dynamic var maybe?

We could use a dynamic var to store the `pprint-fn` from the message that instrumented the function, but it would be even easier to just get `(:print-fn *msg*)` at the time of printing.
This would also make more sense, I think.

Would that work?

Malabarba:
Couldn't this middleware also take into consideration the `pprint` option itself?

For instance, we can:

1. Rename `:pprint-fn` to `:print-fn`
2. `wrap-pprint-fn` uses `clojure.core/print` instead of `pprint/pprint` if the message's `pprint` parameter is nil
3. All other middleware can always just use the `:print-fn` for printing, without worrying about deciding whether to pretty-print or not

Would this be useful?

cichli:
> One thing you could do currently is check for the presence of any of these libs, and use it if it's available.

Yeah, this is the way to go. I'll just need to check if their APIs have remained stable for the past few versions, or if we need to sniff the version that is available.

> We could use a dynamic var to store the pprint-fn from the message that instrumented the function, but it would be even easier to just get (:print-fn *msg*) at the time of printing.
> This would also make more sense, I think.

Oh, of course, the debugger relies on eval - so we know that `*msg*` is bound. That definitely makes more sense. Thanks!

> Couldn't this middleware also take into consideration the pprint option itself? ...

Hmm, I like the idea, but I think the `pprint` option only really applies to `eval` right now - the `ex-data` map in the stacktrace display should probably always be pretty-printed. Let's leave it as `pprint-fn` for now, and possibly add a separate `print-fn` in the future if we need it?

One idea I did have is an op that would let you press a button while viewing a stacktrace to either `pprint` or `pr` the map into a result buffer, or onto the kill ring, for pasting into a gist / the REPL / wherever else (guess this would be handy in the inspector too). For that new op, `print-fn` probably makes more sense than `pprint-fn`.

Malabarba:
> Oh, of course, the debugger relies on eval - so we know that *msg* is bound. That definitely makes more sense. Thanks!

Actually, now that you mentioned it, I remember that's not always the case.
Instrumented can be executed outside an eval op, like in a jetty server.

It's not a huge deal, but it means we can't just rely on `*msg*` being set. We have to either: (1) check if it's set and fallback on pr during execution; or (2) get the `:pprint-fn` from the `*msg*` and bind it in a dynamic var (this would go in the `breakpoint` macro, at the same place where we bind `*skip-breaks*` and `*locals*`).
(If you're not sure what I'm talking about here, I can take care of that after we merge this).

I'm fine with both options. The second would be more troublesome, but would ensure we respect `pprint` in more situations.

>> Couldn't this middleware also take into consideration the pprint option itself? ...

> Hmm, I like the idea, but I think the pprint option only really applies to eval right now - the ex-data map in the stacktrace display should probably always be pretty-printed. Let's leave it as pprint-fn for now, and possibly add a separate print-fn in the future if we need it?

Makes sense. :+1:

bbatsov:
I guess this should be rebased and merged.

bbatsov:
@cichli ping :-)

cichli:
@bbatsov pong :-). Sorry for letting this one languish - I'm finishing off the elisp bits for this now.

bbatsov:
Roger that. :-)

On Thursday, 3 December 2015, Michael Griffiths <notifications@github.com>
wrote:

> @bbatsov <https://github.com/bbatsov> pong :-). Sorry for letting this
> one languish - I'm finishing off the elisp bits for this now.
>
> —
> <https://github.com/clojure-emacs/cider-nrepl/pull/247#issuecomment-161620968>
> .
>


-- 
Best Regards,
Bozhidar Batsov

http://www.batsov.com

bbatsov:
I see some odd test failures. Did someone look into them?

expez:
@benedekfazekas seems mranderson is failing in some of these test runs

benedekfazekas:
ack. will check

cichli:
Looks like all of the JDK6 builds are failing on master? https://travis-ci.org/clojure-emacs/cider-nrepl/builds/94641627

benedekfazekas:
A possible workaround could be to tell mranderson to skip java class repackaging: `source-deps :skip-javaclass-repackage true`

context
cider currently does not use any dependency which has java class files in them so this feature is not really needed anyway. and it seems that sometimes (perhaps under java6?!) the zip output stream freaks out on this. tbh mranderson should nicely detect this and skip this step automatically.

bbatsov:
> tbh mranderson should nicely detect this and skip this step automatically.

Yeah, that seems like the smart thing to do.

benedekfazekas:
yup I created an issue so I don't forget but the workaround should unblock the PR

bbatsov:
I removed the support for Clojure 1.5 and 1.6 (and Java 6). Now you can finally wrap this PR.

bbatsov:
Seems that some eastwood check is breaking the build.

bbatsov:
The readme also has to be updated.

Malabarba:
The Eastwood failure is still about the data-readers file.

bbatsov:
We should exclude this file or something. Or ignore the eastwood failures for now.

bbatsov:
:+1:

PR #249, opened 2015-09-03
Track namespace info as part of track-state

Malabarba:
Goes with https://github.com/clojure-emacs/cider/pull/1301

The track-state middleware now keeps the client informed of loaded
namespaces as well. For each namespace, we transmit a map of:

- All interns (from symbol to the var's metadata). Only select metadata
  keys are transmitted, currently that's:
  - `:indent`
  - `:cider-instrumented`
  - `:macro`
  - `:arglists`
  - `:test`
- All refers not from clojure.core (from symbol to fully-qualified var name)
- All aliases (from symbol to namespace name)

### DIFF
--- a/src/cider/nrepl/middleware/debug.clj
+++ b/src/cider/nrepl/middleware/debug.clj
@@ -1,16 +1,16 @@
 (ns cider.nrepl.middleware.debug
   "Expression-based debugger for clojure code"
   {:author "Artur Malabarba"}
-  (:require [clojure.tools.nrepl.transport :as transport]
-            [clojure.tools.nrepl.middleware :refer [set-descriptor!]]
-            [clojure.tools.nrepl.misc :refer [response-for]]
-            [clojure.tools.nrepl.middleware.interruptible-eval :refer [*msg*]]
-            [cider.nrepl.middleware.util.instrument :as ins]
-            [cider.nrepl.middleware.inspect :refer [swap-inspector!]]
+  (:require [cider.nrepl.middleware.inspect :refer [swap-inspector!]]
             [cider.nrepl.middleware.stacktrace :as stacktrace]
             [cider.nrepl.middleware.util.cljs :as cljs]
             [cider.nrepl.middleware.util.inspect :as inspect]
+            [cider.nrepl.middleware.util.instrument :as ins]
             [cider.nrepl.middleware.util.misc :as misc]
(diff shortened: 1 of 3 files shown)


bbatsov:
At a pretty cursory glance the code looks OK. The last two commits should be squash together.

bbatsov:
I'm also worried about some potential performance implications for namespaces with a ton of interned symbols, but I guess there's not much we can do about this (unless we only submit some delta, which would probably complicate the code a lot).

Malabarba:
> I'm also worried about some potential performance implications for namespaces with a ton of interned symbols

I was worried about that at first, but it really shouldn't be a problem on the nrepl side. Generating a map of *all* namespaces and printing it takes barely a millisecond on my system, so it should be fast enough even on very slow systems (like an android device).

If there's any performance performance issue, it would have to be on the Emacs side. But @cichli made a good point that the middleware for completion candidates frequently sends messages with pretty large content, and nobody seems to complain it's slow.

bbatsov:
> If there's any performance performance issue, it would have to be on the Emacs side. But @cichli made a good point that the middleware for completion candidates frequently sends messages with pretty large content, and nobody seems to complain it's slow.

Yeah, I'm worried solely about decoding this data on Emacs's side, but let's assume we're lucky and this won't be a problem. A side problem is that this data might dominate the nREPL log, which would limit it's usefulness, but we can think about this additionally. 

Address my remarks and let's merge this.

Malabarba:
> A side problem is that this data might dominate the nREPL log,

Good point. I'll look into turning this into a button on the log buffer, so you'd have a clickable ellipsis which expands into the whole thing.

Malabarba:
OK. I think it's ready.
Finally managed to add cljs support too.

bbatsov:
:+1:

bbatsov:
```
~/p/cider-nrepl git:master ❯❯❯ ./build.sh deploy clojars
Could not find artifact cider:cider-nrepl:jar:0.10.0-20150910.054519-24 in clojars (https://clojars.org/repo/)
This could be due to a typo in :dependencies or network issues.
If you are behind a proxy, try setting the 'http_proxy' environment variable.
FAILED
```

@Malabarba can you try pushing to clojars yourself?

Malabarba:
@bbatsov done! :+1:

Malabarba:
@bbatsov done! :+1:

PR #253, opened 2015-09-15
Rewrite the namespace part of track-state

Malabarba:
- Don't build the namespace map ourselves, just use ns-map.
- Don't track changes in namespaces from jar files.
- Since the client still needs to know what's in these jar namespaces,
  we send their info on the first message, but we ONLY do that for
  namespaces that are :require'd by a source namespace. This guarantees
  we only send namespaces that are potentially useful, and reduces the
  amount of garbage we send on the first message

This fixes a lag of about 300ms between receiving eval result and
receiving the "done" status, which translated as a lag on printing the
repl prompt.

This might solve clojure-emacs/cider#1320, depending on the size of
project namespaces, but we should probably still offer a way to disable
it.

### DIFF
--- a/src/cider/nrepl/middleware/track_state.clj
+++ b/src/cider/nrepl/middleware/track_state.clj
@@ -1,11 +1,18 @@
 (ns cider.nrepl.middleware.track-state
   "State tracker for client sessions."
   {:author "Artur Malabarba"}
-  (:require [cider.nrepl.middleware.util.cljs :as cljs]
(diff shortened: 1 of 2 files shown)


bbatsov:
:+1:

bbatsov:
Btw, a further optimization might be not to send all the vars unless the client actually needs them (depending on the config option for the font-locking). We can just add some command users have to run if they change the config.

Malabarba:
Fix clojure-emacs/cider#1321.

PR #255, opened 2015-09-16
[Fix clojure-emacs/cider#1320] Do state tracking in a future

Malabarba:
This should fix all lags associated with the new namespace tracking,
both the long delay at first connection and the ~1sec lag in the REPL.

### DIFF
--- a/src/cider/nrepl/middleware/track_state.clj
+++ b/src/cider/nrepl/middleware/track_state.clj
@@ -7,7 +7,9 @@
             [cljs-tooling.util.analysis :as cljs-ana]
             [clojure.java.classpath :as cp]
             [clojure.tools.namespace.find :as ns-find]
-            [clojure.tools.nrepl.middleware :refer [set-descriptor!]])
+            [clojure.tools.nrepl.middleware :refer [set-descriptor!]]
+            [clojure.tools.nrepl.misc :refer [response-for]]
+            [clojure.tools.nrepl.transport :as transport])
   (:import clojure.lang.Namespace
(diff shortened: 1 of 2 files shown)


bbatsov:
Won't an agent be a better fit for this?

Malabarba:
Is it ok if the function used to update agents in not pure?

expez:
I think you're risking retries when using agents, but if the only IO is reading from disk then that should be fine.

bbatsov:
I think agents never trigger retries, so this should not be a concern. After all people are implementing log systems and things like this using agents. In a transaction an agent won't be executed until the transaction commits, but this is different.

Malabarba:
I'm running some tests here, and it looks like agents do not trigger retries indeed.

    (def at (agent 0))
    (def c (atom 0))
    (do (dotimes [x 1000]
          (future (Thread/sleep 200)
                  (send at #(do (swap! c inc)
                                (inc %)))))
        (println @c))


This always prints `1000`. Whereas, if I replace `(agent 0)` with `(atom 0)` and replace `send` with `swap!` then I always get at least `1030`.

Not a super rigorous test, but it's convinced me. :)

Malabarba:
https://clojurians.slack.com/archives/clojure/p1442501641001191

expez:
I'm not on slack but I just looked at the implementation and you guys are indeed right.  There are no retries because the agent holds an `ActionQueue` internally.

bbatsov:
So, it seems we have clarity here. Agent it is! :-)

bbatsov:
Apart from the failing tests the code looks good to me.

PR #257, opened 2015-09-18
New middleware: out.clj

Malabarba:
Automatically changes the root binding of *out* to print to any active
connections. An active connection is a session that has sent at least
one "eval" op and hasn't been closed.
We use an eval message, instead of the clone op, because there's no
guarantee that the message that sent the clone op will properly handle
output replies.

https://github.com/clojure-emacs/cider/issues/1328

### DIFF
--- a/src/cider/nrepl/middleware/out.clj
+++ b/src/cider/nrepl/middleware/out.clj
@@ -1,0 +1,95 @@
+(ns cider.nrepl.middleware.out
+  "Change *out* to print on sessions in addition to process out.
+
+  Automatically changes the root binding of *out* to print to any
+  active sessions. An active session is one that has sent at least one
+  \"eval\" op.
+
+  We use an eval message, instead of the clone op, because there's no
+  guarantee that the channel that sent the clone message will properly
+  handle output replies."
+  (:require [cider.nrepl.middleware.util.cljs :as cljs]
+            [clojure.tools.nrepl.middleware :refer [set-descriptor!]]
+            [clojure.tools.nrepl.middleware.interruptible-eval :as ie]
+            [clojure.tools.nrepl.middleware.session :as session])
+  (:import [java.io PrintWriter Writer]))
+
+;;; OutStream
+(defonce original-out *out*)
+
+(declare tracked-sessions-map)
+
+(defmacro with-out-binding
+  "Run body with v bound to the output stream of each msg in msg-seq.
+  Also run body with v bound to `original-out`."
+  [[v msg-seq] & body]
+  `(do (let [~(with-meta v {:tag Writer}) original-out]
+         ~@body)
+       (doseq [{:keys [~'session] :as ~'msg} ~msg-seq]
+         (let [~(with-meta v {:tag Writer}) (get @~'session #'*out*)]
+           (try (binding [ie/*msg* ~'msg]
+                  ~@body)
+                ;; If a channel is faulty, dissoc it.
+                (catch Exception ~'e
+                  (swap! tracked-sessions-map dissoc
+                         (:id (meta ~'session)))))))))
+
+(defn fork-out
+  "Returns a PrintWriter suitable for binding as *out* or *err*. All
+  operations are forwarded to all output bindings in the sessions of
+  messages in addition to the server's usual PrintWriter (saved in
+  `original-out`)."
+  [messages]
+  (PrintWriter. (proxy [Writer] []
+                  (close [] (.flush ^Writer this))
+                  (write
+                    ([x]
+                     (with-out-binding [out messages]
+                       (.write out x)))
+                    ([x ^Integer off ^Integer len]
+                     (with-out-binding [out messages]
+                       (.write out x off len))))
+                  (flush []
+                    (with-out-binding [out messages]
+                      (.flush out))))
+                true))
+
+;;; Known eval sessions
+(def tracked-sessions-map
+  "Map from session ids to eval `*msg*`s.
+  Only the most recent message from each session is stored."
+  (atom {}))
+
+(defn tracked-sessions-map-watch [_ _ _ new-state]
+  (let [o (fork-out (vals new-state))]
+    ;; FIXME: This won't apply to Java loggers unless we also
+    ;; `setOut`, but for that we need to convert a `PrintWriter` to a
+    ;; `PrintStream` (or maybe just not use a `PrintWriter` above).
+    ;; (System/setOut (PrintStream. o))
+    (alter-var-root #'*out* (constantly o))))
+
+(add-watch tracked-sessions-map :update-out tracked-sessions-map-watch)
+
+(defn maybe-register-session
+  "Add msg to `tracked-sessions-map` if it is an eval op."
+  [{:keys [op session] :as msg}]
+  (try
+    (when (= op "eval")
+      (when-let [session (:id (meta session))]
+        (swap! tracked-sessions-map assoc session
+               (select-keys msg [:transport :session :id]))))
+    (catch Exception e nil)))
+
+(defn wrap-out [handler]
+  (fn [msg]
+    (maybe-register-session msg)
+    (handler msg)))
+
+(set-descriptor!
+ #'wrap-out
+ (cljs/expects-piggieback
+  {:requires #{#'session/session}
+   :expects #{"eval"}
+   :handles
+   {"out-middleware"
+    {:doc "Change #'*out* so that it also prints to active sessions, even outside an eval scope."}}}))
--- a/README.md
+++ b/README.md
@@ -44,6 +44,7 @@ middleware to `:nrepl-middleware` under `:repl-options`.
                   cider.nrepl.middleware.stacktrace/wrap-stacktrace
                   cider.nrepl.middleware.test/wrap-test
                   cider.nrepl.middleware.trace/wrap-trace
+                  cider.nrepl.middleware.out/wrap-out
                   cider.nrepl.middleware.undef/wrap-undef]}
 
 ```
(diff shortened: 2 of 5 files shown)


src/cider/nrepl/middleware/out.clj
```diff
@@ -0,0 +1,86 @@
...
+;;; Known eval sessions
+(def eval-messages
+  "Map from session ids to eval `*msg*`s.
+  Only the most recent message from each session is stored."
+  (atom {}))
+
+(defn eval-messages-watch [_ _ _ new-state]
+  (let [o (fork-out (vals new-state))]
+    ;; FIXME: This won't apply to Java loggers unless we also setOut.
+    ;; (System/setOut (PrintStream. o))
```
expez: What's the reason for postponing this part?

src/cider/nrepl/middleware/out.clj
```diff
@@ -0,0 +1,86 @@
...
+            [clojure.tools.nrepl.middleware.session :as session])
+  (:import [java.io PrintWriter Writer]))
+
+;;; OutStream
+(defonce original-out *out*)
+
+(declare eval-messages)
+
+(defmacro do-out-messages
+  "Run body with v bound to the output stream of each msg in msg-seq.
```
expez: It's customary to use the `with-` prefix for macros which set up new execution contexts.

src/cider/nrepl/middleware/out.clj
```diff
@@ -0,0 +1,86 @@
...
+  Also run body with v bound to `original-out`."
+  [[v msg-seq] & body]
+  `(do (let [^Writer ~v original-out]
+         ~@body)
```
expez: This is executing the eval requests twice, right?  Can't we just eval once while capturing out and then write whatever we captured to the two other sinks?

src/cider/nrepl/middleware/out.clj
Malabarba: Because you can't directly convert a PrintWriter to a PrintStream.

src/cider/nrepl/middleware/out.clj
Malabarba: No. The body here is just write or flush. See the usage below. 
We have to bind the msg because the PrintWriters created by nrepl rely on that. It's not because we're going to eval the message or anything.

src/cider/nrepl/middleware/out.clj
```diff
@@ -0,0 +1,86 @@
+(ns cider.nrepl.middleware.out
+  "Change *out* to print on connections in addition to process out.
+
+  Automatically changes the root binding of *out* to print to any
```
bbatsov: Won't it be simpler to just send some message from the sessions we care about when we create them (e.g. `:init-output`). It might be nice to have a way to "unsubscribe" as well.

src/cider/nrepl/middleware/out.clj
Malabarba: And it's customary to use do for macros that run its body multiple times, like doseq or dotimes. So how about `do-with-out-messages`? :-)

src/cider/nrepl/middleware/out.clj
Malabarba: I just figured the less book-keeping the better.

bbatsov: Fair enough. Let's go with your approach.

src/cider/nrepl/middleware/out.clj
```diff
@@ -0,0 +1,86 @@
...
+  (PrintWriter. (proxy [Writer] []
+                  (close [] (.flush ^Writer this))
+                  (write [& [x ^Integer off ^Integer len]]
+                    (do-out-messages [out messages] (.write out x off len)))
+                  (flush []
+                    (do-out-messages [out messages] (.flush out))))
+                true))
+
+;;; Known eval sessions
+(def eval-messages
```
bbatsov: I find this name a bit confusing, but I can't propose a great alternative. Perhaps something along the lines of `tracked-sessions-map`?

src/cider/nrepl/middleware/out.clj
```diff
@@ -0,0 +1,86 @@
+(ns cider.nrepl.middleware.out
+  "Change *out* to print on connections in addition to process out.
+
+  Automatically changes the root binding of *out* to print to any
+  active connections. An active connection is a session that has sent
```
bbatsov: Mixing up connection and session is a bit confusing IMO. Maybe we should talk only about sessions here.

bbatsov:
The code looks good to me. You'll have to mention this middleware in the README I guess, even though it doesn't provide any new ops. And when squashing the commits you should update the text of the first commit.

bbatsov:
One more thing - some tests for this would be nice to have. Maybe something printing output from a future or something. Not sure what's the best way to test it.

src/cider/nrepl/middleware/out.clj
```diff
@@ -0,0 +1,87 @@
...
+
+;;; OutStream
+(defonce original-out *out*)
+
+(declare tracked-sessions-map)
+
+(defmacro do-with-out-messages
+  "Run body with v bound to the output stream of each msg in msg-seq.
+  Also run body with v bound to `original-out`."
+  [[v msg-seq] & body]
```
bbatsov: Btw, what does `v` stand for?

bbatsov:
Where are we with this?

Malabarba:
Readme and tests missing. Haven't started on them yet.

src/cider/nrepl/middleware/out.clj
```diff
@@ -0,0 +1,87 @@
...
+
+(defn fork-out
+  "Returns a PrintWriter suitable for binding as *out* or *err*. All
+  operations are forwarded to all output bindings in the sessions of
+  messages in addition to the server's usual PrintWriter (saved in
+  `original-out`)."
+  [messages]
+  (PrintWriter. (proxy [Writer] []
+                  (close [] (.flush ^Writer this))
+                  (write [& [x ^Integer off ^Integer len]]
```
gtrak: This `&`  + destructuring seems extraneous and I think is responsible for a stacktrace. Well, removing & and the destructuring makes things work :-) :

```
ERROR [2015-09-18 22:03:06,794] com.yammer.dropwizard.jersey.LoggingExceptionMapper: Error handling a request: 8cf72e1a7bae9e6a
! java.lang.IllegalArgumentException: No matching method found: write for class java.io.OutputStreamWriter
! at clojure.lang.Reflector.invokeMatchingMethod(Reflector.java:80)
! at clojure.lang.Reflector.invokeInstanceMethod(Reflector.java:28)
! at cider.nrepl.middleware.out$fork_out$fn__20648.doInvoke(out.clj:45) ~[na:na]
! at clojure.lang.RestFn.invoke(RestFn.java:423)
! at cider.nrepl.middleware.out.proxy$java.io.Writer$ff19274a.write(Unknown Source) ~[na:na]
! at java.io.PrintWriter.write(PrintWriter.java:405) ~[na:1.8.0_45-internal]
! at java.io.PrintWriter.append(PrintWriter.java:1063) ~[na:1.8.0_45-internal]
! at java.io.PrintWriter.append(PrintWriter.java:56) ~[na:1.8.0_45-internal]
! at clojure.core$pr.doInvoke(core.clj:3563)
! at clojure.lang.RestFn.applyTo(RestFn.java:139)
! at clojure.core$apply.invoke(core.clj:630)
! at clojure.core$prn.doInvoke(core.clj:3593)
! at clojure.lang.RestFn.applyTo(RestFn.java:137)
! at clojure.core$apply.invoke(core.clj:630)
```

I'm using this impl of fork-out:
```
(defn fork-out
  "Returns a PrintWriter suitable for binding as *out* or *err*. All
  operations are forwarded to all output bindings in the sessions of
  messages in addition to the server's usual PrintWriter (saved in
  `original-out`)."
  [messages]
  (PrintWriter. (proxy [Writer] []
                  (close [] (.flush ^Writer this))
                  (write
                    ([c]
                     (do-with-out-messages [out messages] (.write out c)))
                    ([s ^Integer off ^Integer len]
                     (do-with-out-messages [out messages] (.write out s off len))))
                  (flush []
                    (do-with-out-messages [out messages] (.flush out))))
                true))
```

gtrak: I think it also explains all these build failures: https://travis-ci.org/clojure-emacs/cider-nrepl/builds/81060333

Malabarba: hm... I guess it's not safe to call `(.write out s off len)` on the system's standard out. I'll fix it, thanks.

gtrak: The reflector can't find the right java method when off/len are null in the case that the 1-ary version of .write was called.  I think that's because they're primitive arguments, even if it could find it, it would fail to cast a null to a primitive.  It might be jvm-version-dependent as well.

Malabarba:
Thanks @gtrak, I believe the exception is fixed now.

I've added a readme entry and a test as well. The test is terribly superficial, but my knowledge of this whole Writer/Print/Stream family of Java inheritance is very limited, so I can't do much better.
I'd be very thankful if someone could write some proper tests for the `fork-out` function.

bbatsov:
Guess this will do for now. Squash the commits and we'll refine the code down the road.

src/cider/nrepl/middleware/out.clj
```diff
@@ -0,0 +1,95 @@
...
+  messages in addition to the server's usual PrintWriter (saved in
+  `original-out`)."
+  [messages]
+  (PrintWriter. (proxy [Writer] []
+                  (close [] (.flush ^Writer this))
+                  (write
+                    ([x]
+                     (with-out-binding [out messages]
+                       (.write out x)))
+                    ([x ^Integer off & [^Integer len]]
```
gtrak: I still don't get why you need this optional argument, there's only arity 1 and 3 .write methods in http://docs.oracle.com/javase/7/docs/api/java/io/Writer.html, so len will never not be null if it gets called.

Malabarba: > I still don't get why you need this argument as optional, there's only arity 1 and 3 .write methods in Writer

Because I didn't know that. :)

Malabarba:
Ok. I believe the failing tests are the same ones that fail on master

PR #261, opened 2015-09-25
don't override apple.awt.UIElement system property if explicitly set.

phronmophobic:
Currently, the system property "apple.awt.UIElement" is unconditionally to "true" set right before the system property is used to decided whether or not the application will show an application icon in the osx dock. There is no way to override this behavior when using the repl. This commit only sets "apple.awt.UIElement" to "true" when the property is not already set via jvm flags or otherwise.

I'm working on a swing application where I would actually like the default behavior when apple.awt.UIElement is not set to true or is left unset. Basically, I like being able to command-tab between the GUI I'm working on and emacs where I'm making edits to the GUI.


bbatsov:
I'm OK with the change.

You should update the matching comment to reflect it and update the commit message itself to match the project's style (capitalized first word, not `.` at the end of the title).

phronmophobic:
Great! Updated the commit message and the comment. Thanks!

expez:
Thanks! :+1:

PR #262, opened 2015-09-25
Extract more detail from compiler exceptions

mainej:
This patch re-introduces the inner 'cause' of compiler exceptions.

These exceptions are not always informative, which may be why they were
hidden in 0a57231f1. For example when the compiler is complaining about
a syntax error, the exception often refers to a line number far from the
actual syntax problem.

However, the exceptions *can* be quite useful, for example when they
complain about an undeclared variable.

This patch re-introduces this additional detail. I'd be equally happy
with completely removing `extract-location`.  The output from the two
approaches is similar, and perhaps less code is better.

### DIFF
--- a/src/cider/nrepl/middleware/stacktrace.clj
+++ b/src/cider/nrepl/middleware/stacktrace.clj
@@ -100,11 +100,11 @@
 
 (defn extract-location
   "If the cause is a compiler exception, extract the useful location information
-  from its message and discard the string representation of its inner cause."
+  from its message."
   [{:keys [class] :as cause}]
   (if (= class "clojure.lang.Compiler$CompilerException")
     (update-in cause [:message] str/replace
-               #".* (compiling:)\((.*)\)" "Error $1 $2")
+               #".*?: (.*?), (compiling:)\((.*)\)" "Error $2 $3 $1")
     cause))
 
 ;; CLJS REPLs use :repl-env to store huge amounts of analyzer/compiler state
--- a/test/clj/cider/nrepl/middleware/stacktrace_test.clj
+++ b/test/clj/cider/nrepl/middleware/stacktrace_test.clj
@@ -25,11 +25,13 @@
 (def form1 '(throw (ex-info "oops" {:x 1} (ex-info "cause" {:y 2}))))
 (def form2 '(do (defn oops [] (+ 1 "2"))
                 (oops)))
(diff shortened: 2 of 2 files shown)


src/cider/nrepl/middleware/stacktrace.clj
```diff
@@ -104,7 +104,7 @@
   [{:keys [class] :as cause}]
   (if (= class "clojure.lang.Compiler$CompilerException")
     (update-in cause [:message] str/replace
-               #".* (compiling:)\((.*)\)" "Error $1 $2")
+               #".*?: (.*?), (compiling:)\((.*)\)" "Error $2 $3 $1")
```
Malabarba: Isn't `.*?` the same as `.*`?

mainej:
Looks like all the Travis builds have been failing recently... This commit fails the same part of the build matrix that other recent commits fail.

src/cider/nrepl/middleware/stacktrace.clj
mainej: Nope, slightly different [1].  In this case, you would probably be safe with either variation, but I'm being paranoid.

```
user=> (clojure.string/replace-first "a:b c:d" #"(.*):(.*)" "$1x$2")
"a:b cxd"
user=> (clojure.string/replace-first "a:b c:d" #"(.*?):(.*)" "$1x$2")
"axb c:d"
```

[1] http://www.regular-expressions.info/repeat.html#lazy

expez:
This looks good to me.  It would be nice to have a test for this, though.

> Looks like all the Travis builds have been failing recently... 

Yes, this is quite embarrassing :/

mainej:
Passing tests added, though Travis is still failing in part of the build matrix.

jeffh:
Any updates on this PR? I'd love this to be merged in :)

bbatsov:
The commits have to be squashed together. Apart from that it looks ok.

mainej:
Commits squashed, though Travis is still failing in part of the build matrix.

bbatsov:
Did you rebase your code on top of the current `master`. The tests are no longer failing there.

mainej:
Ah, rebased... tests now passing.

expez:
Thanks @mainej :+1:

bbatsov:
@jeffvalk btw, why did you opt to filter those out?

PR #268, opened 2015-10-17
Abbreviate printing of some objects

Malabarba:
Instead of printing as
`  #object[clojure.core$_PLUS_ 0x4e648e99 "clojure.core$_PLUS_@4e648e99"]`
they are now printed as
`  #function clojure.core/+`

### DIFF
--- a/src/cider/nrepl.clj
+++ b/src/cider/nrepl.clj
@@ -1,5 +1,6 @@
 (ns cider.nrepl
   (:require [clojure.tools.nrepl.server :as nrepl-server]
+            [cider.nrepl.print-method]
             [cider.nrepl.middleware.apropos]
             [cider.nrepl.middleware.classpath]
             [cider.nrepl.middleware.complete]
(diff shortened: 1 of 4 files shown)


bbatsov:
That's a welcome change, but it will probably be more valuable if we apply it to eval results in general - by this I mean changing the eval results in case when something returns a var. I mentioning this, because of we update it only in the output there would inconsistency between the output and the eval results.

Malabarba:
eval results are converted to strings via the print-method method. So they are already affected by this.

bbatsov:
> eval results are converted to strings via the print-method method. So they are already affected by this.

Good to know. So, this should probably be out of the `out` middleware then. I'd normally expect to have only stuff directly related to the middleware there (which prompted my original comment).

Malabarba:
Where do you think it could go? 
If I  create a new file for it, I think I'll have to do some configuring to make sure it's loaded.

bbatsov:
Not sure actually. My biggest concern is that some people might want to turns off the middleware without the custom printing or vice versa and as it stands - they can't do so.

Malabarba:
I think I'll put it in a file called `print-method.clj` *outside* the `middleware` directory (since it's actually not a middleware). And I'll define a variable for disabling this.

bbatsov:
Sounds good.

Malabarba:
Done. ADded tests and method support

Malabarba:
Alright if I merge this?

bbatsov:
Allow me. :-)

PR #278, opened 2015-12-18
Don't inject the middleware if its version requirements aren't met

### LINKED ISSUE #277: Check for Clojure and Leiningen version compatibility in the plugin
If Clojure < 1.7 or Lein < 2.5.2, log a warning and don't add the dependency or middleware to the project.

cichli:
Fixes #277.

The REPL will still start, and the client will warn about all the missing ops when it connects. Client-side, we should probably also check the Clojure version (using the `describe` op response), and perhaps sniff the `nrepl-server` buffer for the Leiningen warning. Then we can show a more meaningful warning in the REPL, since I don't think people tend to look at `nrepl-server` that often. That obviously only applies to `cider-jack-in` users - hopefully those connecting to a REPL running at the terminal will see it there :-).

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -45,7 +45,9 @@
                                                      cider.nrepl.middleware.trace/wrap-trace
                                                      cider.nrepl.middleware.track-state/wrap-tracker
                                                      cider.nrepl.middleware.undef/wrap-undef]}
-                   :dependencies [[org.clojure/tools.nrepl "0.2.12"]]}
+                   :dependencies [[org.clojure/tools.nrepl "0.2.12"]
+                                  ;; For developing the Leiningen plugin.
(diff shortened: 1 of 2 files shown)


Malabarba:
LGTM. But maybe the warning could say something simpler, like: cider-nrepl will not be activated (present?) in this REPL.

bbatsov:
Looks good to me as well. As for the message - I'm fine either way.

expez:
This is brilliant.  We need exactly this for refactor-nrepl too.

//cc @benedekfazekas

cichli:
@Malabarba agreed; I've reduced it to just `cider-nrepl will not be included in your project`.

PR #284, opened 2016-01-09
Add print method and tests for atoms

seanirby:
Tests pass.  I've pasted my comment from the issue thread below for reference.

Current print-method for an Atom is here:
https://github.com/clojure/clojure/blob/bc186508ab98514780efbbddb002bf6fd2938aee/src/clj/clojure/core_print.clj#L387-L408

Which looks like:
```
=> (println (atom {:foo :bar}))
#object[clojure.lang.Atom 0x54274a2b {:status :ready, :val {:foo :bar}}]
```
It looks like ':status' will always be ':ready' for an atom since atoms do not implement the "IPending" interface.  I propose we drop this information in the new print method since it's not useful.  New print output looks like the following:

```
=>(println (atom "{:foo :bar}"))
#atom[{:foo :bar} 0x54274a2b]
```

The identity hash code is moved to the end to keep it consistent with the print output for multifn.

### DIFF
--- a/src/cider/nrepl/print_method.clj
+++ b/src/cider/nrepl/print_method.clj
@@ -1,6 +1,6 @@
 (ns cider.nrepl.print-method
   (:require [clojure.string :as s])
-  (:import [clojure.lang AFunction MultiFn Namespace]
+  (:import [clojure.lang AFunction Atom MultiFn Namespace]
            java.io.Writer))
 
 ;; Extending `print-method` defined in clojure.core, to provide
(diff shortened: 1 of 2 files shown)


bbatsov:
:+1: Thanks!

Malabarba:
Thanks for this!

>  I've pasted my comment from the issue thread below for reference.

For future record. I usually like to paste comments on the commit message too, or at least paste a link to the discussion.

src/cider/nrepl/print_method.clj
```diff
@@ -25,6 +25,13 @@
        (do ~@(map #(list '.write 'w %) strings))
        (#'clojure.core/print-object ~arg ~'w))))
 
+;;; Atoms
+;; Ex: #atom[{:foo :bar} 0x54274a2b]
+(def-print-method Atom c
+  "#atom["
+  (print-str @c)
```
Malabarba: This should be `pr-str`, otherwise strings will print wrong (as you can see in the test).

seanirby:
@Malabarba,

I did notice that but I assuming that was default behavior since I was testing existing behavior using 'println'.  I guess I didn't understand the subtleties between the different prints.  Sorry about that! I'll update this soon using 'pr-str'.

PR #285, opened 2016-01-11
Use 'pr-str' for getting a derefed atom string.

seanirby:
Previously, atoms containing strings would print incorrectly.  This way atom with strings will print correctly with both 'pr' and 'print'


Malabarba:
Drop the final period on the commit message and we're good to merge. 👍

seanirby:
Done, thanks for looking

PR #288, opened 2016-01-25
set version to 0.10.2

cap10morgan:


cap10morgan:
Oops, let me redo that from the right branch.

PR #291, opened 2016-01-30
Implement a Trace command in the debugger

Malabarba:
This command skips all further breakpoints (like :continue), but prints
their values via println.


Malabarba:
Tests fixed. I'll update the readme and changelog on the other repo when we merge this.

bbatsov:
:+1: LGTM

PR #296, opened 2016-02-05
New feature: Enlighten

Malabarba:
This adds a middleware that instruments code in “eval” ops that provide an “enlighten” key. When the code is evaluated, information is sent through the debug channel regarding value of local variables and return value of functions.

In addition to the “enlighten” key (which is made available by Emacs as a minor mode), this feature can also be accessed by adding `#light` before a top-level sexp.


Malabarba:
Cider side is here: https://github.com/clojure-emacs/cider/pull/1545

bbatsov:
The code looks good to me. Just mention the new middleware in the README.

PR #297, opened 2016-02-05
Implement a print-method for agents, futures, delays, promises, etc

Malabarba:
More specifically, clojure.lang.IDeref. This would include atoms too,
but we already have a specialized print_method for them.

### DIFF
--- a/src/cider/nrepl/print_method.clj
+++ b/src/cider/nrepl/print_method.clj
@@ -69,3 +69,20 @@
   "#namespace["
   (format "%s" (ns-name c))
   "]")
+
+;;; Agents, futures, delays, promises, etc
+(defn- deref-name [c]
+  (let [class-name (translate-class-name c)]
+    (if-let [[_ short-name] (re-find #"^clojure\.lang\.([^.]+)" class-name)]
+      (.toLowerCase short-name)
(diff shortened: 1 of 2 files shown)


src/cider/nrepl/print_method.clj
```diff
@@ -44,7 +43,11 @@
       (s/replace "_EQ_" "=")
       (s/replace "_SLASH_" "/")
       (s/replace "_STAR_" "*")
-      (s/replace "_" "-"))
+      (s/replace "_" "-")))
```
expez: Is this what `demunge` does?

Malabarba: Looks like it is. :-) I'll change it.

bbatsov:
The code looks good, but I notice we've got a bunch of failing tests.

Malabarba:
I believe the tests are fixed now. There are still some cljfmt issues on master, but we'll get to them in time.

bbatsov:
> I believe the tests are fixed now. There are still some cljfmt issues on master, but we'll get to them in time.

Let's just allow failures for them, as I'm tired of all those false positives (cloverage, cljfmt & eastwood).

Malabarba:
👍

PR #298, opened 2016-02-06
Rewrote inspect-value as a multimethod so developers can write methods for it. 

pdenno:
I rewrote inspect-value as a multimethod so that developers can write methods for it. inspect is already multimethods, but inspect-value was an ordinary function. I have a print-method on a type I defined. With this patch, it uses it, even in collections displayed in the inspector.

### DIFF
--- a/src/cider/nrepl/middleware/util/inspect.clj
+++ b/src/cider/nrepl/middleware/util/inspect.clj
@@ -85,48 +85,64 @@
        s/join
        (format fmt)))
 
-(defn inspect-value
-  ([value]
-   (cond (atom? value) (pr-str value)
+(defn value-types [value]
+  (cond
+    (atom? value) :atom
+    (and (instance? Seqable value) (empty? value)) :seq-empty
+    (and (map? value) (< (count value) 5))         :map
+    (map? value)                                   :map-long
+    (and (vector? value) (< (count value) 5))      :vector
+    (vector? value)                                :vector-long
+    (and (list? value) (< (count value) 5))        :list
+    (list? value)                                  :list-long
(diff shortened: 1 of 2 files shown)


bbatsov:
Makes sense I guess. A few remarks:

* The commit message should not end with a `.`
* There's a bit of trailing whitespace in your commit
* You should probably add a test or two covering the addition of new print functions

src/cider/nrepl/middleware/util/inspect.clj
```diff
@@ -370,3 +386,5 @@
 (defn inspect-print [x]
   (doseq [component (:rendered (inspect-render (fresh) x))]
     (inspect-print-component component)))
+
```
Malabarba: What are these extra lines?

pdenno: A mistake.

bbatsov:
@pdenno Ping :-)

pdenno:
I wrote some tests, got rid of the extra lines at the end of the file.

bbatsov:
Looks good. Squash the two commits together and we're good to go.

pdenno:
Done, I think!

bbatsov:
I still see 3 commits. You have to use an interactive rebase - see this for details https://ariejan.net/2011/07/05/git-squash-your-latests-commits-into-one/

pdenno:
Thanks. I'll try again.

bbatsov:
@pdenno ping :-)

pdenno:
I could not fix this pull request; I started a new one. Sorry for the problems this may have caused.

PR #299, opened 2016-02-08
Add error handling for classpath op

sanjayl:
Wraps  transport with a try/catch block and returns an error
message if there's a problem with getting the classpath. Added
tests for this as well.


bbatsov:
:+1:

Malabarba:
Just checking, but will the handler on the emacs side understand this error message?

bbatsov:
Yeah, it will. This is generic - see `nrepl-send-sync-request`. It doesn't actually check the error type, though.

PR #300, opened 2016-02-09
Add format-code exception & delete right-margin op

sanjayl:
Wrap format-code-reply's send in try/catch.
Deleted unused right-margin option.
Added tests to get to 100% coverage.

### DIFF
--- a/src/cider/nrepl/middleware/format.clj
+++ b/src/cider/nrepl/middleware/format.clj
@@ -12,11 +12,16 @@
 
 (defn format-code-reply
   [{:keys [transport code] :as msg}]
-  (transport/send
-   transport
-   (response-for msg
-                 :formatted-code (fmt/reformat-string code)
-                 :status :done)))
+  (try
+    (transport/send
+     transport
+     (response-for msg
+                   :formatted-code (fmt/reformat-string code)
+                   :status :done))
+    (catch Exception e
+      (transport/send
+       transport
+       (response-for msg (err-info e :format-code-error))))))
 
 (defn- read-edn
   "Returns a vector of EDN forms, read from the string s."
@@ -67,5 +72,6 @@
    "format-edn"
    {:doc "Reformats the given EDN data, returning the result as a string."
     :requires {"edn" "The data to format."}
-    :optional {"right-margin" "The maximum column width of the formatted result."}
+    :optional {"print-right-margin" "The maximum column width of the formatted result."
+               "pprint-fn" "Fully qualified name of the print function to be used."}
     :returns {"formatted-edn" "The formatted data."}}}})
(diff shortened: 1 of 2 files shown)


src/cider/nrepl/middleware/format.clj
```diff
@@ -67,5 +72,4 @@
    "format-edn"
    {:doc "Reformats the given EDN data, returning the result as a string."
     :requires {"edn" "The data to format."}
-    :optional {"right-margin" "The maximum column width of the formatted result."}
```
bbatsov: Did you check the client code as well? Seems there are references to this param there.

bbatsov: //cc @cichli

sanjayl:
You're right, it looks like the client code is setting the right margin; however, I think it uses a different key.

The `cider--nrepl-pprint-request-plist` client-side method sets a key called `print-right-margin`, not `right-margin`. Likewise, the `pprint` middleware looks for the `:print-right-margin` key. The pprint test suite also tests via `:print-right-margin`.

If I'm not mistaken, I should put that optional key back in, but change the string to `"print-right-margin"`. I couldn't find any uses of `right-margin`, but take that with a huge grain of salt because I haven't had a ton of time to verify.

bbatsov:
Guess you're right. Unfortunately I don't have time to closely examine this myself.

Malabarba:
> The cider--nrepl-pprint-request-plist client-side method sets a key called print-right-margin, not right-margin. Likewise, the pprint middleware looks for the :print-right-margin key. The pprint test suite also tests via :print-right-margin.

Could you tell me which call is that (which file/line) so that I can look back in history a bit? I just recently refactored this code and I could have messed up.

bbatsov:
https://github.com/clojure-emacs/cider/blob/277ff6408f81ee699502193c43e8f9fe08c34eb8/cider-client.el#L639 and https://github.com/clojure-emacs/cider/blob/0e3004912b6340ac215a1c3eeba224c3be53e8c4/cider-interaction.el#L1102

sanjayl:
Here's where the change is made from `right-margin` to `print-right-margin` in the client code:
https://github.com/clojure-emacs/cider/commit/a8bf042786d7d054882418e8f4fed9ed648e2250

I'm going through the cider-nrepl code and changing instances of `right-margin` to `print-right-margin`. The one place I'm confused about is in the [pprint-transport](https://github.com/clojure-emacs/cider-nrepl/blob/master/src/cider/nrepl/middleware/pprint.clj#L89-L97). I'm not sure what's going on in this function or the implications of changing the destructuring assignment? Any pointers would be appreciated.

bbatsov:
@cichli ping :-)

bbatsov:
In the absence of further feedback I'll merge this. @sanjayl Thanks!

PR #303, opened 2016-02-26
Conditional breakpoints

grammati:
Not entirely ready, but I would like to get some more eyes on this and see whether others agree with the approach.

The feature is conditional breakpoints - that is, breakpoints that will only enter the debugger when an expression is truthy.

Conditions are attached to forms as `:break/when` metadata. Example:

```clojure
(for [i (range 3)]
  #dbg ^{:break/when (odd? i)
  (inc i))
```

Evaluating this, you will stop in the debugger only once, when `i` is 1.

However, the primary use case (at least for me) is likely to be instrumenting a function with a conditional breakpoint. I have implemented this in cider as `C-u C-u C-M-x`, which prompts for an expression. This can be an expression of the function arguments. The form is then sent to cider-nrepl with `#dbg ^{:break/when (the expression)}` prefixed.

Making this work required a slightly hairy special-case that moves `:break/when` metadata down from the defn form and onto the forms inside the body of the macroexpanded function.

To get some confidence that I have the slightest idea what I'm doing, I wrote a bunch of new tests, in a new test namespace. These interact with cider-nrepl only by sending and receiving nrepl messages, just as a client (i.e. emacs) would.

Thanks everyone, please let me know what you think.

### DIFF
--- a/src/cider/nrepl/middleware/debug.clj
+++ b/src/cider/nrepl/middleware/debug.clj
@@ -79,6 +79,10 @@
       true)
     false))
 
+(defn- seq= [a b]
+  ;; To deal with, eg: (= () nil) => true
+  (= (seq a) (seq b)))
+
 (defn skip-breaks?
   "True if the breakpoint at coordinates should be skipped.
   If the first element of `*skip-breaks*` is :all, return true.
@@ -98,7 +102,7 @@
         :trace  false
         ;; From :out, skip some breaks.
         :deeper (let [parent (take (count skip-coords) coordinates)]
-                  (and (= skip-coords parent)
+                  (and (seq= skip-coords parent)
                        (> (count coordinates) (count parent))))
         ;; From :here, skip some breaks.
         :before (coord< coordinates skip-coords)))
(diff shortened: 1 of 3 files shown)


Malabarba:
Wow. First of all, thanks a ton! :-)
This is a feature that's been requested several times already, so it's very welcome.
Also, thanks so much for writing so many tests! :-)

I'll make some comments directly in the code, but overall I like how you've implemented it.

src/cider/nrepl/middleware/debug.clj
```diff
@@ -139,10 +143,10 @@
 ;;; `wrap-debug` receives an initial message from the client, stores
 ;;; it in `debugger-message`, and `breakpoint` answers it when asking
 ;;; for input.
-(def debugger-message
```
Malabarba: I'm fine with this change, but I'd rather leave it to a separate commit (unless it has some relation to this feature).

project.clj
```diff
@@ -68,4 +68,6 @@
                                          merge-meta [[:inner 0]]}}}
 
              :eastwood {:plugins [[jonase/eastwood "0.2.3"]]
-                        :eastwood {:config-files ["eastwood.clj"]}}})
+                        :eastwood {:config-files ["eastwood.clj"]}}
+
+             :debug {:jvm-opts ["-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=5009"]}})
```
Malabarba: Could you explain what this line is for?

grammati: Ah, yes, unrelated to this PR. I put that in most of my projects so that I can attach Intellij to the running process and debug with Cursive. I can remove itl

grammati:
I pushed a PR for a rudimentary version of the front-end portion, here: https://github.com/clojure-emacs/cider/pull/1591

src/cider/nrepl/middleware/debug.clj
```diff
@@ -346,20 +350,40 @@
...
+  (let [has-condition? (contains? (meta form) :break/when)
+        condition      (:break/when (meta form))
+        parent-coor    (vec (butlast coor))]
+    (if has-condition?
+      `(with-debug-bindings ~extras
+         (let [skip?# (not ~condition)]
+           ;; If there is a condition and it is falsy, we need to skip
+           ;; the current level (:deeper than parent coor), but only
+           ;; once. Next time, we need to test the condition again.
+           (binding [*skip-breaks* (if skip?#
```
Malabarba: Do we need :deeper? Isn't that the same as skipping everything inside this `binding`?

src/cider/nrepl/middleware/debug.clj
```diff
@@ -346,20 +350,40 @@
...
+       ;; depth.
+       (= (first @*skip-breaks*) :trace)
+       (do (print-step-indented ~(count coor) '~original-form val#)
+           val#)
+       ;; Nothing special. Here's the actual breakpoint logic.
+       :else (->> (pr-short val#)
+                  (assoc *extras* :debug-value)
+                  (read-debug-command val#)))))
+
+(defmacro maybe-breakpoint
```
Malabarba: This macro always introduces a `breakpoint`. It might be skipped during execution, but it's always there.

I prefer to call this macro `breakpoint`, and change the name of the macro above to something like `breakpoint-implementation`.

src/cider/nrepl/middleware/util/instrument.clj
```diff
@@ -246,6 +248,36 @@
   ;; Don't use `postwalk` because it destroys previous metadata.
   (walk-indexed #(tag-form %2 breakfunction) form))
 
+(defn transfer-break-conditions
```
Malabarba: This part is a little tricky. Can we leave it out for now? 

Firstly, it shouldn't be here. Doing something like this is the job of `debug.clj`, not `instrument.clj`.

Second, I realise you need this so that your Emacs command will work, but I still want to discuss what's the best interface for this feature.

project.clj
Malabarba: Yes please. :-)

bbatsov: Debug cider in cider. :-)

src/cider/nrepl/middleware/debug.clj
grammati: My thinking at the time was that if the "skipped" form calls another function which is instrumented, we do want to stop in the debugger, and that using `:all` would prevent that. I can add some test cases to verify.

Malabarba: Good point!
And yes, a test case would be nice.

grammati:
@Malabarba, I pushed a commit that addresses all your comments, I think.

I pared back the functionality to be just the basic conditional break. I removed the special-casing of defns that moved the condition down into the body of the fn. I can re-introduce that in another PR, where we can discuss the best interface.

I also added a few more tests, including one (unrelated to conditional breakpoints) that I think should pass but does not. Can you have a look at `call-instrumented-fn-when-stepping-out`? Perhaps my expectations of how the debugger should behave are just wrong.

Thanks!

Malabarba:
Ok, thanks! I'll have a look at the tests when I get a chance. :+1:

grammati:
I rebased this onto master and pushed again

Malabarba:
Ok. Looks like your tests are good. The only 2 failures are the cljs tests which are also failing on master. I've also tested your code locally and it seems to work well.
I'm going to give this a final read and merge. :+1:

Malabarba:
Ok. This should now be on clojar.
Thanks again for the feature and specially for the tests. :-)

PR #304, opened 2016-02-29
Generalize out.clj code to support err

Malabarba:
Related to #1588

### DIFF
--- a/src/cider/nrepl/middleware/out.clj
+++ b/src/cider/nrepl/middleware/out.clj
@@ -17,42 +17,50 @@
 
 ;;; OutStream
 (defonce original-out *out*)
+(defonce original-err *err*)
 
 (declare tracked-sessions-map)
 (declare unsubscribe-session)
 
 (defmacro with-out-binding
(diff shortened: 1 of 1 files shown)


bbatsov:
The changes generally look good. I'd rename this middleware to `output`, so it's clear it's not specific to `out`. I'd also update the ns comment.

src/cider/nrepl/middleware/out.clj
```diff
@@ -23,36 +23,43 @@
...
   "Run body with v bound to the output stream of each msg in msg-seq.
-  Also run body with v bound to `original-out`."
-  [[v msg-seq] & body]
-  `(do (let [~(with-meta v {:tag Writer}) original-out]
+  Also run body with v bound to `original-out`.
+  type is either :out or :err."
+  [[v msg-seq type] & body]
+  `(do (let [~(with-meta v {:tag Writer}) (case ~type
+                                            :out #'original-out
+                                            :err #'original-err)]
```
bbatsov: You forgot to define `original-err`.

bbatsov:
Guess this will do for now.

PR #305, opened 2016-02-29
multimethod for inspect-value - second try

pdenno:
This addresses the same issue as my first botched pull request: I defined inspect-value as a multimethod so that users could write methods for it.

I am still a total klutz with git. The rebase looked OK, but I don't think I understood the issue with "fast forwarding." Thus I found it easier to start over. My apologies.

### DIFF
--- a/src/cider/nrepl/middleware/util/inspect.clj
+++ b/src/cider/nrepl/middleware/util/inspect.clj
@@ -85,48 +85,64 @@
        s/join
        (format fmt)))
 
-(defn inspect-value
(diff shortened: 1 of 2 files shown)


bbatsov:
Looks good. Just change the commit message to "Rewrite inspect-value as a multimethod" or something like this to match our commit message style.

bbatsov:
I've fixed the commit name and merged it manually.

pdenno:
Thanks!

PR #307, opened 2016-03-03
Add tests for info middleware and small correction

sanjayl:

- [x ] The commits are consistent with our [contribution guidelines](CONTRIBUTING.md)
- [x ] You've added tests to cover your change(s)
- [x ] All tests are passing
- [x ] The new code is not generating reflection warnings
- [n/a ] You've updated the readme (if adding/changing middleware)

Thanks!

Added coverage tests for the info middleware to get 100% coverage. There
are some TODO's listed in the test code since I wasn't clear on the
precise intent of certain methods.

There was a small correction made to the format-response function (the
stringified namespace wasn't merged into the response map where it
should have been, but no bug resulted since the stringified namespace
was added in subsequent steps anyway).

Slightly changed the format-nested function from an explicit
for loop since it was difficult to test 100% through the macro generated `for` code. The
new code should be equivelent to the old but it looks a bit clunkier.

### DIFF
--- a/test/clj/cider/nrepl/middleware/info_test.clj
+++ b/test/clj/cider/nrepl/middleware/info_test.clj
@@ -1,8 +1,12 @@
 (ns cider.nrepl.middleware.info-test
-  (:require [clojure.test :refer :all]
+  (:require [clojure.data]
+            [clojure.test :refer :all]
             [clojure.java.io :as io]
             [clojure.repl :as repl]
-            [cider.nrepl.middleware.info :as info]))
+            [clojure.set :as set]
+            [cider.nrepl.middleware.info :as info]
+            [cider.nrepl.middleware.util.misc :as util]
+            [cider.nrepl.test-session :as session]))
 
 (defn file
   [x]
@@ -22,7 +26,8 @@
   (is (= (class (file "test/clj/cider/nrepl/middleware/info_test.clj"))
          java.net.URL))
   (is (relative "clojure/core.clj"))
-  (is (nil? (relative "notclojure/core.clj"))))
+  (is (nil? (relative "notclojure/core.clj")))
+  (is (nil? (info/resource-path "jar:file:fake.jar!/fake/file.clj"))))
 
 (deftest test-boot-resource-path
   (let [tmp-dir-name (System/getProperty "java.io.tmpdir")
@@ -41,7 +46,15 @@
         (is (= (.getPath (file tmp-file-name))
                tmp-file-path))
         (finally
-          (System/clearProperty "fake.class.path"))))))
+          (System/clearProperty "fake.class.path"))))
+    (testing "when classpath is a jar"
+      (let [tmp-jar-path "jar:file:fake/clojure.jar"]
+        (try
+          (System/setProperty "fake.class.path" tmp-jar-path)
+          (is (= (-> (#'cider.nrepl.middleware.info/boot-class-loader) .getURLs last str)
+                 "file:fake/clojure.jar"))
+          (finally
+            (System/clearProperty "fake.class.path")))))))
 
 (deftype T [])
 
@@ -56,7 +69,7 @@
 
   (is (not (info/info-clj 'clojure.core (gensym "non-existing"))))
   (is (not (info/info-clj 'cider.nrepl.middleware.info-test (gensym "non-existing")))
-      "Check that deftype T (which returns nil for .getPackage), doesn't throw")
+      "Check that deftype T (which returns nil for .getPackage), doesn't throw") ;; TODO: This doesn't seem to test Type/Protocol resolution as stated?
 
   (is (= (the-ns 'clojure.core) (:ns (info/info-clj 'cider.nrepl.middleware.info 'str))))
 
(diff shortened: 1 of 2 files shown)


test/clj/cider/nrepl/middleware/info_test.clj
```diff
@@ -56,7 +69,7 @@
 
   (is (not (info/info-clj 'clojure.core (gensym "non-existing"))))
   (is (not (info/info-clj 'cider.nrepl.middleware.info-test (gensym "non-existing")))
-      "Check that deftype T (which returns nil for .getPackage), doesn't throw")
+      "Check that deftype T (which returns nil for .getPackage), doesn't throw") ;; TODO: ???
```
bbatsov: Not sure what the todo stands for here.

test/clj/cider/nrepl/middleware/info_test.clj
```diff
@@ -163,6 +176,215 @@
   (is (re-find #"cider-nrepl" (:file (#'info/maybe-add-file {:ns (find-ns 'cider.nrepl.middleware.info)}))))
   (is (not (re-find #"/form-init[^/]*$" (:file (info/var-meta (eval '(do (in-ns 'cider.nrepl.middleware.info) (def pok 10)))))))))
 
+(deftest javadoc-info-unit-test
+  (testing "Get an HTTP URL for a sun/oracle Javadoc"
```
bbatsov: You should capitalize names like Sun, Oracle, Clojure...

test/clj/cider/nrepl/middleware/info_test.clj
```diff
@@ -163,6 +176,215 @@
...
+          exp-type java.net.URL]
+      (is (= url-type exp-type))))
+
+  (testing "Get fall through URL type for other Javadocs (external libs?)"
+    (let [reply (info/javadoc-info "http://some/other/url")
+          url (:javadoc reply)]
+      (is (= url "http://some/other/url")))))
+
+;; TODO: These cljx need to be fleshed out. Right now just in to get 100% coverage
+;; TODO: ...or the cljx stuff could be removed since it's deprecated now?
```
bbatsov: It's deprecated, but some people still use it, as they want to support older Clojure releases (older than 1.7).

test/clj/cider/nrepl/middleware/info_test.clj
```diff
@@ -163,6 +176,215 @@
...
+            individual (:java.lang.Exception candidates)]
+        (is (contains? candidates :java.lang.NoSuchFieldException))
+        (is (contains? candidates :java.lang.Package))
+        (is (contains? candidates :java.lang.LinkageError))
+
+        (is (= (:throws individual) []))
+        (is (= (:member individual) "toString"))
+        (is (= (:modifiers individual) "#{:public}"))))
+
+    (testing "boot stuff"
```
bbatsov: boot stuff - boot support

test/clj/cider/nrepl/middleware/info_test.clj
```diff
@@ -163,6 +176,215 @@
...
+      (is (= (:eldoc response) [["try" "expr*" "catch-clause*" "finally-clause?"]])))
+
+    (let [response (session/message {:op "eldoc" :symbol "." :ns "user"})]
+      (is (= (:status response) #{"done"})))
+
+    ;;TODO: Since this is an eldoc java op, shouldn't the args should be :class and :member?
+    (let [response (session/message {:op "eldoc" :symbol "toString" :ns (ns-name *ns*)})]
+      (is (set/subset? #{[] ["this"] ["int"] ["this" "char"]} (set (:eldoc response)))))))
+
+(deftest misses
```
bbatsov: Maybe `missing-info` is a better name for this.

bbatsov:
I've got one concern with using library functions/methods in the tests - their metadata might change between Clojure/Java releases, which would cause the tests to fail. Not a major issue, but something worth keeping in mind.

test/clj/cider/nrepl/middleware/info_test.clj
sanjayl: The test doesn't seem to do what the description says. It looks the test tries to resolve a function that doesn't exist in the current namespace -- I can't see anywhere where it tries to resolve the protocol T. (I'm still trying to learn about Protocols/Records/Types, so I wasn't sure if I'm missing something here)

test/clj/cider/nrepl/middleware/info_test.clj
sanjayl: Done

test/clj/cider/nrepl/middleware/info_test.clj
sanjayl: Oh ok, makes sense. I was thinking that since the current release of Cider requires 1.7+, that it might be safe to tear out. Will remove this TODO.

test/clj/cider/nrepl/middleware/info_test.clj
sanjayl: Done

test/clj/cider/nrepl/middleware/info_test.clj
sanjayl: Good point. Made the change and pulled out a seperate `missing-eldoc` group of tests.

sanjayl:
> I've got one concern with using library functions/methods in the tests - their metadata might change between Clojure/Java releases, which would cause the tests to fail. Not a major issue, but something worth keeping in mind.

Hmm, interesting point. I tried to throw some real world (ie, external) data into the integration tests since the unit tests already tested against internal cider-nrepl functions. I tried to use Clojure functions that have been stable for a couple releases, but you're right that they could always change from underneath us.

Perhaps we could keep the integration tests the way they are for now and if the tests become a moving target, I'd be happy to redo them against internal test functions. The info middleware is kinda ripe for refactoring, so I just wanted to have a really robust set of tests that pedantically verify some data to help with that process.

bbatsov:
Sounds like a plan! :-) Thanks!

sanjayl:
@bbatsov I didn't realize how crazily broken these tests are when run against different JVMs! I'll be redoing the tests over the next couple days

Malabarba:
@sanjayl Could you find some time to fix  the integration test? I've commented it out on master, but it could have been pointing to a genuine bug. 
Here's the failure report:

```
FAIL in (integration-tests) (info_test.clj:311)
eldoc op
expected: (set/subset? #{[] ["int"] ["this"] ["this" "char"]} (set (:eldoc response)))
  actual: (not (set/subset? #{[] ["int"] ["this"] ["this" "char"]} #{["this" "c"] ["this" "u" "sb" "radix" "digits"] ["b"] ["s"] ["i"] [] ["d"] ["f"] ["this" "radix"] ["this"] ["i" "radix"]}))
```

Malabarba:
Never mind. I just noticed you have a PR open for that.

sanjayl:
Yeah, I think it was due to differences between the Oracle and openjdk
libraries (test was written against openjdk originally). The PR branch is
written against classes in our code base, which should obviate this error

PR #311, opened 2016-03-06
Debugger: make skip-breaks depend on current code

grammati:
Previously, the var `*skip-breaks*` was storing a mode (:deeper,
:before, etc.) and a sequence of coordinates. However, it did not store
any information indicating what code the coordinates referred to.

This would lead to strange behavior when an :out or :here operation
would skip over code that called another instrumented function. The
debugger would treat the coordinates in *skip-breaks* as if they
referred to the code of the called function, and would seem (to the
user) to skip randomly over some or all of the function.

The fix:

1) Store *skip-breaks* as a map containing :mode, :coor, and :code,
where :code is the string of code that the :out or :here operation was
performed on.

2) Add logic to `skip-breaks?` to only skip if the to-be-skipped code
and the currently-being-debugged code are the same.

### DIFF
--- a/src/cider/nrepl/middleware/debug.clj
+++ b/src/cider/nrepl/middleware/debug.clj
@@ -57,10 +57,23 @@
 ;;;; ## Internal breakpoint logic
 ;;; Variables and functions used for navigating between breakpoints.
 (def ^:dynamic *skip-breaks*
-  "Boolean or vector to determine whether to skip a breakpoint.
+  "Map used to determine whether to skip a breakpoint.
   Don't set or examine this directly, it is bound in the session
   binding map, use `skip-breaks!` and `skip-breaks?` instead.
   Its value is discarded at the end each eval session."
+  nil)
+
+(def ^:dynamic *extras*
+  "Bound by `with-debug-bindings` to a modified version of the original `*msg*`."
+  {})
+
+(def ^:dynamic *locals*
+  "Bound by `with-debug-bindings` to a map from local symbols to values."
+  {})
+
+(def ^:dynamic *pprint-fn*
+  "Bound by `with-debug-bindings` to the pretty-printing function determined by
+  the `wrap-pprint-fn` middleware."
   nil)
 
 (defn coord<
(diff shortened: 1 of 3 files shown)


Malabarba:
Thanks @grammati. This is something I've ran into a couple of times too.

The only thing that makes me unsure about this fix is the fact that assumes that each `defn` was instrumented in a separate `*msg*`.
Right now, this is always true, but that's actually something I consider a bug in the current implementation. (Currently, debugger just fails to instrument multiple functions in the same message, such as with `load-buffer` or `eval-region`).

That said, unless I get off my ass and fix the implementation itself, I can't really think of a better way to fix the `:here/:out` bug than the one you propose.

src/cider/nrepl/middleware/debug.clj
```diff
@@ -102,21 +116,30 @@
...
     true))
 
 (defn skip-breaks!
   "Set the value of *skip-breaks* for the top-level breakpoint.
-  If bool-or-vec is a vector, mode should be :deeper or :before (see
-  `skip-breaks?`). Otherwise mode should be :all or false."
-  [mode & [bool-or-vec]]
-  (reset! *skip-breaks* (when mode
-                          (cons mode bool-or-vec))))
+  Additionl arguments depend on mode, and should be:
```
Malabarba: additionAl

src/cider/nrepl/middleware/debug.clj
```diff
@@ -435,14 +445,13 @@
       (try
         (read-string code)
         (catch Exception e)))
-    (binding [*skip-breaks* (atom nil)]
```
Malabarba: I agree with this change, but PRs are easier to read if unrelated changes are left in separate commits. ;-)

src/cider/nrepl/middleware/debug.clj
grammati: Thanks, clearly I wasn't using flyspell :)

src/cider/nrepl/middleware/debug.clj
grammati: Fair point. I guess I considered this to be somewhat related, in that I made sure to carefully examine all uses of `*skip-breaks*`, and saw that this one was a leftover. Happy to move it to a separate PR if you prefer.

bbatsov: Separate commit in the same PR would be fine.

grammati: OK, I put this part in a separate commit, rebased, and pushed.

Malabarba:
Thanks again. I'll merge this as soon as I get a chance to read it again.

PR #312, opened 2016-03-06
Force step out

grammati:
NOTE: Depends on https://github.com/clojure-emacs/cider-nrepl/pull/311

Add optional force parameter to :out and :here ops.

When an `:out` or `:here` operation jumps past calls to other
instrumented functions, the debugger will normally stop in those
functions.

Passing a truthy `force` parameter to one of these ops will cause the
debugger to skip all breaks outside of the currently debugged form,
effectively forcing it to immediately step to the location specified in the 
`:out` or `:here` operation.

### DIFF
--- a/src/cider/nrepl/middleware/debug.clj
+++ b/src/cider/nrepl/middleware/debug.clj
@@ -98,30 +98,38 @@
 
 (defn skip-breaks?
   "True if the breakpoint at coordinates should be skipped.
-  If the mode of `*skip-breaks*` is :all, return true.
 
-  Otherwise, the mode should be :deeper, :before, or :trace.
-  If :deeper, return true if the given coordinates are deeper than the
-  coordinates stored in `*skip-breaks*`.
-  If :before, return true if they represent a place before the coordinates
-  in `*skip-breaks*`.
-  If :trace, return false."
+  The `*skip-breaks*` map stores a `mode`, `coordinates`, the `code` that it
+  applies to, and a `force?` flag. Behaviour depends on the `mode`:
+   - :all - return true, skipping all breaks
+   - :trace - return false, skip nothing
+   - :deeper - return true if the given coordinates are deeper than the
+               coordinates stored in `*skip-breaks*`, in the same code
+   - :before - return true if the given coordinates represent a place before
+               the coordinates in `*skip-breaks*`, in the same code
+
+  For :deeper and :before, if we are not in the same code (i.e. we have stepped
+  into another instrumented function), then return the value of `force?`."
   [coordinates]
   (if (seq coordinates)
-    (when-let [{mode :mode skip-coords :coor code :code} @*skip-breaks*]
-      (case mode
-        ;; From :continue, skip everything.
-        :all    true
-        ;; From :trace, never skip.
-        :trace  false
-        ;; From :out, skip some breaks.
-        :deeper (let [parent (take (count skip-coords) coordinates)]
-                  (and (= code (:code *extras*))
-                       (seq= skip-coords parent)
-                       (> (count coordinates) (count parent))))
-        ;; From :here, skip some breaks.
-        :before (and (= code (:code *extras*))
-                     (coord< coordinates skip-coords))))
+    (when-let [{mode :mode skip-coords :coor
+                code-to-skip :code force? :force?} @*skip-breaks*]
+      (let [same-defn? (identical? code-to-skip (:code *extras*))]
+        (case mode
+          ;; From :continue, skip everything.
+          :all    true
+          ;; From :trace, never skip.
+          :trace  false
+          ;; From :out, skip some breaks.
+          :deeper (if same-defn?
+                    (let [parent (take (count skip-coords) coordinates)]
+                      (and (seq= skip-coords parent)
+                           (> (count coordinates) (count parent))))
+                    force?)
+          ;; From :here, skip some breaks.
+          :before (if same-defn?
+                    (coord< coordinates skip-coords)
+                    force?))))
     ;; We don't breakpoint top-level sexps, because their return value
     ;; is already displayed anyway.
     true))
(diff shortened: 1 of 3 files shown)


grammati:
To clarify my intent here: when I step "out" or "here", sometimes I want the debugger to stop in another function that I have instrumented, and sometimes I just want to go immediately to the place I specified.

This PR is provide an option to do either. I was thinking that the "forced" (non-breaking) versions of `out` and `here` could be bound to capitalized variants in cider: `O` and `H`. Thoughts?

BTW, I realize that this would increase the number of debugger commands to the point where they might not fit nicely in the overlay anymore, but perhaps it could be changed to show the 5 or 6 most common commands along with a `?` to pop up the full listing somehow. (I would also find `f` for "forward", as in edebug useful, as well as maybe `r` for "return", to step out of the top-level defn).

Malabarba:
> This PR is provide an option to do either. I was thinking that the "forced" (non-breaking) versions of out and here could be bound to capitalized variants in cider: O and H. Thoughts?

Yes, this sounds good. 

> BTW, I realize that this would increase the number of debugger commands to the point where they might not fit nicely in the overlay anymore, but perhaps it could be changed to show the 5 or 6 most common commands along with a ? to pop up the full listing somehow.

Agreed. 👍

bbatsov:
> This PR is provide an option to do either. I was thinking that the "forced" (non-breaking) versions of out and here could be bound to capitalized variants in cider: O and H. Thoughts?

I'm ok with this, although using other lowercase letters might be better in terms of consistency. Over time I've grown to dislike uppercase keyboard shortcuts.



> BTW, I realize that this would increase the number of debugger commands to the point where they might not fit nicely in the overlay anymore, but perhaps it could be changed to show the 5 or 6 most common commands along with a ? to pop up the full listing somehow. (I would also find f for "forward", as in edebug useful, as well as maybe r for "return", to step out of the top-level defn).

Or we can have a 2 line overlay. But yeah - showing just the most common commands is totally fine.

Malabarba:
By the way, I haven't commented anything because I haven't read the code yet (while we deal with the other PR). :-)

bbatsov:
What's our status here?

Malabarba:
Looks like two of the commits on this PR are already in master. @grammati could you rebase on master?

grammati:
OK, it's rebased.

src/cider/nrepl/middleware/debug.clj
```diff
@@ -98,30 +98,39 @@
...
-        :deeper (let [parent (take (count skip-coords) coordinates)]
-                  (and (= code (:code *extras*))
-                       (seq= skip-coords parent)
-                       (> (count coordinates) (count parent))))
-        ;; From :here, skip some breaks.
-        :before (and (= code (:code *extras*))
-                     (coord< coordinates skip-coords))))
+    (when-let [{mode :mode skip-coords :coor
+                code-to-skip :code force? :force?} @*skip-breaks*]
+      (let [same-defn? (= code-to-skip (:code *extras*))]
```
Malabarba: I think `identical?` here would be faster and would be slightly more correct (it would cover the case where there are two functions with the same code content instrumented in different namespaces (not that that would eve happen)).

src/cider/nrepl/middleware/debug.clj
```diff
@@ -98,30 +98,39 @@
...
+   - :all - return true, skipping all breaks
+   - :trace - return false, skip nothing
+   - :deeper - return true if the given coordinates are deeper than the
+               coordinates stored in `*skip-breaks*`, in the same code
+   - :before - return true if the given coordinates represent a place before
+               the coordinates in `*skip-breaks*`, in the same code
+
+  For :deeper and :before, if we are not in the same code (i.e. we have stepped
+  into another instrumented function), then return the value of `force?`.
+  "
```
Malabarba: Pull this onto the line above.

src/cider/nrepl/middleware/debug.clj
```diff
@@ -98,30 +98,39 @@
...
+    (when-let [{mode :mode skip-coords :coor
+                code-to-skip :code force? :force?} @*skip-breaks*]
+      (let [same-defn? (= code-to-skip (:code *extras*))]
+        (case mode
+          ;; From :continue, skip everything.
+          :all    true
+          ;; From :trace, never skip.
+          :trace  false
+          ;; From :out, skip some breaks.
+          :deeper (or (and force? (not same-defn?))
```
Malabarba: I don't think this does what the docstring claims. Shouldn't it be something like `(if (not same-defn?) force? (let ...))`?

test/clj/cider/nrepl/middleware/debug_integration_test.clj
```diff
@@ -94,11 +94,13 @@
      (nrepl-send {:op :debug-input :input ~(str op) :key (current-key)})))
 
 (def-debug-op :next)
-(def-debug-op :out)
```
Malabarba: Why is this no longer necessary?

src/cider/nrepl/middleware/debug.clj
grammati: Done, I rearranged the logic as you suggest - it looks much clearer now, thanks.

test/clj/cider/nrepl/middleware/debug_integration_test.clj
grammati: It's defined just below, as a defmethod, because it takes an optional `force` parameter now.

grammati:
I pushed another commit addressing your comments, @Malabarba, thanks for the review.

Malabarba:
Merged. Thanks again @grammati 
I've implemented the elisp side for this in [this commit](https://github.com/clojure-emacs/cider/commit/e670a7e94460c872068a50b2e784ad32bc2d2e87). Let me know if it works.

And feel free to open a PR for a help key. :-)

PR #314, opened 2016-03-07
[c.n.m.u.inspect] Treat infinite sequences as lists

alexander-yakushev:
This should hopefully fix clojure-emacs/cider#1602.

An infinite syntax is not a `list?`. Neither are Cons cells, and some other things. `sequential?` predicate covers all of them.

### DIFF
--- a/src/cider/nrepl/middleware/util/inspect.clj
+++ b/src/cider/nrepl/middleware/util/inspect.clj
@@ -93,8 +93,9 @@
     (map? value)                                   :map-long
     (and (vector? value) (< (count value) 5))      :vector
     (vector? value)                                :vector-long
-    (and (list? value) (< (count value) 5))        :list
-    (list? value)                                  :list-long
+    (and (seq? value) (not (counted? value)))      :lazy-seq
+    (and (seq? value) (< (count value) 5))         :list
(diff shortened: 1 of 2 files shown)


bbatsov:
The change looks good. We should just add some regression unit test using an infinite sequence.

src/cider/nrepl/middleware/util/inspect.clj
```diff
@@ -93,8 +93,8 @@
     (map? value)                                   :map-long
     (and (vector? value) (< (count value) 5))      :vector
     (vector? value)                                :vector-long
-    (and (list? value) (< (count value) 5))        :list
-    (list? value)                                  :list-long
+    (and (sequential? value) (< (count value) 5))  :list
```
bbatsov: Isn't `seq?` what we need here?

alexander-yakushev: It should be `seq?` indeed. Foiled again!

alexander-yakushev:
@bbatsov RE: predicates. Just written that down http://bytopia.org/2016/03/08/what-is-a-list/

bbatsov:
Nice! :+1:

PR #315, opened 2016-03-09
Fix Javadoc URLs for Java 1.8

sanjayl:

- [x] All tests are passing

Thanks!

An issue where the URLs for Javadoc in Java 1.8 has a different format
than prior Java version, namely, parenthesis surrounding argument lists
have been changed to dashes. Added tests as well.

### DIFF
--- a/src/cider/nrepl/middleware/util/java.clj
+++ b/src/cider/nrepl/middleware/util/java.clj
@@ -75,7 +75,10 @@
   ([class member argtypes]
    (str (javadoc-url class) "#" member
         (when argtypes
-          (str "(" (str/join ",%20" argtypes) ")")))))
+          (if (< (Integer/parseInt util/java-api-version) 8)
+            ;; Pre JDK 1.8 vs Post JDK 1.8 Javadoc URL style
+            (str "(" (str/join ",%20" argtypes) ")")
+            (str "-" (str/join "-" (map #(str/replace % #"\[\]" ":A") argtypes)) "-"))))))
 
 ;;; ## Class Metadata Assembly
 ;; We construct metadata at the class level, first using `reflect-info` to
--- a/test/clj/cider/nrepl/middleware/info_test.clj
+++ b/test/clj/cider/nrepl/middleware/info_test.clj
@@ -16,6 +16,21 @@
   [x]
   (:resource (info/file-info x)))
 
+(deftest test-javadoc-url
+  (testing "java 1.7"
+    (is (= "http://docs.oracle.com/javase/7/docs/api/java/lang/StringBuilder.html#charAt(int)"
+           (with-redefs [util/java-api-version "7"]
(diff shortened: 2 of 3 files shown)


test/clj/cider/nrepl/middleware/info_test.clj
```diff
@@ -16,6 +16,28 @@
   [x]
   (:resource (info/file-info x)))
 
+(deftest test-javadoc-url
+  (testing "java 1.6"
```
bbatsov: We don't support Java 6 anymore (although Clojure itself does), so this test is not necessary. Plus, it's the same behaviour as in 7.

src/cider/nrepl/middleware/util/java.clj
```diff
@@ -73,9 +73,14 @@
        (str/replace "$" ".")
        (str ".html")))
   ([class member argtypes]
-   (str (javadoc-url class) "#" member
-        (when argtypes
-          (str "(" (str/join ",%20" argtypes) ")")))))
+   (let [java-version (read-string util/java-api-version)
```
Malabarba: I'd rather use `parse-int` or `Integer.`.

sanjayl:
Made the changes rec'ed above. Added a few more changes to get as close as possible to the new URL spec (although from what I've surmised, we'll never be 100% since the URLs are created via parsing the source, not via reflection). Changed the unit tests a bit.

Apologizing in advance at the ugliness of this code, I wasn't sure how to go about fixing it since the necessary changes were scattered across the code base.

src/cider/nrepl/middleware/util/java.clj
```diff
@@ -73,9 +73,19 @@
        (str/replace "$" ".")
        (str ".html")))
   ([class member argtypes]
-   (str (javadoc-url class) "#" member
-        (when argtypes
-          (str "(" (str/join ",%20" argtypes) ")")))))
+   (let [java-version (Integer/parseInt util/java-api-version)
+         old-format   (< java-version 8)
+         open-paren   (if old-format "(" "-")
```
bbatsov: Let's just have two string templates - one for old and one for the new format and insert into them whatever's needed. Or just build the new old or new string without actually introducing those locals. They serve no purpose and their names are confusing for the new format (e.g. close-paren is a "-"). I'd rather have two functions that build the old/new format or something like this.

Malabarba: Yes, I think it's better to just have one `if` where each branch is the entire logic for that version.

bbatsov: Or one `if` branching out to two helper functions. :-) Anyways, we're on the same page here.

sanjayl:
Yeah, that was pretty dumb of me. Here's the redo

bbatsov:
Perfect! :+1:

PR #316, opened 2016-03-10
Include cider-nrepl anyway when clj is excluded

arrdem:
This patch allows the cider-nrepl middleware to be injected into a project which does not depend on `org.clojure/clojure`, if and only if the user has explicitly provided `:exclusions [... org.clojure/clojure]`
in the `project.clj` file. This allows users who know what they're doing to take on the responsibility of providing an appropriate build of Clojure and thus use CIDER with forks albeit with a warning.

Replaces https://github.com/clojure-emacs/cider/pull/1603

### DIFF
--- a/src/cider_nrepl/plugin.clj
+++ b/src/cider_nrepl/plugin.clj
@@ -15,21 +15,34 @@
     v))
 
 (defn middleware
-  [{:keys [dependencies] :as project}]
-  (let [lein-version-ok? (lein/version-satisfies? (lein/leiningen-version) "2.5.2")
-        clojure-version (->> dependencies
-                             (some (fn [[id version & _]]
-                                     (when (= id 'org.clojure/clojure)
-                                       version))))
-        clojure-version-ok? (if (nil? clojure-version)
-                              ;; Lein 2.5.2+ uses Clojure 1.7 by default
-                              lein-version-ok?
-                              (lein/version-satisfies? clojure-version "1.7.0"))]
+  [{:keys [dependencies exclusions] :as project}]
+  (let [lein-version-ok?    (lein/version-satisfies? (lein/leiningen-version) "2.5.2")
+        clojure-excluded?   (some #(= % 'org.clojure/clojure) exclusions)
+        clojure-version     (when-not clojure-excluded?
+                              (->> dependencies
+                                   (some (fn [[id version & _]]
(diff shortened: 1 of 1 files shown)


src/cider_nrepl/plugin.clj
```diff
@@ -28,8 +31,10 @@
 
     (when-not lein-version-ok?
       (lein/warn "Warning: cider-nrepl requires Leiningen 2.5.2 or greater."))
-    (when-not clojure-version-ok?
+    (when (or (not clojure-version-ok?) clojure-excluded?)
       (lein/warn "Warning: cider-nrepl requires Clojure 1.7 or greater."))
+    (when-not clojure-excluded?
```
Malabarba: Shouldn't this be a when?

src/cider_nrepl/plugin.clj
```diff
@@ -28,8 +31,10 @@
 
     (when-not lein-version-ok?
       (lein/warn "Warning: cider-nrepl requires Leiningen 2.5.2 or greater."))
-    (when-not clojure-version-ok?
+    (when (or (not clojure-version-ok?) clojure-excluded?)
```
Malabarba: Shouldn't this be a `(when-not (or clojure-version-ok? clojure-excluded?)`?

src/cider_nrepl/plugin.clj
arrdem: Yes. Fixing.

src/cider_nrepl/plugin.clj
arrdem: No. This was deliberate. If Clojure has been excluded, then `clojure-version-ok?` will be computed using the default based on Leiningen's version, which should be OK, and `clojure-version-ok?` will presumably be `true`. If Clojure is not in the `:dependencies`, we do want to emit a warning here, so `(or (not true) true) -> (or false true) -> true` is appropriate.

bbatsov:
Sprinkle some comments throughout the changes to explain the need for them. In the absence of tests non-obvious things should be clarified via comments.

arrdem:
Updated with comments, removed `set` call which was never used again, replaced with a `(some = ...)` which is boolean equivalent. Warning logic simplified so as to be more obvious.

bbatsov:
:+1:

PR #317, opened 2016-03-11
Expand Info middleware tests

### LINKED ISSUE #308: cider.nrepl.middleware.info/javadoc-info should take *remote-javadocs* into account
Currently we use a manual regexp method, but I think that's mostly the same as taking `clojure.java.javadoc/*remote-javadocs*` into account, and doing that might be more extensible by 3rd party tools.
Not sure though. Does anyone actually use that variable?

sanjayl:

- [x] All tests are passing
- [n/a] You've updated the readme (if adding/changing middleware)

Thanks!

Added tests to exercise the Javadoc URL issue [Fix #308].

Added test to see if the *remote-javadocs* functionality works (same fix as
above).

Changed some of the testing targets to internal test classes/methods
instead of testing against code outside of our control (when possible). As
a result, added a java source tree under the test directory and added
this to the test-clj profile.

Brings coverage back up to 100%.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -55,7 +55,8 @@
              :master {:repositories [["snapshots" "https://oss.sonatype.org/content/repositories/snapshots"]]
                       :dependencies [[org.clojure/clojure "1.8.0-master-SNAPSHOT"]]}
 
-             :test-clj {:test-paths ["test/clj"]}
+             :test-clj {:test-paths ["test/clj"]
+                        :java-source-paths ["test/java"]}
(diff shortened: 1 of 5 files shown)


Malabarba:
This is 100% tests, and it looks like they're passing on Travis (the two failures are unrelated), and I don't see any crimes against humanity in there, so I've taken the liberty to merge.

Thanks :+1:

sanjayl:
Hahaha, well let me know if any of my PRs start water boarding the code
base so I can get them turned around

PR #321, opened 2016-03-14
Improve pprint middleware tests & simplify code

sanjayl:

- [x] All tests are passing

Thanks!

The prior code had a difficult to test code block in `resolve-pprint-fn`
as well as a small piece of round about code in `wrap-pprint` , so these
functions were streamlined to help with testing/understanding. Behavior
is slightly changed since the resolve-pprint-fn will now throw different
exceptions based on the root cause of the exception (no such namespace
in classpath, no such var in an existing namespace, etc. This required a
slight change in the format-middleware tests). Otherwise, the
functionality of the code should be the same.

Added coverage tests to get to 100%. Error handling is actually done
elsewhere in the middleware stack (within the `eval` op).

### DIFF
--- a/src/cider/nrepl/middleware/pprint.clj
+++ b/src/cider/nrepl/middleware/pprint.clj
@@ -23,12 +23,9 @@
 
 (defn- resolve-pprint-fn
   [sym]
-  (let [var (some-> sym u/as-sym resolve)]
-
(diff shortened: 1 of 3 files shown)


bbatsov:
:+1: The changes look good to me.

PR #323, opened 2016-03-19
Store exceptions so sync ops can get stacktraces

sanjayl:

- [ ] All tests are passing

Thanks!

The u/err-info function will now automatically store exceptions in
simple map based single-use data store. This will allow the stacktrace
middleware to act on Exceptions that were not bound to *e. See bug 1420
in CIDER: https://github.com/clojure-emacs/cider/issues/1420.

See also CIDER PR 1617: https://github.com/clojure-emacs/cider/pull/1617

### DIFF
--- a/src/cider/nrepl/middleware/stacktrace.clj
+++ b/src/cider/nrepl/middleware/stacktrace.clj
@@ -3,6 +3,7 @@
   {:author "Jeff Valk"}
   (:require [cider.nrepl.middleware.pprint :as pprint]
             [cider.nrepl.middleware.util.cljs :as cljs]
+            [cider.nrepl.middleware.util.storage :as c-store]
             [clojure.repl :as repl]
             [clojure.string :as str]
             [clojure.tools.nrepl.middleware :refer [set-descriptor!]]
@@ -156,10 +157,10 @@
 ;;; ## Middleware
 
 (defn wrap-stacktrace-reply
-  [{:keys [session transport pprint-fn] :as msg}]
-  ;; no stacktrace support for cljs currently - they are printed by piggieback anyway
-  (if-let [e (and (not (cljs/grab-cljs-env msg))
-                  (@session #'*e))]
+  [{:keys [session transport pprint-fn storage-key] :as msg}]
+  (if-let [e (cond (cljs/grab-cljs-env msg) nil   ;; no cljs stacktraces (printed by piggieback)
+                   storage-key (c-store/query! storage-key) ;; stored Exceptions for sync-calls
+                   :else (@session #'*e))] ;; bound *e for non-sync "eval" calls
     (doseq [cause (analyze-causes e pprint-fn)]
       (t/send transport (response-for msg cause)))
     (t/send transport (response-for msg :status :no-error)))
@@ -170,8 +171,8 @@
   info for the most recent exception."
   [handler]
   (fn [{:keys [op] :as msg}]
-    (if (= "stacktrace" op)
-      (wrap-stacktrace-reply msg)
+    (case op
+      "stacktrace" (wrap-stacktrace-reply msg)
       (handler msg))))
 
 ;; nREPL middleware descriptor info
@@ -183,5 +184,6 @@
    :handles {"stacktrace"
              {:doc (str "Return messages describing each cause and stack frame "
                         "of the most recent exception.")
-              :optional pprint/wrap-pprint-fn-optional-arguments
-              :returns {"status" "\"done\", or \"no-error\" if `*e` is nil"}}}}))
+              :optional (merge pprint/wrap-pprint-fn-optional-arguments
+                               {"storage-key" "The hash key for the Exception you're looking for if not looking for `*e`"})
+              :returns {"status" "\"done\", or \"no-error\" if no Exception in `*e` or the storage map."}}}}))
--- a/src/cider/nrepl/middleware/util/storage.clj
+++ b/src/cider/nrepl/middleware/util/storage.clj
@@ -1,0 +1,41 @@
+(ns cider.nrepl.middleware.util.storage)
+
+(def ^:private store (ref {}))
+(def ^:private exp-index (ref '()))
+(def ^:private capacity 10)
+
+(defn add!
+  "Adds the provided kv pair to a single-retrieval store with
+  queue-capacity expiring behavior; i.e. not only is kv removed from
+  the store when sucessfully queried, but if addition of the supplied
+  item causes store size to increase over the default capacity of 10
+  items, then the oldest kv pair in the store is also
+  expired (removed). Returns the value added to the store."
+  [k v]
+  (dosync
+   (let [untrimmed-exp           (conj @exp-index k)
+         untrimmed-store         (assoc @store k v)
+         [trimmed-exp trimmable] (split-at capacity untrimmed-exp)
+         trimmed-store           (apply dissoc untrimmed-store trimmable)]
+     (ref-set store trimmed-store)
+     (ref-set exp-index trimmed-exp)
+     v)))
+
+(defn query!
+  "Queries the single-retrieval store. Finds value associated with the
+  supplied key `k` and then removes the kv pair from storage. If `k`
+  not present, returns nil and no changes are made."
+  [k]
+  (dosync
+   (let [reply (get @store k)]
+     (ref-set store (dissoc @store k))
+     (ref-set exp-index (remove #(= k %) @exp-index))
+     reply)))
+
+(defn refresh!
+  "Refreshes the single-retrieval, queue-capacity store. Removes all kv
+  pairs from the store."
+  []
+  (dosync
+   (ref-set store {})
+   (ref-set exp-index '())))
--- a/src/cider/nrepl/middleware/util/misc.clj
+++ b/src/cider/nrepl/middleware/util/misc.clj
@@ -1,6 +1,7 @@
 (ns cider.nrepl.middleware.util.misc
   (:require [clojure.string :as str]
-            [clojure.stacktrace :as stacktrace]))
+            [clojure.stacktrace :as stacktrace]
+            [cider.nrepl.middleware.util.storage :as c-store]))
 
 (def java-api-version
   (try (-> (System/getProperty "java.version") (str/split #"\.") second)
@@ -60,7 +61,10 @@
 (prefer-method transform-value clojure.lang.Sequential clojure.lang.Associative)
 
 (defn err-info
-  [ex status]
-  {:ex (str (class ex))
-   :err (with-out-str (stacktrace/print-cause-trace ex))
-   :status #{status :done}})
+  [ex & statuses]
+  (let [ex-key (.hashCode ex)]
+    (c-store/add! ex-key ex)
+    {:ex (str (class ex))
+     :err (with-out-str (stacktrace/print-cause-trace ex))
+     :status (set (conj statuses :done))
+     :storage-key ex-key}))
--- a/test/clj/cider/nrepl/middleware/format_test.clj
+++ b/test/clj/cider/nrepl/middleware/format_test.clj
@@ -80,10 +80,11 @@
       (is (= #{"done"} status))))
 
   (testing "format-edn returns an error if the given EDN is malformed"
-    (let [{:keys [err status]} (session/message {:op "format-edn"
-                                                 :edn unmatched-delimiter-edn-sample})]
+    (let [{:keys [err status] :as response} (session/message {:op "format-edn"
+                                                              :edn unmatched-delimiter-edn-sample})]
       (is (= #{"edn-read-error" "done"} status))
-      (is (.startsWith err "clojure.lang.ExceptionInfo: Unmatched delimiter"))))
+      (is (.startsWith err "clojure.lang.ExceptionInfo: Unmatched delimiter"))
+      (is (:storage-key response))))
 
   (testing "format-edn respects the :print-right-margin slot"
     (let [wide-edn-sample     "[1 2 3 4 5 6       7 8     9    0]"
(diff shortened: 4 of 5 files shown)


src/cider/nrepl/middleware/util/storage.clj
```diff
@@ -0,0 +1,16 @@
+(ns cider.nrepl.middleware.util.storage)
+
+;; TODO: How should we expire the cache?
+;; Currently effectively single use storage for use with non-sync exceptions,
+;; but since this can likely be used for any type of data,
+;; perhaps we should have some formal autopurge functionality so that this
+;; doesn't grow unbounded?
+(def ^:dynamic *storage* (atom {}))
```
bbatsov: Does this need to be dynamic? Most of the comment should actually the var's docstring IMO.

As for expiring the cache the last 10 exceptions or something like this and simply drop the oldest when we ran out of space.

src/cider/nrepl/middleware/util/storage.clj
```diff
@@ -0,0 +1,16 @@
...
+
+(defn query [k]
```
bbatsov: This should have a docstring.

src/cider/nrepl/middleware/util/storage.clj
```diff
@@ -0,0 +1,16 @@
...
+  (let [reply (get @*storage* k)]
+    (swap! *storage* dissoc k)
+    reply))
+
+(defn add [k v]
```
bbatsov: Probably this should have one as well.

src/cider/nrepl/middleware/util/storage.clj
bbatsov: I'm also not sure about the structure of the cache - if this is going to be used for other things it'd be hard to keep track of what is what with the current implementation.

sanjayl:
Here's my second cut at the problem. I haven't coded Clojure state before, so it's probably not very idiomatic. For example, I was considering using `alter` instead of `set-ref`, but the functions I am using to mutate the collections have arg lists that are not congruent with `alter's` expectations (except the `dissoc` call). I was considering adding inline anon functions to reorder the arguments to the alter call, but thought it would muddy up the intent?

> I'm also not sure about the structure of the cache - if this is going to be used for other things it'd be hard to keep track of what is what with the current implementation.

Could you give me some more information/direction on this aspect of the problem?

Thanks for your help!

src/cider/nrepl/middleware/stacktrace.clj
```diff
@@ -156,13 +157,16 @@
...
 (defn wrap-stacktrace-reply
-  [{:keys [session transport pprint-fn] :as msg}]
+  [{:keys [session transport pprint-fn storage-key] :as msg}]
   ;; no stacktrace support for cljs currently - they are printed by piggieback anyway
-  (if-let [e (and (not (cljs/grab-cljs-env msg))
-                  (@session #'*e))]
-    (doseq [cause (analyze-causes e pprint-fn)]
-      (t/send transport (response-for msg cause)))
-    (t/send transport (response-for msg :status :no-error)))
+  (letfn [(get-ex [] (if storage-key
```
Malabarba: There's no need to factor this out as function if it's only used once. I would inline this, and replace the whole `and+not+if` construction below with a `cond`.

src/cider/nrepl/middleware/util/storage.clj
```diff
@@ -0,0 +1,42 @@
...
+  queue-capacity expiring behavior; i.e. not only is kv removed from
+  the store when sucessfully queried, bual also if addition of the
+  supplied item causes store size to increase over the default
+  capacity of 10 items, then the oldest kv pair in the store is
+  expired (removed). Returns the value added to the store."
+  [k v]
+  (dosync
+   (let [untrimmed-exp   (conj @exp-index k)
+         untrimmed-store (assoc @store k v)
+         trimmable       (drop capacity untrimmed-exp)
```
Malabarba: IIUC, you can replace this line and the one below with:

```clj
[trimmed trimmable] (split-at capacity untrimmed-exp)
```

src/cider/nrepl/middleware/stacktrace.clj
sanjayl: Oh wow, that's so much clearer now thanks!

src/cider/nrepl/middleware/util/storage.clj
sanjayl: Done!

bbatsov:
> Could you give me some more information/direction on this aspect of the problem?

I meant that now you can have just 10 items globally in the cache. If we wanted to store different type of data we might have wanted to have e.g. 10 of every type. Not a big concern right now when it's used just for exceptions, but it's something I wanted to note anyways. Maybe it'd be simpler to have different hashes for different caches anyways.

src/cider/nrepl/middleware/util/storage.clj
```diff
@@ -0,0 +1,41 @@
+(ns cider.nrepl.middleware.util.storage)
+
+(def ^:private store (ref {}))
```
bbatsov: You might want to add docstrings to those vars.

src/cider/nrepl/middleware/util/storage.clj
```diff
@@ -0,0 +1,41 @@
...
+
+(defn add!
+  "Adds the provided kv pair to a single-retrieval store with
+  queue-capacity expiring behavior; i.e. not only is kv removed from
+  the store when sucessfully queried, but if addition of the supplied
+  item causes store size to increase over the default capacity of 10
+  items, then the oldest kv pair in the store is also
+  expired (removed). Returns the value added to the store."
+  [k v]
+  (dosync
```
bbatsov: Is it just me, or is this caching code more complex than it needs to be. Btw, I guess you're aware of https://github.com/clojure/core.cache

bbatsov: Alternatively we can store just the last sync exception. I've been thinking now that's not really clear why we'd need older exceptions.

Malabarba: > Is it just me, or is this caching code more complex than it needs to be.

Kind of. It's also simpler than it appears to be at first.

For the intended implementation (a limited-size cache accessible by keys), I think the complexity is acceptable. But I have no idea whether the intended implementation is actually the one we need. If we just want most recent exception then this could be made much simpler of course.

sanjayl:
> Alternatively we can store just the last sync exception. I've been thinking now that's not really clear why we'd need older exceptions. 

Yeah, I agree that it's overkill for 99.99% of situations; it might have utility if one nREPL instance has multiple sessions running concurrently (I'm not even sure if this is possible, but I'm assuming it is) with an error being thrown in each session. In that case, you'd need to have more storage capacity than just the most recent exception. Seems very unlikely and I can switch it back to a simple map if that's the case.

> Is it just me, or is this caching code more complex than it needs to be. Btw, I guess you're aware of https://github.com/clojure/core.cache

I initially considered the `core.cache` library but the semantics don't fit with our needs since it's a memoization cache backed by a function rather than a storage cache backed by a collection. As such, it doesn't have a good way to remove items to my knowledge.

 I have mixed feelings about the complexity of the code. The code does seem more involved than I was initially picturing, but I couldn't think of a simpler way to layer FIFO ordering/removal on top of a hash-map.

I'm open to trying something else, just let me know and I'll take a stab at it or roll it back to something simpler and temp-var like.

Thanks!

bbatsov:
> Yeah, I agree that it's overkill for 99.99% of situations; it might have utility if one nREPL instance has multiple sessions running concurrently (I'm not even sure if this is possible, but I'm assuming it is) with an error being thrown in each session. In that case, you'd need to have more storage capacity than just the most recent exception. Seems very unlikely and I can switch it back to a simple map if that's the case.

You can just have one cache where the session is the key and save the last exception for the session in questions. I believe this is going to be the optimal solution.

sanjayl:
> You can just have one cache where the session is the key and save the last exception for the session in questions. I believe this is going to be the optimal solution.

OK, sounds good. Just as a warning: this will require changing all the calls to `util.misc/err-info` (currently doesn't have any notion of the session or message). I don't mind doing that if this is the preferred solution, I just didn't want your head to explode with the updated pull request!!

bbatsov:
That's ok. Guess it will motivate us more to abstract away this error handling is some dedicated middleware that wraps around other middleware. :-)

bbatsov:
P.S. Adding one such middleware will probably be easier than adjusting `err-info` (and more beneficial in the long run).

sanjayl:
Hahahaha! Excellent point! OK, I've got the next 2 days off work so I'll see what I can do

sanjayl:
Probably a really stupid question, but is there a reason why we need to do a second roundtrip for the `stacktrace` analysis? Instead of using it as a middleware, could `stacktrace` be used like a util ns and error reply packets can get stuffed with the nice stacktrace data at the time of their creation?

bbatsov:
Not quite sure about this. @jeffvalk authored most of this code, so it's best to ask him about the reasoning behind this decision.

sanjayl:
I pushed my [first draft at abstracting away the error code](https://github.com/sanjayl/cider-nrepl/commit/989b14484f4ed100521511ac2e889bb1a5773935) up to my fork. I started it out as a util namespace [based on the ideas I presented before](https://github.com/clojure-emacs/cider-nrepl/pull/318#issuecomment-196062006). (I know you had a middleware in mind, it was just easier for me to quickly put it together and debug if it was a stand alone utility. I think we could still use this code as an impl to a middleware if you'd like to go down that route.)

Prior to this version, all of the `<op>-reply` functions required knowledge about sessions, transports, exception handling, error-codes, etc. In the majority of cases, all that is now taken away. Now, the only thing an `<op>-reply` function needs to do is create small, focused response map that gets merged into the overall response map.

Take a look at [cider.middleware.util.error-handling/with-safe-transport](https://github.com/sanjayl/cider-nrepl/blob/util-error/src/cider/nrepl/middleware/util/error_handling.clj#L118) to see where the abstractions happen; or take a look at the changes made in one of the middlewares (for example, something small like the [resource middleware](https://github.com/sanjayl/cider-nrepl/blob/util-error/src/cider/nrepl/middleware/resource.clj#L20)) to get a gist of it. Despite the `resource middleware` being extremely simple, it has transport, exception handling, stacktrace storage, etc all out of the box.

I changed over most of the middleware and added tests. There were a few middlewares where they were doing something very custom with transports, state, etc where I didn't think this new wrapper would work (debug, refresh, stacktrace, test, and track-state).

I wouldn't be apposed to scrapping this completely and taking a different tact, just let me know what you have in mind.

Here's some pseudocode on how to use it:
```clojure
(ns cider.nrepl.middleware.foo

(defn foo-reply [msg]
  {:foo :answer})

;; other impl fns

(wrap-foo
 [handler]
 (with-safe-transport handler
   "foo-op" foo-reply ;; uses the default error handler
   "invariant-query" {:answer 3.14} ;; default error handler
   "ping" :pong ;; puts :pong in the status set. default error handler
   "broken-op" [broken-reply :service-down] ;; custom error handler, adds :service down to status set
   "retired-op" [broken-reply [:service-down :forever]] ;; custom error handler, adds :service-down and :forever to status set
   "very-broken-op" [fn-that-fails-big custom-clean-up-fn]))

```

bbatsov:
I'll have to go over the code in greater detail, but it looks like a reasonable approach. I favoured the idea of using a middleware mostly to avoid adding any error handling code to middlewares directly for the basic error handling case, but your idea offers certain benefits as well. 

One thing that immediately stands out to me as that `with-safe-transport` is a function and typically things named with `with-something` are macros.

sanjayl:
> One thing that immediately stands out to me as that with-safe-transport is a function and typically things named with with-something are macros.

I'm hoping to eventually change `with-safe-transport` from a function to a macro, which is why I started with that name. I wanted to get the behavior sketched out first before working on the macro to make debugging more straightforward.

The existing approach seems to work pretty well, but I think the inline (hash-map) op/err actions would be safer with a macro-based approach since the (potentially unsafe) evaluation can be delayed until it's wrapped in a try/catch by the macro. E.g:

```clojure
(defn wrap-foo-reply [handler]
  (with-safe-transport handler
    "broken-inline-reply-op" {:example (+ 1 nil)}))
```

jeffvalk:
> Probably a really stupid question, but is there a reason why we need to do a second roundtrip for the stacktrace analysis? Instead of using it as a middleware, could stacktrace be used like a util ns and error reply packets can get stuffed with the nice stacktrace data at the time of their creation?

@sanjayl The separate round-trip is for use flexibility: you can analyze the exception now, at some future point, or not at all. Moreover, if multiple exceptions occur in one logical process (like what `cider-test` does), you can select which one you want to display, and revisit ones you've previously seen and dismissed.

Of course, you can certainly use the `stacktrace` functions from other middleware. No need to create a new util namespace. This is precisely what `cider-test` does (see [here](https://github.com/clojure-emacs/cider-nrepl/blob/95436cb237438f14a8680bfa371255af1294a76c/src/cider/nrepl/middleware/test.clj#L257)), which might be a good template for what you're thinking.

sanjayl:
Thanks for the reply Jeff! It definitely helped shine a light on my understanding of `stacktrace` and `test`. I'll have to dig in a bit deeper, been kinda avoiding those portions of the codebase and sticking with the superficial stuff.

bbatsov:
@sanjayl Guess you can update the PR with your new branch, so we can start doing the review. In the mean time I can add some notes there.

sanjayl:
updated version with the "storage" aspect ripped out added as a new PR since this one was opened under the guise of adding storage. Can reject this PR with extreme prejudice

bbatsov:
OK, closing this one. I'll try to review the new PR later today.

PR #325, opened 2016-03-25
quote project dir when used in regex

jgeraerts:

- [ ] All tests are passing

Thanks!

Since windows uses backslashes this results in an invalid regular
expression throwing a PatternSyntaxException. By passing the
project-root through Pattern/quote it is escaped properly for use in
a regular expression

### DIFF
--- a/src/cider/nrepl/middleware/util/namespace.clj
+++ b/src/cider/nrepl/middleware/util/namespace.clj
@@ -28,8 +28,7 @@
   "Find all namespaces defined in source paths within the current project."
   []
   (->> (cp/classpath-directories)
-       (filter #(re-find (re-pattern (str "^" project-root))
-                         (str %)))
+       (filter #(.startsWith (str %) project-root))
        (mapcat ns-find/find-namespaces-in-dir)))
 
 (defn inlined-dependency?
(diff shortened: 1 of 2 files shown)


src/cider/nrepl/middleware/util/namespace.clj
```diff
@@ -24,11 +25,15 @@
   (str (System/getProperty "user.dir")
        (System/getProperty "file.separator")))
 
+(defn- re-quote
```
Malabarba: I think there's no need for this function. We can just use Pattern/quote directly.

bbatsov: Indeed.

bbatsov:
You should capitalise the first word in the commit's title (quote -> Quote). In the commit's description the final sentence is missing its `.`.

jgeraerts:
Commit updated with requests.

jgeraerts:
I was also wondering if I could use `#(String/startsWith (str %) project-root)` instead of the whole regex setup.

bbatsov:
> I was also wondering if I could use #(String/startsWith (str %) project-root) instead of the whole regex setup.

`String` has a static method `startsWith`? In general yeah - the simpler the solution the better.

jgeraerts:
no sorry it's wrong. it should be `(.startsWith .... )`.

bbatsov:
As mentioned above - I believe this simpler approach will work just fine, so feel free to update the PR.

jgeraerts:
i've replace the code with `(.startsWith ... )` construct.

bbatsov:
One final remark - now the title of the commit doesn't correspond with the actual change.

jgeraerts:
done

bbatsov:
:+1: Thanks!

Malabarba:
Thanks!

Issue #313, opened 2016-03-07
compliment can cause cider-nrepl to err out


expez:
## Expected behavior

Errors in `compliment` only result in missing completions

## Actual behavior

Input is interrupted and repl is filled with stacktracke: 

```
java.lang.NullPointerException: null
 at java.util.concurrent.ConcurrentHashMap.get (ConcurrentHashMap.java:936)
    clojure.lang.Namespace.find (Namespace.java:188)
    clojure.core$find_ns.invoke (core.clj:3976)
    clojure.core$the_ns.invoke (core.clj:4008)
    clojure.core$ns_map.invoke (core.clj:4022)
    cider.inlined_deps.compliment.v0v2v7.compliment.sources.ns_mappings$candidates.invoke (ns_mappings.clj:53)
    clojure.lang.Var.invoke (Var.java:388)
    cider.inlined_deps.compliment.v0v2v7.compliment.core$eval52698$completions__52701$iter__52703__52707$fn__52708.invoke (core.clj:95)
    clojure.lang.LazySeq.sval (LazySeq.java:40)
    clojure.lang.LazySeq.seq (LazySeq.java:49)
    clojure.lang.RT.seq (RT.java:507)
    clojure.core/seq (core.clj:137)
    clojure.core$tree_seq$walk__5045$fn__5046.invoke (core.clj:4739)
    clojure.lang.LazySeq.sval (LazySeq.java:40)
    clojure.lang.LazySeq.seq (LazySeq.java:49)
    clojure.lang.LazySeq.more (LazySeq.java:85)
    clojure.lang.RT.more (RT.java:683)
    clojure.core/rest (core.clj:73)
    clojure.core$flatten.invoke (core.clj:6851)
    cider.inlined_deps.compliment.v0v2v7.compliment.core$eval52698$completions__52701.invoke (core.clj:91)
    cider.nrepl.middleware.complete$complete.invoke (complete.clj:19)
    cider.nrepl.middleware.complete$complete_reply.invoke (complete.clj:34)
    cider.nrepl.middleware.complete$wrap_complete$fn__53008.invoke (complete.clj:55)
    clojure.tools.nrepl.middleware$wrap_conj_descriptor$fn__40282.invoke (middleware.clj:22)
    cider.nrepl.middleware.ns$wrap_ns$fn__51752.invoke (ns.clj:110)
[...]

PR #327, opened 2016-03-31
Abstract away op error handling

sanjayl:

- [x ] The commits are consistent with our [contribution guidelines](CONTRIBUTING.md)
- [ x] You've added tests to cover your change(s)
- [ x] All tests are passing
- [ x] The new code is not generating reflection warnings

Thanks!

A new util ns in `cider.nrepl.middleware.util.error-handling` which
provides the `with-safe-transport` macro, abstracting away transport,
error handling, and stacktrace analysis. Most middlewares were updated
to use this system, and the tests were updated to reflect these changes.

In order to utilize the nicely formatted stacktraces stuffed into these
responses, concurrent changes have been made to CIDER and will be
submitted as a PR as well.

### DIFF
--- a/src/cider/nrepl/middleware/inspect.clj
+++ b/src/cider/nrepl/middleware/inspect.clj
@@ -1,5 +1,6 @@
 (ns cider.nrepl.middleware.inspect
   (:require [cider.nrepl.middleware.util.cljs :as cljs]
+            [cider.nrepl.middleware.util.error-handling :refer [base-error-response]]
             [cider.nrepl.middleware.util.inspect :as inspect]
             [cider.nrepl.middleware.util.misc :as u]
             [clojure.tools.nrepl.middleware :refer [set-descriptor!]]
@@ -47,77 +48,35 @@
   [handler msg]
   (handler (eval-msg msg)))
 
-(defn pop-reply
-  [{:keys [transport] :as msg}]
-  (try
-    (let [inspector (swap-inspector! msg inspect/up)]
-      (transport/send
-       transport
-       (response-for msg :value (:rendered inspector) :status :done)))
-    (catch Exception e
-      (transport/send
-       transport
-       (response-for msg (u/err-info e :inspect-pop-error))))))
+(defn- success [{:keys [transport] :as msg} inspector]
+  (transport/send transport (response-for msg :value (:rendered inspector) :status :done)))
 
-(defn push-reply
-  [{:keys [idx transport] :as msg}]
-  (try
-    (let [inspector (swap-inspector! msg inspect/down idx)]
-      (transport/send
-       transport
-       (response-for msg :value (:rendered inspector) :status :done)))
-    (catch Exception e
-      (transport/send
-       transport
-       (response-for msg (u/err-info e :inspect-push-error))))))
+(defn- failure [{:keys [transport] :as msg} err err-kw]
+  (transport/send transport (base-error-response msg err err-kw :done)))
 
-(defn refresh-reply
-  [{:keys [transport] :as msg}]
-  (try
-    (let [inspector (swap-inspector! msg #(or % (inspect/fresh)))]
-      (transport/send
-       transport
-       (response-for msg :value (:rendered inspector) :status :done)))
-    (catch Exception e
-      (transport/send
-       transport
-       (response-for msg (u/err-info e :inspect-refresh-error))))))
+(defn pop-reply [msg]
+  (try (success msg (swap-inspector! msg inspect/up))
+       (catch Exception e (failure msg e :inspect-pop-error))))
 
-(defn next-page-reply
-  [{:keys [transport] :as msg}]
-  (try
-    (let [inspector (swap-inspector! msg inspect/next-page)]
-      (transport/send
-       transport
-       (response-for msg :value (:rendered inspector) :status :done)))
-    (catch Exception e
-      (transport/send
-       transport
-       (response-for msg (u/err-info e :inspect-next-page-error))))))
+(defn push-reply [msg]
+  (try (success msg (swap-inspector! msg inspect/down (:idx msg)))
+       (catch Exception e (failure msg e :inspect-push-error))))
 
-(defn prev-page-reply
-  [{:keys [transport] :as msg}]
-  (try
-    (let [inspector (swap-inspector! msg inspect/prev-page)]
-      (transport/send
-       transport
-       (response-for msg :value (:rendered inspector) :status :done)))
-    (catch Exception e
-      (transport/send
-       transport
-       (response-for msg (u/err-info e :inspect-prev-page-error))))))
+(defn refresh-reply [msg]
+  (try (success msg (swap-inspector! msg #(or % (inspect/fresh))))
+       (catch Exception e (failure msg e :inspect-refresh-error))))
 
-(defn set-page-size-reply
-  [{:keys [page-size transport] :as msg}]
-  (try
-    (let [inspector (swap-inspector! msg inspect/set-page-size page-size)]
-      (transport/send
-       transport
-       (response-for msg :value (:rendered inspector) :status :done)))
-    (catch Exception e
-      (transport/send
-       transport
-       (response-for msg (u/err-info e :inspect-set-page-size-error))))))
+(defn next-page-reply [msg]
+  (try (success msg (swap-inspector! msg inspect/next-page))
+       (catch Exception e (failure msg e :inspect-next-page-error))))
+
+(defn prev-page-reply [msg]
+  (try (success msg (swap-inspector! msg inspect/prev-page))
+       (catch Exception e (failure msg e :inspect-prev-page-error))))
+
+(defn set-page-size-reply [msg]
+  (try (success msg (swap-inspector! msg inspect/set-page-size (:page-size msg)))
+       (catch Exception e (failure msg e :inspect-set-page-size-error))))
 
 (defn wrap-inspect
   "Middleware that adds a value inspector option to the eval op. Passing a
--- a/src/cider/nrepl/middleware/util/error_handling.clj
+++ b/src/cider/nrepl/middleware/util/error_handling.clj
@@ -1,0 +1,135 @@
+(ns cider.nrepl.middleware.util.error-handling
+  "Utilities to safely reply to op requests and help deal with the
+  errors/exceptions that might arise from doing so."
+  (:refer-clojure :exclude [error-handler])
+  (:require [clojure.set :as set]
+            [clojure.stacktrace :as stacktrace]
+            [clojure.tools.nrepl.transport :as transport]
+            [clojure.tools.nrepl.misc :refer [response-for]]
+            [cider.nrepl.middleware.stacktrace :as cider-stacktrace]))
+
+;;; UTILITY FUNCTIONS
+
+(defn error-summary
+  "Takes a `java.lang.Exception` as `ex` and returns a map summarizing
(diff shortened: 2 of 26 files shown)


src/cider/nrepl/middleware/util/error_handling.clj
```diff
@@ -0,0 +1,137 @@
...
+  head being the op-action and the tail being the error-action, or if
+  the default error handler is sufficient, then the op name can be
+  paired directly to the op-action.
+
+  Actions can be functions, maps, non-associate collections, and
+  single items such as kw's, strings, numbers, etc. The utilization of
+  each type is discussed above in the `selector` method."
+  [pass-through & pairings]
+  `(fn [{op# :op transport# :transport :as msg#}]
+     (try
```
Malabarba: While I understand the intent behind error-handling the error-handler itself, this will cause a stack of dozens of nested `try` blocks. Maybe that's ok. But then, maybe we don't need this outer `try` block either.

Malabarba:
Thanks, I like this implementation.
Looks like, you've done a lot more than error handling. :smile: From what I understand, you've written an error handler and an `op-handler`.

The `op-handler` (which I like) does actually need that `with-safe-transport` macro. 
But the error-handler (the exception catching code) could have been done in a middleware, outside that macro. In fact, I think I would still prefer if the exception handling was done by a middleware, this way it would also benefit some our ops that don't yet take advantage of the `with-safe-transport` macro, and it could even benefit middlewares from other plugins (like refactor-nrepl).
Here's what the middleware would look like:

    (defn wrap-errors [h]
      (fn [msg]
        (try
          (h msg)
          (catch Exception e
            (transport/send (:transport msg)
                            (error-handler :default msg e))))))
    
    (set-descriptor! #'wrap-errors {})

The only thing we lose from this implementation is the ability to specify a different error status (currently I see `:toggle-trace-error` and `:macroexpansion-error` are using this). However, I believe these custom error statuses were only introduced because we lacked proper error-handling in the middlewares. So now that we have it we don't need them anymore.

bbatsov:
I agree with @Malabarba about the middleware for pretty much the same reasons he outlined. This would also protect us from internal errors in nREPL itself.  

> The only thing we lose from this implementation is the ability to specify a different error status (currently I see :toggle-trace-error and :macroexpansion-error are using this). However, I believe these custom error statuses were only introduced because we lacked proper error-handling in the middlewares. So now that we have it we don't need them anymore.

What exactly is different there? The code looks exactly the same as in other middleware to me.

src/cider/nrepl/middleware/inspect.clj
```diff
@@ -47,77 +48,35 @@
...
-       (response-for msg (u/err-info e :inspect-set-page-size-error))))))
+(defn- success [{:keys [transport] :as msg} inspector]
+  (transport/send transport (response-for msg :value (:rendered inspector) :status :done)))
+
+(defn- failure [{:keys [transport] :as msg} err err-kw]
+  (transport/send transport (base-error-response msg err err-kw :done)))
+
+(defn pop-reply [msg]
+  (try (success msg (swap-inspector! msg inspect/up))
+       (catch Exception e (failure msg e :inspect-pop-error))))
```
bbatsov: I wonder if the Elisp code cares about those different error codes. Seems to me there's still room to cleanup/simplify all those ops.

src/cider/nrepl/middleware/inspect.clj
```diff
@@ -47,77 +48,35 @@
...
+
+(defn prev-page-reply [msg]
+  (try (success msg (swap-inspector! msg inspect/prev-page))
+       (catch Exception e (failure msg e :inspect-prev-page-error))))
+
+(defn set-page-size-reply [msg]
+  (try (success msg (swap-inspector! msg inspect/set-page-size (:page-size msg)))
+       (catch Exception e (failure msg e :inspect-set-page-size-error))))
 
 (defn wrap-inspect
```
bbatsov: Was this not updated because of the `eval` in it?

sanjayl:
> Thanks, I like this implementation.
Looks like, you've done a lot more than error handling. :smile: From what I understand, you've written an error handler and an op-handler.

Thanks Artur...yeah, I *might* have taken some liberties with mission creep as I went through the code base. Probably time to figure out how to selectively commit hunks in magit :laughing:.

>  In fact, I think I would still prefer if the exception handling was done by a middleware, this way it would also benefit some our ops that don't yet take advantage of the with-safe-transport macro, and it could even benefit middlewares from other plugins (like refactor-nrepl).

> I agree with @Malabarba about the middleware for pretty much the same reasons he outlined. This would also protect us from internal errors in nREPL itself.

These are really good points. A couple questions regarding how to implement this: Is there a way to ensure that the error-middleware is the outermost middleware, or will it have to be part of convention that the error-middleware gets preferential ordering in the stack? What about when we're dynamically injecting middleware (such as when refactor-nrepl gets injected on jack-in), can we guarantee that will be injected within the error-middleware?

> What exactly is different there? The code looks exactly the same as in other middleware to me.

RIght now it looks the same because virtually none of the middleware is doing anything unique with regards to error handling, but having the error handlers defined at the level of the "op -> action" mapping affords a simple mechanism for op-specific custom error handling. This would still be possible but would be a bit more cumbersome if all exception handling was done within a catch-all middleware (ie, some lookup table that the error-middleware could use to resolve a custom error handler given the information in the `msg` and `Exception`).

In my mind, there should be a 3 tiered system:

1. Middleware "in the know" would be able to deal with ops and their errors all in the same place for the vast majority of operations by using the `with-safe-transport` macro. 
2. The `base-error-response` function allows for standardized error handling for those small number of cases where the default macro isn't appropriate. (much like how `util.misc/err-info` is currently used...which is incidentally what `refactor-nrepl` uses for its error handling and why I couldn't remove that function entirely)
3. The middleware would serve as a catch all and deal with cases where error-handling wasn't considered or is out of our control

Thoughts on this layered approach? Certainly would be simpler to only use a middleware, but I don't have a great feeling about dealing with ops and their errors in 2 very different places.

src/cider/nrepl/middleware/inspect.clj
sanjayl: It's more because of how the `inspect-reply` function handles creating a response. It writes a non-terminated response and expects the {:status #{:done}} cap to happen deeper down in the handler. The macro expects a response that can be immediately terminated. I didn't feel like mucking around with this too much because I didn't understand it well.

bbatsov:
> RIght now it looks the same because virtually none of the middleware is doing anything unique with regards to error handling, but having the error handlers defined at the level of the "op -> action" mapping affords a simple mechanism for op-specific custom error handling.

That's a good point as well. There's something else to consider, though - you can have several different error codes per one op if several different things can go wrong. In such cases (not sure if we have them in the current codebase or not) you'd still have to roll out some custom error handling code. 

Anyways, I guess I'd be fine with both approaches - it's obviously that they are both superior to what we're doing right now...

src/cider/nrepl/middleware/inspect.clj
bbatsov: I see. Probably we should refactor this down the road. It's a pretty old middleware, that hasn't be updated significantly in quite a while.

src/cider/nrepl/middleware/inspect.clj
sanjayl: You're right, the Elisp code has **absolutely no idea** about these error codes. They could be removed entirely, but I think they're good to have so that we can "tune" the error reporting UI based on the type of error returned (which is something I've been experimenting with). Eval/big errors should certainly get the stacktrace viewer, some other types should just get a minibuffer message, and other types of errors might just get swallowed. The user would also be able to customize this by fooling around with some variables.

I think there are definitely a lot of places to clean up the op handling code across the board. The biggest gain would probably involve using the "inline-response map" feature of the macro and getting rid of all these `<op>-reply` functions that now basically don't do anything. I think this would really give us a clear line between interface and implementation as well.

project.clj
```diff
@@ -56,7 +56,8 @@
                       :dependencies [[org.clojure/clojure "1.8.0-master-SNAPSHOT"]]}
 
              :test-clj {:test-paths ["test/clj"]
-                        :java-source-paths ["test/java"]}
+                        :java-source-paths ["test/java"]
+                        :resource-paths ["test/resources"]}
```
bbatsov: Why is this needed? I don't see anything related in the PR (guess I might be missing something).

sanjayl:
> That's a good point as well. There's something else to consider, though - you can have several different error codes per one op if several different things can go wrong. In such cases (not sure if we have them in the current codebase or not) you'd still have to roll out some custom error handling code.

Hmmm, interesting. Do you mean that a single op would travel down the middleware stack and throw errors in multiple middlewares?

project.clj
sanjayl: Sorry, not related to the PR. When I was going through the code base, I saw that `resouce-test` wasn't actually testing anything -- it was in effect doing something like asserting `(= nil nil)` -- because the `test.txt` resource didn't have its enclosing directory added to the resource path.

src/cider/nrepl/middleware/util/error_handling.clj
sanjayl: Oh wow I didn't even think about how ugly this would look once macroexpanded. The outer `try` is needed so you can do the inline responses safely (say for something that has a constant reply), but we'll probably never use this and could remove it entirely.

bbatsov:
> Do you mean that a single op would travel down the middleware stack and throw errors in multiple middlewares?

I meant one op could potentially throw different errors in different situations. Right now we have just one error per op, but that mostly because we don't really process the errors on the client-side. E.g. there can be 5 different macroexpansion errors. Not a big deal right, just some food for thought.

src/cider/nrepl/middleware/util/error_handling.clj
bbatsov: Yeah, let's do this.

project.clj
bbatsov: I see. Guess this should be fixed in a separate PR.

sanjayl: OK, just submitted a PR for this, will rebase this branch off master if it gets merged in

bbatsov:
Guess you'll have to do that rebase. :-)

sanjayl:
Done! Took out the "outer try" from the macro and rebased after the `resource-test` fix

Malabarba:
> These are really good points. A couple questions regarding how to implement this: Is there a way to ensure that the error-middleware is the outermost middleware, or will it have to be part of convention that the error-middleware gets preferential ordering in the stack? 

I don't know. I know that we could add `:require #{"wrap-errors"}` to every one of our middlewares. But maybe there's a simpler way to specify that `wrap-errors` should be above every other middleware.
@arrdem Would you know of a way to do that?

> What about when we're dynamically injecting middleware, can we guarantee that will be injected within the error-middleware?

I think Cider's dynamic injection should not interfere with this in any way.

Malabarba:
Oops, long night. I meant to ping @cemerick 
Do you know if there's a way to ensure that a midddleware gets placed above all others in nREPL?

bbatsov:
I've decided to merge the PR in its current form. We can always do the middleware change later if we decide that'd be best. Thanks for the huge amount of your you did here @sanjayl! It's much appreciated!

sanjayl:
No problem! I'll be taking a stab at #313, the first part ("mute all middleware errors") definitely sounds straightforward. The "mute this specific middleware error" button sounds like a very interesting UI idea and I'll let you know if it's doable.

PR #329, opened 2016-03-31
Fix exception when instrumenting code with embedded namespace

grammati:
Instrumentation would throw when trying to macroexpand code containing a
namespace object with metadata.

The logging macros in clojure.tools.logging embed the current namespace
object into the code at macroexpansion time. Instrumenting a function
that contained logging calls would throw an exception if the current
namespace had metadata (eg: a docstring).

This is because Namespace has the unusual property of implementing
IMeta (so you can call `(meta *ns*)`), but not IObj (so `(with-meta *ns*
...)` throws).

### DIFF
--- a/src/cider/nrepl/middleware/util/meta.clj
+++ b/src/cider/nrepl/middleware/util/meta.clj
@@ -16,7 +16,9 @@
 
 (defn strip-meta [form]
   (if (meta form)
-    (with-meta form nil)
+    (try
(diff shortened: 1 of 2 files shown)


bbatsov:
:+1: Good catch! Thanks!

You might also want to mention this in CIDER's changelog.

Malabarba:
Thanks indeed!

PR #331, opened 2016-04-03
Add ns and name to eldoc's response

bbatsov:
Hey everyone!

Can you take a look at this code and tell me if something looks off to you (or even better - test it). I'm working on some eldoc improvements but this code is locking up every time I try it from Emacs, although the functions work as expected when tested in a REPL). I'm completely at a loss right now... This looks super simple, but for some reason it's not behaving properly. Guess I did something super silly...

### DIFF
--- a/src/cider/nrepl/middleware/info.clj
+++ b/src/cider/nrepl/middleware/info.clj
@@ -284,15 +284,19 @@
   (map #(mapv str %) raw-eldoc))
 
 (defn eldoc
-  [msg]
-  (if-let [raw-eldoc (extract-eldoc (info msg))]
+  [info]
+  (if-let [raw-eldoc (extract-eldoc info)]
     (format-eldoc raw-eldoc)))
 
 (defn eldoc-reply
   [msg]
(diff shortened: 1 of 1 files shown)


bbatsov:
Hmm, seems the tests are locking up as well after this change. What have I done...

Malabarba:
Is :ns a string? You'll get hangs if you try to send Clojure objects over a key other than :value. 
You have to print it first.

bbatsov:
It's not. Guess this is the problem... I wonder why I didn't get a meaningful error, though...

Malabarba:
You just don't. I had similar hangs with the state_tracker. I think I remember raging about it on gitter. :-)

bbatsov:
@Malabarba this was the problem indeed. 

P.S. We really need to start raising errors for this. I lost 1 hour wondering what was wrong...

Malabarba:
Yes, we really do =/

bbatsov:
I'll file a separate ticket for this.

Issue #332, opened 2016-04-03
Returning an non-primitive object from a middleware locks up the middleware


bbatsov:
I'm not quite certain whether this is a cider or a cider-nrepl issue, but I'm reporting it here, so we won't forget about it.

## Expected behavior

Something like this should raise an error.

## Actual behavior

Everything locks up, no errors are raised.

## Steps to reproduce the problem

Edit the `eldoc-reply` method to return an ns object, instead of a string and try running eldoc in CIDER.

More info - https://github.com/clojure-emacs/cider-nrepl/pull/331#issuecomment-205024081 

### cider-nrepl version

0.12-snapshot

### Java version

1.8

### Operating system

OS X El Capitan

sanjayl:
I used `tcpflow` to capture the communication across nREPL's port, and it looks like the cider-nrepl side is never generating a reply. (the command I used was `sudo tcpflow -i any -C -J port <nrepl-port-number>`).

I also used the java console to see if there are any hung threads on the server side, but it doesn't look like there are.

I'm guessing it's going to be on the tools.nrepl side of things? I could probably change the `with-safe-transport` macro to `pr` all the non-primitive values in the return map. I have a busy week at work this week so not sure if I'll be able to try it out, but I'll keep you updated if I do get a chance.

bbatsov:
Thanks for looking into this @sanjayl!

sanjayl:
Couple of quick questions:

1. Would you prefer that the non-bencodable items causing this type of error get automatically converted to something bencodable (likely a string), or would you prefer an error bubble up so that we're not silently transforming "out of bounds" data?
2. Do you know if tools.nrepl is actively being maintained and accepting patches, or should I just focus on putting a workaround in the cider-nrepl code first?

bbatsov:
> Would you prefer that the non-bencodable items causing this type of error get automatically converted to something bencodable (likely a string), or would you prefer an error bubble up so that we're not silently transforming "out of bounds" data?

Let's raise an error.

> Do you know if tools.nrepl is actively being maintained and accepting patches, or should I just focus on putting a workaround in the cider-nrepl code first?

Yeah, it's still maintained by @cemerick. You'll have to submit the patch via nREPL's JIRA, though.

bbatsov:
P.S. In general it's best to fix everything that can be fixed in nREPL in nREPL (as this benefits all nREPL users). Unfortunately it often takes a while to merge something there, so this is not always feasible in the short term.

[...]

PR #330, opened 2016-04-01
add transform-value dispatch value for BigDecimal

jgeraerts:
Before submitting a PR make sure the following things have been done:

- [ ] The commits are consistent with our [contribution guidelines](CONTRIBUTING.md)
- [ ] You've added tests to cover your change(s)
- [ ] All tests are passing
- [ ] The new code is not generating reflection warnings
- [ ] You've updated the readme (if adding/changing middleware)

Thanks!

BigDecimal was handled as a Number returning the initial value
causing an IllegalStateException in clojure.tools.nrepl.bencode. This
patch adds an additional multimethod dispatching on BigDecimal making
it a string.

PR #333, opened 2016-04-07
Throw error if sending non-bencodable message

sanjayl:

- [x] All tests are passing

Thanks!

Here's a first draft of something that will report back an error if the situation in #332 is detected. I tried fixing it directly in tools.nrepl, but I was having a tough time working through that code. This seems a bit hacky, but I think it should work. Let me know if you have more elegant ideas on working around this (ways to fix the issue in tools.nrepl, making our own transport, etc).

This tries to work around the hung communication related to sending
non-bencodable information through the standard Transport. If
non-bencodable information is currently sent in the reply message, it
can only be assigned to the value to the :value key, otherwise, the
Transport.send will silently fail and seize up. This patch will throw an
error to the user if this type of situation is detected.


jgeraerts:
This would also handle #330 gracefully.

src/cider/nrepl/middleware/util/error_handling.clj
```diff
@@ -70,6 +73,37 @@
         (coll? input)  :status-coll
         (keyword? input) :status-item))
 
+(defn- shallow-bencodable?
+  "Returns false if `item`'s type can't be bencoded as defined by the
+  algorithm in `clojure.tools.nrepl.bencode/write-bencode`. Does not
+  traverse the subelements of a collection to ensure that the
```
bbatsov: subelements? Should this be just elements?

src/cider/nrepl/middleware/util/error_handling.clj
```diff
@@ -97,6 +131,7 @@
 
 (defmethod op-handler :inline-reply [answer msg]
   (let [terminal-status (set/union #{:done} (normalize-status (:status answer)))]
+    (deep-bencodable-or-fail (dissoc answer :value))
```
bbatsov: Probably you should add a comment explaining this.

src/cider/nrepl/middleware/util/error_handling.clj
```diff
@@ -70,6 +73,37 @@
...
+    (string? item)  :string
+    (symbol? item)  :named
+    (keyword? item) :named
+    (map? item)     :map
+    (or (nil? item) (coll? item) (.isArray (class item))) :list
+    :else false))
+
+(defn- deep-bencodable-or-fail
+  "Walks through the object provided by `item` and returns true if
+  the entity and all subelements are bencodable as defined by the
```
bbatsov: subelements -> elements

bbatsov:
Seems like a reasonable approach, although in the long term this should still be addressed in nREPL. @cemerick Any thoughts?

src/cider/nrepl/middleware/util/error_handling.clj
sanjayl: Done

src/cider/nrepl/middleware/util/error_handling.clj
sanjayl: Done

bbatsov:
@sanjayl Thanks! Guess you can also file an upstream ticket, so we don't forget about an upstream solution to the problem.

sanjayl:
Done! http://dev.clojure.org/jira/browse/NREPL-81

PR #335, opened 2016-04-11
Debugger: Step in

grammati:
This commit implements "step in" for the debugger.

When stepping past a form that is a call to a function stored in a var,
prompt the user with the option to step in to the called function.

The prompt takes the form "Step in to foo/bar?", where "foo/bar" is the
function as it appears in the source, and the options are "yes" and "no.

Selecting "no" will display the result of the function, just as it
always has. Note that because the key for "no" is the same as the key
for "next", those in the habit of rapidly hitting "n" will notice little
difference to their workflow, although more presses of "n" will be
required - one extra for each function call.

Selecting "yes" will instrument the called function and then step in to
it. The instrumentation is done in a special way. It is not simply
instrumented in the normal way, because that would cause all subsequent
calls to it to trigger the debugger. Instead, an instrumented version of
the function is compiled and put into the var's metadata, and this
instrumented version is explicitly called in response to the "yes"
command.

Most normal functions can be stepped in to, including both those in
source files and in jars.

Limitations:

The debugger needs to be able to find on-disk source code for a function
in order to instrument it. Functions that have been eval'ed, but not
saved, cannot be stepped into.

Initially, I have disabled the "step in" prompt for functions in
clojure.core, and enabled it for everything else. This could be made
more customizable in the future, perhaps with additional options to the
"step in?" prompt, such as "never", "always", etc.

Only functions that can be resolved to a var can be stepped in to.
Stepping in to higher-order functions does not work (eg: `((:f m) x)`).

Opportunities:

This implements most of the ability to instrument functions without
having to send the source code over from emacs - the var name is enough.
Perhaps we could instrument functions from the namespace browser, or
from a stacktrace view (eg: instrument all the functions that are part
of my project and appear in this stacktrace).

### DIFF
--- a/src/cider/nrepl/middleware/debug.clj
+++ b/src/cider/nrepl/middleware/debug.clj
@@ -294,6 +294,20 @@
                               last :stacktrace)}]})
   (read-debug-command value extras))
 
+(def debug-commands
+  {"c" :continue
+   "e" :eval
+   "h" :here
+   "i" :in
+   "j" :inject
+   "l" :locals
+   "n" :next
+   "o" :out
+   "p" :inspect
+   "q" :quit
+   "s" :stacktrace
+   "t" :trace})
+
 (defn read-debug-command
   "Read and take action on a debugging command.
   Ask for one of the following debug commands using `read-debug`:
@@ -314,8 +328,8 @@
   a :code entry, its value is used for operations such as :eval, which
   would otherwise interactively prompt for an expression."
   [value extras]
-  (let [commands (cond->> [:next :in :continue :out :here :inspect :locals :inject :eval :stacktrace :trace :quit]
-                   (not (map? *msg*)) (remove #{:quit})
+  (let [commands (cond->> debug-commands
+                   (not (map? *msg*)) (dissoc "q")
                    (cljs/grab-cljs-env *msg*) identity)
         response-raw (read-debug extras commands nil)
 
--- a/src/cider/nrepl/middleware/debug.clj
+++ b/src/cider/nrepl/middleware/debug.clj
@@ -1,20 +1,21 @@
 (ns cider.nrepl.middleware.debug
   "Expression-based debugger for clojure code"
   {:author "Artur Malabarba"}
-  (:require [cider.nrepl.middleware.inspect :refer [swap-inspector!]]
+  (:require [cider.nrepl.middleware.info :as info]
+            [cider.nrepl.middleware.inspect :refer [swap-inspector!]]
             [cider.nrepl.middleware.pprint :as pprint]
             [cider.nrepl.middleware.stacktrace :as stacktrace]
             [cider.nrepl.middleware.util.cljs :as cljs]
             [cider.nrepl.middleware.util.inspect :as inspect]
             [cider.nrepl.middleware.util.instrument :as ins]
-            [cider.nrepl.middleware.util.meta :as m]
             [cider.nrepl.middleware.util.misc :as misc]
+            [clojure.java.io :as io]
             [clojure.tools.nrepl.middleware :refer [set-descriptor!]]
             [clojure.tools.nrepl.middleware.interruptible-eval :refer [*msg*]]
             [clojure.tools.nrepl.misc :refer [response-for]]
             [clojure.tools.nrepl.transport :as transport]
-            [clojure.walk :as walk])
-  (:import [clojure.lang Compiler$LocalBinding]))
+            [cider.nrepl.middleware.util.misc :as u])
+  (:import [clojure.lang Compiler$LocalBinding LineNumberingPushbackReader]))
 
 (defn random-uuid-str []
   (letfn [(hex [] (format "%x" (rand-int 15)))
@@ -190,6 +191,8 @@
 (def print-length (atom nil))
 (def print-level (atom nil))
 
+(defonce step-in-to-next? (atom false))
+
 (defn pr-short
   "Like `pr-str` but limited in length and depth."
   [x]
@@ -311,7 +314,7 @@
   a :code entry, its value is used for operations such as :eval, which
   would otherwise interactively prompt for an expression."
   [value extras]
-  (let [commands (cond->> [:next :continue :out :here :inspect :locals :inject :eval :stacktrace :trace :quit]
+  (let [commands (cond->> [:next :in :continue :out :here :inspect :locals :inject :eval :stacktrace :trace :quit]
                    (not (map? *msg*)) (remove #{:quit})
                    (cljs/grab-cljs-env *msg*) identity)
         response-raw (read-debug extras commands nil)
@@ -320,8 +323,10 @@
         (if (map? response-raw) response-raw
             {:response response-raw})
         extras (dissoc extras :inspect)]
+    (reset! step-in-to-next? false)
     (case response
       :next     value
+      :in       (do (reset! step-in-to-next? true) value)
       :continue (do (skip-breaks! :all) value)
       :out      (do (skip-breaks! :deeper (butlast (:coor extras)) (:code extras) force?)
                     value)
@@ -343,6 +348,110 @@
             *print-level*  2]
     (pr form))
   (println "=>" (pr-short value)))
+
+(defn read-var
+  "Find the source of the var `v`.
+  Return a map of the var's metadata (:file, :line, :column, etc.) as well as:
+    - :form : The form, as read by `clojure.core/read`, and
+    - :code : The source code of the form
+  Return nil if the source of the var cannot be found."
+  ;; TODO - maybe this function should go in info.clj?
+  [v]
+  (when-let [{:keys [file line column] :as var-meta} (info/var-meta v)]
+    ;; file can be either absolute (eg: functions that have been eval-ed with
+    ;; C-M-x), or relative to some path on the classpath.
+    (when-let [res (or (io/resource file)
+                       (let [f (io/file file)]
+                         (when (.exists f)
+                           f)))]
+      (with-open [rdr (LineNumberingPushbackReader. (io/reader res))]
+        ;; Skip to the right line
+        (dotimes [_ (dec line)]
+          (.readLine rdr))
+
+        ;; Reader that collects the code as a string. Adapted from
+        ;; clojure.repl/source.
+        (let [text     (StringBuilder.)
+              collect? (atom false)
+              pbr      (proxy [LineNumberingPushbackReader] [rdr]
+                         (read []
+                           (let [i (proxy-super read)]
+                             (when @collect?
+                               (.append text (char i)))
+                             i))
+                         (unread [c]
+                           (when @collect?
+                             (.deleteCharAt text (dec (.length text))))
+                           (proxy-super unread c)))
+              ;; Fix bogus column number of 1, which really means 0
+              column   (if (= 1 column) 0 column)]
+
+          ;; Give it the right line and column number. We can just set the
+          ;; line number directly, but there is no setColumnNumber method, so
+          ;; we have to hack it a bit.
+          (.setLineNumber pbr (.getLineNumber rdr))
+          (dotimes [_ column]
+            (.read pbr))
+
+          ;; Now start collecting the code
+          (reset! collect? true)
+
+          (let [form (read {} pbr)
+                code (str text)]
+            (assoc var-meta
+                   :form form
+                   :code code)))))))
+
+(declare debug-reader)
+
+(defn instrument-var-for-step-in
+  "Attach an instrumented version of the function in `v` as metadata to `v`,
+  leaving the contents of the var uninstrumented."
+  [v]
+  (when-not (::instrumented (meta v))
+    (when-let [{:keys [ns file form] :as var-meta} (read-var v)]
+      (let [full-path (misc/transform-value (:file (info/file-info file)))]
+        (binding [*ns*   ns
+                  *file* file
+                  *msg*  (-> *msg*
+                             (merge var-meta)
+                             (assoc :file full-path))]
+          (-> form
+              debug-reader
+              ins/instrument-tagged-code
+              eval)
+          (let [instrumented @v]
+            (eval form)
+            (alter-meta! v assoc ::instrumented instrumented)))))))
+
+(defn step-in?
+  "Return true if we can and should step in to the function in the var `v`.
+  The \"should\" part is determined by the value in `step-in-to-next?`, which
+  gets set to true by the user sending the \"step in\" command."
+  [v coor]
+  (when (and @step-in-to-next?
+             (not (skip-breaks? coor))
+             (not= :trace (:mode @*skip-breaks*))
+             (not (::ins/breakfunction (meta v))))
+    (try
+      (instrument-var-for-step-in v)
+      true
+      (catch Exception e
+        ;; TODO - how do we inform the user that we failed?
+        false))))
+
+(defn looks-step-innable?
+  "Decide whether a form looks like a call to a function that we could
+  instrument and step into."
+  [form]
+  (when (and (seq? form) (symbol? (first form)))
+    (let [v (resolve (first form))      ; Note: special forms resolve to nil
+          m (meta v)]
+      ;; We do not go so far as to actually try to read the code of the function
+      ;; at this point, which is at macroexpansion time.
+      (and v
+           (not (:macro m))
+           (not (:inline m))))))
 
 ;;; ## High-level functions
 (defmacro with-debug-bindings
@@ -372,18 +481,29 @@
      ~@body))
 
 (defmacro breakpoint-implementation [form {:keys [coor]} original-form]
-  `(let [val# ~form]
-     (cond
-       (skip-breaks? ~coor) val#
-       ;; The length of `coor` is a good indicator of current code
-       ;; depth.
-       (= (:mode @*skip-breaks*) :trace)
-       (do (print-step-indented ~(count coor) '~original-form val#)
-           val#)
-       ;; Nothing special. Here's the actual breakpoint logic.
-       :else (->> (pr-short val#)
-                  (assoc *extras* :debug-value)
-                  (read-debug-command val#)))))
+  (let [val-form (if (looks-step-innable? form)
+                   (let [[fn-sym & args] form]
+                     ;; Evaluate args first, delaying the decision of whether or
+                     ;; not to step in until the last possible moment.
+                     `(let [args# [~@args]
+                            step-in?# (step-in? (var ~fn-sym) ~coor)]
+                        (apply (if step-in?#
+                                 (::instrumented (meta (var ~fn-sym)))
+                                 ~fn-sym)
+                               args#)))
+                   form)]
+    `(let [val# ~val-form]
+       (cond
+         (skip-breaks? ~coor) val#
+         ;; The length of `coor` is a good indicator of current code
+         ;; depth.
+         (= (:mode @*skip-breaks*) :trace)
+         (do (print-step-indented ~(count coor) '~original-form val#)
+             val#)
+         ;; Nothing special. Here's the actual breakpoint logic.
+         :else                (->> (pr-short val#)
+                                   (assoc *extras* :debug-value)
+                                   (read-debug-command val#))))))
 
 (defmacro breakpoint
   "Send the result of form and its coordinates to the client.
--- a/test/clj/cider/nrepl/middleware/debug_integration_test.clj
+++ b/test/clj/cider/nrepl/middleware/debug_integration_test.clj
@@ -3,7 +3,8 @@
             [clojure.test :refer :all]
             [clojure.tools.nrepl :as nrepl]
             [clojure.tools.nrepl.server :as nrepl.server]
-            [clojure.tools.nrepl.transport :as transport])
+            [clojure.tools.nrepl.transport :as transport]
+            [clojure.java.io :as io])
   (:import java.util.UUID
            [java.util.concurrent TimeUnit LinkedBlockingQueue]))
 
@@ -95,6 +96,7 @@
 
 (def-debug-op :next)
 (def-debug-op :continue)
+(def-debug-op :in)
 
 (defmethod debugger-send :out [_ & [force?]]
   (nrepl-send {:op :debug-input :input (str {:response :out :force? force?}) :key (current-key)}))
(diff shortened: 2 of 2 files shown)


grammati:
Front-end part: https://github.com/clojure-emacs/cider/pull/1660

src/cider/nrepl/middleware/debug.clj
```diff
@@ -344,6 +345,112 @@
     (pr form))
   (println "=>" (pr-short value)))
 
+(defn read-var
+  "Find the source of the var `v`.
+  Return a map of the var's metadata (:file, :line, :column, etc.) as well as:
+    - :form : The form, as read by `clojure.core/read`, and
+    - :code : The source code of the form
+  Return nil if the source of the var cannot be found."
+  ;; TODO - maybe this function should go in info.clj?
```
Malabarba: Or in `util/misc.clj`. Which do you prefer @bbatsov?

Malabarba:
Wow. First of all, thanks (again) for the effort. I'll review the code tonight.

Secondly, I'm not 100% behind the final behaviour. Prompting the user at every function call might get annoying fast. I'll play with it a bit, but I think I would prefer Edebug's interface for this.

In Edebug, the user types `i` before a `(`. This will skip all breakpoints until that position, and then step into the function call.
In your implementation that would be equivalent to typing `h` and then answering `yes`.

Still, most of the code seems solid, and some of that stuff will even make it easier to implement some other stuff Bozhidar has been asking for (debug namespace).

grammati:
Yes, I was also concerned about whether prompting to step in to every function would get annoying. So far, I think it seems much less annoying than I had feared, probably because I already have my finger on "n" and pressing it extra times is little effort. But I would love to hear other people's opinions on this.

I chose not to emulate edebug for this because, frankly, I find it limiting in a couple of ways. First, you have to manually move the point - it does not naturally step to a place where you can just step in with a keystroke. Secondly:

"...remember to use this command before any of the arguments are evaluated, since otherwise it will be too late." [edebug manual]

This has always seemed like a pretty severe limitation to me, since I often want to see what some arguments evaluate to before I can decide whether I want to step in or not.

[Aside: what I most often want, really, is a "go back in time" command - that is, after seeing the result of a function call I think "that's not right!", and I wish there was a command to re-run it, but step in this time. But let's save that for another PR :) ]

Anyway, I look forward to hearing you thoughts on this @Malabarba, and thanks again for all your time on reviewing this. I'll probably have some time this week to work on whatever changes are needed.

Malabarba:
> But I would love to hear other people's opinions on this.

Me too. We'll give it a try like this (after 0.12 is released) and see what people think.
The debugger has enough users that we should get good feedback.

> [Aside: what I most often want, really, is a "go back in time" command - that is, after seeing the result of a function call I think "that's not right!", and I wish there was a command to re-run it, but step in this time. But let's save that for another PR :) ]

Yes. I was thinking the same. What I would like the most would be for a `step-in` key to show up on the prompt in the same step that displays the return value.
And it's perfectly plausible to implement in the CIDER debugger.

> Anyway, I look forward to hearing you thoughts on this @Malabarba, 

I've been playing around with it, and I spotted an issue.

1. Define a function at the end of a file.
2. Evaluate it (`C-M-x`) without saving the file.
3. Debug some code that calls that function and answer `y` when prompted to step-in.

You'll get the Clojure exception below.
Obviously, it happens because the reader is trying to read past the end of the file. There's not much you can do to solve this situation, but you should catch the exception and fail gracefully.

    2. Unhandled clojure.lang.LispReader$ReaderException
       java.lang.IllegalArgumentException: Value out of range for char: -1
    
               LispReader.java:  292  clojure.lang.LispReader/read
               LispReader.java:  196  clojure.lang.LispReader/read
               LispReader.java:  185  clojure.lang.LispReader/read
                      core.clj: 3666  clojure.core/read
                      core.clj: 3639  clojure.core/read
                     debug.clj:  395  cider.nrepl.middleware.debug/read-var
                     debug.clj:  348  cider.nrepl.middleware.debug/read-var
                     debug.clj:  408  cider.nrepl.middleware.debug/instrument-var-for-step-in
                     debug.clj:  403  cider.nrepl.middleware.debug/instrument-var-for-step-in
                     debug.clj:  437  cider.nrepl.middleware.debug/step-in?
                     debug.clj:  427  cider.nrepl.middleware.debug/step-in?
    boot.user1598949278402411967.clj:  120  financial-account.service/eval37744/fn
    boot.user1598949278402411967.clj:  120  financial-account.service/eval37744
    boot.user1598949278402411967.clj:   21  financial-account.service/eval37744
                 Compiler.java: 6927  clojure.lang.Compiler/eval
                 Compiler.java: 6890  clojure.lang.Compiler/eval
                      core.clj: 3105  clojure.core/eval
                      core.clj: 3101  clojure.core/eval
                     debug.clj:  567  cider.nrepl.middleware.debug/instrument-and-eval
                     debug.clj:  566  cider.nrepl.middleware.debug/instrument-and-eval
                      Var.java:  379  clojure.lang.Var/invoke
                      main.clj:  240  clojure.main/repl/read-eval-print/fn
                      main.clj:  240  clojure.main/repl/read-eval-print
                      main.clj:  258  clojure.main/repl/fn
                      main.clj:  258  clojure.main/repl
                      main.clj:  174  clojure.main/repl
                   RestFn.java: 1523  clojure.lang.RestFn/invoke
        interruptible_eval.clj:   87  clojure.tools.nrepl.middleware.interruptible-eval/evaluate/fn
                      AFn.java:  152  clojure.lang.AFn/applyToHelper
                      AFn.java:  144  clojure.lang.AFn/applyTo
                      core.clj:  646  clojure.core/apply
                      core.clj: 1881  clojure.core/with-bindings*
                      core.clj: 1881  clojure.core/with-bindings*
                   RestFn.java:  425  clojure.lang.RestFn/invoke
        interruptible_eval.clj:   85  clojure.tools.nrepl.middleware.interruptible-eval/evaluate
        interruptible_eval.clj:   55  clojure.tools.nrepl.middleware.interruptible-eval/evaluate
        interruptible_eval.clj:  222  clojure.tools.nrepl.middleware.interruptible-eval/interruptible-eval/fn/fn
        interruptible_eval.clj:  190  clojure.tools.nrepl.middleware.interruptible-eval/run-next/fn
                      AFn.java:   22  clojure.lang.AFn/run
       ThreadPoolExecutor.java: 1142  java.util.concurrent.ThreadPoolExecutor/runWorker
       ThreadPoolExecutor.java:  617  java.util.concurrent.ThreadPoolExecutor$Worker/run
                   Thread.java:  745  java.lang.Thread/run
    
    1. Caused by java.lang.IllegalArgumentException
       Value out of range for char: -1
    
                       RT.java: 1019  clojure.lang.RT/charCast
                     debug.clj:  376  cider.nrepl.middleware.debug/read-var/fn
                           nil:   -1  cider.nrepl.middleware.debug.proxy$clojure.lang.LineNumberingPushbackReader$ff19274a/read
               LispReader.java:  150  clojure.lang.LispReader/read1
               LispReader.java:  238  clojure.lang.LispReader/read
               LispReader.java:  196  clojure.lang.LispReader/read
               LispReader.java:  185  clojure.lang.LispReader/read
                      core.clj: 3666  clojure.core/read
                      core.clj: 3639  clojure.core/read
                     debug.clj:  395  cider.nrepl.middleware.debug/read-var
                     debug.clj:  348  cider.nrepl.middleware.debug/read-var
                     debug.clj:  408  cider.nrepl.middleware.debug/instrument-var-for-step-in
                     debug.clj:  403  cider.nrepl.middleware.debug/instrument-var-for-step-in
                     debug.clj:  437  cider.nrepl.middleware.debug/step-in?
                     debug.clj:  427  cider.nrepl.middleware.debug/step-in?
    boot.user1598949278402411967.clj:  120  financial-account.service/eval37744/fn
    boot.user1598949278402411967.clj:  120  financial-account.service/eval37744
    boot.user1598949278402411967.clj:   21  financial-account.service/eval37744
                 Compiler.java: 6927  clojure.lang.Compiler/eval
                 Compiler.java: 6890  clojure.lang.Compiler/eval
                      core.clj: 3105  clojure.core/eval
                      core.clj: 3101  clojure.core/eval
                     debug.clj:  567  cider.nrepl.middleware.debug/instrument-and-eval
                     debug.clj:  566  cider.nrepl.middleware.debug/instrument-and-eval
                      Var.java:  379  clojure.lang.Var/invoke
                      main.clj:  240  clojure.main/repl/read-eval-print/fn
                      main.clj:  240  clojure.main/repl/read-eval-print
                      main.clj:  258  clojure.main/repl/fn
                      main.clj:  258  clojure.main/repl
                      main.clj:  174  clojure.main/repl
                   RestFn.java: 1523  clojure.lang.RestFn/invoke
        interruptible_eval.clj:   87  clojure.tools.nrepl.middleware.interruptible-eval/evaluate/fn
                      AFn.java:  152  clojure.lang.AFn/applyToHelper
                      AFn.java:  144  clojure.lang.AFn/applyTo
                      core.clj:  646  clojure.core/apply
                      core.clj: 1881  clojure.core/with-bindings*
                      core.clj: 1881  clojure.core/with-bindings*
                   RestFn.java:  425  clojure.lang.RestFn/invoke
        interruptible_eval.clj:   85  clojure.tools.nrepl.middleware.interruptible-eval/evaluate
        interruptible_eval.clj:   55  clojure.tools.nrepl.middleware.interruptible-eval/evaluate
        interruptible_eval.clj:  222  clojure.tools.nrepl.middleware.interruptible-eval/interruptible-eval/fn/fn
        interruptible_eval.clj:  190  clojure.tools.nrepl.middleware.interruptible-eval/run-next/fn
                      AFn.java:   22  clojure.lang.AFn/run
       ThreadPoolExecutor.java: 1142  java.util.concurrent.ThreadPoolExecutor/runWorker
       ThreadPoolExecutor.java:  617  java.util.concurrent.ThreadPoolExecutor$Worker/run
                   Thread.java:  745  java.lang.Thread/run

expez:
> But I would love to hear other people's opinions on this.

I'd find it annoying to have to hit <kbd>n</kbd> even more to step through things I don't care about.  I often try / want to just hold down <kbd>n</kbd> to advance quickly as it is, but that doesn't work as it will insert a literal character and then trigger the debugger source popup because the buffer was modified.  I know this is an anti-pattern but it's so much easier to just hit <kbd>n</kbd> than move the cursor to where I want it to go and hit whatever the keybinding for 'go down to here is'.  I'm probably not the only idiot doing this.

grammati:
@Malabarba, good catch, I'll add some exception handling for that.

I had also considered the possibility that we could make C-M-x detect, somehow, when it is def-ing a var and attach the source code and/or form to the var's metadata. Then the step-in logic could look for the code there first, before attempting to read it from file. That way you could be sure (I think) of getting the most up-to-date version of the function.

@expez, you're not alone, I type a lot of unwanted "n"s into my code. I sometimes wonder whether debugging should always be done in a dedicated, read-only buffer, instead of in the source buffer.

bbatsov:
Why don't we make this explicit - I'd suggest adding a command to step in the next form (if possible). This is pretty much what most debuggers do.

src/cider/nrepl/middleware/debug.clj
bbatsov: I'm fine with `util/misc.clj`.

bbatsov:
Ah, now I noticed Artur had already suggested pretty much the same thing. I'd be really uncomfortable releasing the feature in its present state as I find prompts on every step extremely annoying and I have a feeling I won't be the only one. :-) If you manage to implement something less intrusive by Saturday I'd even be willing to release this as part of 0.12.

Malabarba:
> I had also considered the possibility that we could make C-M-x detect, somehow, when it is def-ing a var and attach the source code and/or form to the var's metadata.

Yes, that's also something I've been toying with. 
That would not only fix this situation but also allow other improvements, like an AST that doesn't get built separately, (de)instrumenting code without visiting the file, find-usages functionality, etc.

However, it's also something that needs to be done carefully. In order to be able to properly attach this metadata to vars we would need to read the code ourselves, and eval it ourselves one sexp at a time (if we don't do it, it wouldn't work in messages that are evaluating more than one top-level sexp).

> you're not alone, I type a lot of unwanted "n"s into my code. I sometimes wonder whether debugging should always be done in a dedicated, read-only buffer, instead of in the source buffer.

That's only pushing the problem somewhere else. If we create a dedicated buffer for every debug session I would want to hide (or kill) it after the session. However, the same problem would arise. It's very difficult to define a debugging session, so the same situations that would lead to you typing `n` inadvertently will lead to the read-only window flickering out of existence momentarily.

Malabarba:
Back on the topic of the PR. What if we preserve the old behaviour of `n` (step-through without prompts), and implement a new key `N` that steps through **with** prompts (the feature proposed here).

bbatsov:
And what about having an explicit step in? Having different variants of step over sounds a bit strange to me.

Malabarba:
> And what about having an explicit step in?

Then we first have to agree on an interface for that. I agree with @grammati in that I don't love Edebug's interface (having to place point _before_ the sexp is unintuitive and doesn't happen in a regular step-through).

- We could implement it so that you have to hit the key somewhere inside the form, but that also wouldn't happen naturally in some situations. For instance if the function is called without arguments.
- We could implement it so that you have to hit the key _after_ evaluating the form (at the same time the debugger shows you the return value). That's possible, but it will require a bit more effort than the current approach.

Regardless of the solution, I would rather not rush it for the sake of meeting a release date. :-)

If we don't make it in time for your presentation, you could just tell people we're working on it and they should follow us on twitter to find out when it's out. :smile:

grammati:
If you want to get people excited about it, we could put "step-in" behind a flag, so that you would have to explicitly enable it. Then you could announce it at the conference and tell people how to turn it on and try it and give feedback. We would have to complicate the code a bit, temporarily, but I don't think it would be too bad.

bbatsov:
I don't quite get what's wrong with adding this as a different command e.g. `i` and just altering the current behaviour to be "i" means whatever "n" is doing now + confirmation that you want to step in. 

I'd rather avoid adding a flag that we'd just remove in the next release, but that's a viable idea as well. It just seems to me we're overthinking a bit the UI here. We have something that's OK in general, but it's simply not ok to be bound to `n`. And if it weren't bound to `n` we wouldn't need to prompt people at every step if they want to step into something or not.

grammati:
@Malabarba, I would be worried about making re-run the _only_ way to step in, because not all functions are side-effect free. I'm not sure I would want to insert a second record in my database, for example, just so I could step in to a function.

Thinking more about the flag idea: could we just have a debugger command to toggle on or off the step-in functionality globally? Then, with the toggle off, the debugger would work just as it does now. But then when you want the ability to step in to something, you just hit a key to toggle it on, you will start getting the "step in to...?" prompts.

grammati:
@bbatsov, here's a simple example:

Suppose that you step in (by whatever means) to a function `foo` that looks like this:

```clojure
(defn foo []
  (bar))
```

with the debugger, as it works now, the next breakpoint you hit will be showing you the result of `(bar)` - that is, bar will have already been called before you ever had a chance to choose to step in to it.

There are other cases similar to this, where it is not obvious where you would need the point to be when you press the "step-in" command.

bbatsov:
> with the debugger, as it works now, the next breakpoint you hit will be showing you the result of (bar) - that is, bar will have already been called before you ever had a chance to choose to step in to it.

I get this, but I don't get why this can't be a different command. You're still going to have the same problem by augmenting `n`. What I suggested is - you'd be using `i` as you'd now are using `n`, but you won't get the prompts - if something is steppable in, you just step in - if not it's business as usual. Not sure what your issues with such a behaviour are. To me it looks like the current behaviour, simply made more focused. I might be missing out on something, though.

bbatsov:
And to add a bit more detail - I'm basically proposing to go with Artur's idea about "N", but the change the behaviour of the command a bit, so it'd step in by default.

grammati:
Ummm... OK, now I'm not sure why I though this was hard, but I just tried what you suggest (different command for "next" vs. "next but step in if possible", and it seems to work great! I'll push something shortly. Thanks for the discussion, I'm excited about this, I think it's going to make everyone happy!

bbatsov:
Absolutely!

Malabarba:
Sounds good to me.

grammati:
I pushed another commit that removes the prompt, and uses an explicit `:in` command instead. I left it unsquashed for now, so you can see the diff.

I'm very happy with how it's working so far. Thanks, everyone, for convincing me that the annoying prompt was a bad idea - I think it's much better now.

One quirk that fell out of the implementation: `:in` can step you in more than one level in the call stack, if there are no intermediate breaks. eg:

```clojure
(defn bar [x] (str x))
(defn foo [] (bar "foo"))
```

if you step in to a call to `foo`, you will jump straight into `bar`, because `foo` has no break points until after bar has returned. This might be a bit surprising to users, but I'm not sure what, if anything, to do about it.

I would like to do some work on the debugger menu next, because "i" already does double-duty for "inject" and "inspect", and now we have "in". I'll try to sort something out so we can have bindings for all the commands (including "inspect", which has no binding right now).

src/cider/nrepl/middleware/debug.clj
```diff
@@ -420,22 +424,21 @@
...
+  [v coor]
+  (when (and @step-in-to-next?
+             (not (skip-breaks? coor))
+             (not= :trace (:mode @*skip-breaks*))
+             (not (::ins/breakfunction (meta v))))
+    (try
+      (instrument-var-for-step-in v)
+      true
+      (catch Exception e
+        ;; TODO - how do we inform the user that we failed?
```
bbatsov: Guess we can send some nREPL error message.

bbatsov:
The changes look good to me. 

> if you step in to a call to foo, you will jump straight into bar, because foo has no break points until after bar has returned. This might be a bit surprising to users, but I'm not sure what, if anything, to do about it.

Guess we can refine this down the road.

> I would like to do some work on the debugger menu next, because "i" already does double-duty for "inject" and "inspect", and now we have "in". I'll try to sort something out so we can have bindings for all the commands (including "inspect", which has no binding right now).

Ouch. Seems this will need work indeed.

grammati:
@Malabarba, I was thinking of changing the :input-type from a vector of keywords (`[:next :in ...]`) to a vector of pairs: `[[:next "n"] [:in "i"] [:inject "j"] [:inspect "p"] ...], so we can have keys for everything. Does that sound reasonable to you?

src/cider/nrepl/middleware/debug.clj
```diff
@@ -372,18 +482,29 @@
...
-       (do (print-step-indented ~(count coor) '~original-form val#)
-           val#)
-       ;; Nothing special. Here's the actual breakpoint logic.
-       :else (->> (pr-short val#)
-                  (assoc *extras* :debug-value)
-                  (read-debug-command val#)))))
+  (let [val-form (if (looks-step-innable? form)
+                   (let [[fn-sym & args] form]
+                     ;; Evaluate args first, delaying the decision of whether or
+                     ;; not to step in until the last possible moment
```
Malabarba: I like well-punctuated comments. :-) Add a `.` at the end please.

src/cider/nrepl/middleware/debug.clj
```diff
@@ -344,6 +349,111 @@
...
+  (when (and (seq? form) (symbol? (first form)))
+    (let [v (resolve (first form))
+          m (meta v)]
+      ;; We do not go so far as to actually try to read the code of the function
+      ;; at this point, which is at macroexpansion time.
+      (and v
+           (not (:macro m))
+           (not (:inline m))))))
+
+
```
Malabarba: extra blank line?

src/cider/nrepl/middleware/debug.clj
```diff
@@ -344,6 +349,111 @@
...
+  "Decide whether a form looks like a call to a function that we could
+  instrument and step into."
+  [form]
+  (when (and (seq? form) (symbol? (first form)))
+    (let [v (resolve (first form))
+          m (meta v)]
+      ;; We do not go so far as to actually try to read the code of the function
+      ;; at this point, which is at macroexpansion time.
+      (and v
+           (not (:macro m))
```
Malabarba: Might also want to check if it's not a special form. The full list is:

    if do recur throw finally try monitor-exit monitor-enter new quote & var clojure.core/import* . def set! loop* let* letfn* reify* deftype* fn* catch case*

Malabarba: Oh, nevermind. Special forms resolve to nil so it's already covered. Could you just add a comment saying that?

Malabarba:
> I was thinking of changing the :input-type from a vector of keywords ([:next :in ...]) to a vector of pairs: `[ ...], so we can have keys for everything. Does that sound reasonable to you?

Sounds good! But I think a map like `{"n" :next, "i" :in, "j" :inject, "p" :inspect}` would make more sense and be a little more useful.

This way you can do something like `(apply #'concat (nrepl-dict-map #'some-func input-type))` to produce the prompt, and the code that does this:

    (cdr (assq (downcase last-command-event)
               cider--debug-mode-commands-alist))

can be trivially changed to do this

    (cdr (nrepl-dict-get (downcase (string last-command-event))
                         cider--debug-mode-commands-dict))

But it's just a suggestion. Use what looks better when you're implementing it.

Malabarba:
And don't forget to move that function to utils/mics :)

grammati:
Sending the `input-type` as a map is going to break the debugger for everyone who is on the snapshot version, until they update cider. Are we OK with that?

bbatsov:
Yes.

grammati:
Other than squashing, I think this is ready.

I also pushed to https://github.com/clojure-emacs/cider/pull/1660 with the cider code to understand input-type as a map.

I was not able to move `read-var` into misc.clj, because of a circular dependency (read-var calls info/var-meta, and info.clj uses misc). So it's still in debug.clj.

bbatsov:
No need to squash the commits as they are unrelated. Thanks for the hard work!

bbatsov:
Btw, some tests seem to be failing on travis. Guess you should look into this (they fail for me locally as well).

Malabarba:
I think some of the debugger integration tests are suffering from race conditions.

grammati:
One failure is: "FileNotFoundException Could not locate clojure/tools/reader__init.class or clojure/tools/reader.clj". I suspect that's because the namespace is actually something like "mranderson.blah.blah.tools.reader" in that environment.

The intent of that test is to prove that we can instrument and step in to a function in a jar. Maybe I should use a function in clojure.core instead?

The other failure is because of paths not being as expected:
  expected: "file:/home/travis/build/clojure-emacs/cider-nrepl/src/cider/nrepl/middleware/util/misc.clj" 
  actual:  "file:/home/travis/build/clojure-emacs/cider-nrepl/target/srcdeps/cider/nrepl/middleware/util/misc.clj"

I could loosen up the assertions to just look at the start and end of the paths.

bbatsov:
> One failure is: "FileNotFoundException Could not locate clojure/tools/reader__init.class or clojure/tools/reader.clj". I suspect that's because the namespace is actually something like "mranderson.blah.blah.tools.reader" in that environment.

Ignore this one.

Malabarba:
>I could loosen up the assertions to just look at the start and end of the paths.

Yes, that's fine.

PR #340, opened 2016-04-14
Fix failing tests

grammati:
3 Fixes:

- The cond->> form in debug.clj was just wrong
- Removed tests that assumed a specific ordering of debugger commands
- Fixed instrument test for reify, which changed due to no longer
  wrapping the second symbol in `.` forms in a breakpoint

### DIFF
--- a/test/clj/cider/nrepl/middleware/debug_test.clj
+++ b/test/clj/cider/nrepl/middleware/debug_test.clj
@@ -70,16 +70,6 @@
 
 (deftest read-debug-command
   (reset! @#'d/debugger-message {})
-  ;; Check that :next is the first
-  (with-redefs [t/send (fn [trans {:keys [key input-type]}]
-                         (deliver (@d/promises key) (first input-type)))]
-    (is (= 'value (#'d/read-debug-command 'value {}))))
-
-  ;; Check that :quit is the last
-  (binding [*msg* {:session (atom {})}]
-    (with-redefs [d/abort! (constantly :aborted)
-                  d/read-debug (fn [r c & _] (last c))]
-      (is (= :aborted (#'d/read-debug-command 'value {})))))
 
   ;; Check functionality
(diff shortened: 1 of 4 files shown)


bbatsov:
Hmm, seems now we've got even more test failures.

test/clj/cider/nrepl/middleware/debug_test.clj
```diff
@@ -70,16 +70,6 @@
...
-  ;; Check that :next is the first
-  (with-redefs [t/send (fn [trans {:keys [key input-type]}]
-                         (deliver (@d/promises key) (first input-type)))]
-    (is (= 'value (#'d/read-debug-command 'value {}))))
-
-  ;; Check that :quit is the last
-  (binding [*msg* {:session (atom {})}]
-    (with-redefs [d/abort! (constantly :aborted)
-                  d/read-debug (fn [r c & _] (last c))]
-      (is (= :aborted (#'d/read-debug-command 'value {})))))
```
Malabarba: What order are we displaying these options now? I'm fine with deleting  the tests, as long as the frontend does something to maintain the old order.

Malabarba:
Thanks. Looks like you still have a couple of failures in the step-in tests:
https://travis-ci.org/clojure-emacs/cider-nrepl/jobs/123043411#L439

test/clj/cider/nrepl/middleware/debug_test.clj
grammati: Now that the commands are in a map, no specific order. Sorry, I didn't realize the order was significant. Was it intended to be roughly in order of importance / expected frequency of use?

grammati:
OK, I think I have the tests all fixed now, but there was one mysterious build failure: https://travis-ci.org/clojure-emacs/cider-nrepl/jobs/123057343

Malabarba:
That's our dreaded pigieback failure. Don't worry about it.

PR #341, opened 2016-04-14
Don't step in to clojure.core functions

grammati:
The new step-in command makes it easy to step in (often accidentally) to
a function in clojure.core. But instrumenting clojure core functions is
dangerous, and can result in a stack overflow that crashes the nrepl
process. In fact, just instrumenting `clojure.core/symbol?` (without
even debugging it) causes a stack overflow. I'm not entirely sure yet
how this happens, but it seems safest to just prevent stepping in to all
core functions.

### DIFF
--- a/src/cider/nrepl/middleware/debug.clj
+++ b/src/cider/nrepl/middleware/debug.clj
@@ -438,6 +438,12 @@
             (eval form)
             (alter-meta! v assoc ::instrumented instrumented)))))))
 
+(defn safe-to-debug?
+  "Some namespaces are not safe to debug, because doing so can cause a stack
+  overflow that crashes the nrepl process."
+  [ns]
+  (not (#{'clojure.core} (ns-name ns))))
+
 (defn step-in?
   "Return true if we can and should step in to the function in the var `v`.
   The \"should\" part is determined by the value in `step-in-to-next?`, which
@@ -464,6 +470,7 @@
       ;; We do not go so far as to actually try to read the code of the function
       ;; at this point, which is at macroexpansion time.
       (and v
+           (safe-to-debug? (:ns m))
            (not (:macro m))
            (not (:inline m))))))
 
--- a/test/clj/cider/nrepl/middleware/debug_integration_test.clj
+++ b/test/clj/cider/nrepl/middleware/debug_integration_test.clj
@@ -543,30 +543,33 @@
     (<-- {:status ["done"]})))
 
 (deftest step-in-to-function-in-jar
-  ;; Step into clojure.core/shuffle. To do this, we need to find and instrument
-  ;; the source, which is in a jar file.
-  (--> :eval "(ns user.test.step-in)")
+  ;; Step into clojure.tools.nrepl.server/handle*. To do this, we need to find
+  ;; and instrument the source, which is in a jar file. Note that this function
+  ;; is used because it is not marked as a :source-dep, so we can rely on the
+  ;; namespace remaining unmunged, which is important when these tests run on
+  ;; travis CI.
+  (--> :eval "(ns user.test.step-in
+                (:require [clojure.tools.nrepl.server :as server]))")
   (<-- {:ns "user.test.step-in"})
   (<-- {:status ["done"]})
 
   (--> :eval
        "#dbg
-        (defn foo [c]
-          (shuffle c))")
+        (defn foo [m]
+          (server/handle* m identity 23))")
   (<-- {:value "#'user.test.step-in/foo"})
   (<-- {:status ["done"]})
 
-  (--> :eval "(foo (range 2))")
-  (<-- {:debug-value "(0 1)" :coor [3 1]})
+  (--> :eval "(foo {})")
+  (<-- {:debug-value "{}" :coor [3 1]})
   (--> :in)
 
-  (let [msg  (<-- {:debug-value "(0 1)"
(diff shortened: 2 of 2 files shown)


grammati:
Oops, I just realized I changed a debug test to step in to a clojure.core function just this morning! It's going to fail... stand by.

bbatsov:
I'd suggest making this configurable - we can either use a function that would return true if some ns should be skipped or a list of namespaces. At any rate - a tiny bit of extra code will buy us a lot of flexibility. The advantage of using a function would be that you could check for namespaces starting with some common prefix (e.g. `clojure.tools`).

src/cider/nrepl/middleware/debug.clj
```diff
@@ -464,6 +464,7 @@
       ;; We do not go so far as to actually try to read the code of the function
       ;; at this point, which is at macroexpansion time.
       (and v
+           (not= 'clojure.core (ns-name (:ns m)))
```
bbatsov: Even if we hardcode this temporarily, it probably makes sense to check for all built-in namespaces (or at least all the commonly used ones).

Malabarba:
> P.S. Buy this I mean this should configurable on the Emacs side and the config should be passed when the debugger is initialized.

I think the only risk of overflow here comes from the fact that the debugger is instrumenting functions used by the debugger. So one calls the other in an endless recursion.

If that's the case, we don't need to make this customizable, we only need to exclude all namespaces that debug.clj depends on (be it directly or indeirectly).

grammati:
OK, I already pushed a commit that makes it at least possible (if not easy) to configure this, before I saw your comment @Malabarba.

I think in the short term, we should just prevent people from accidentally stepping into a clojure.core function that's going to crash their process. Then we can refine the list of excluded namespaces as we find troublesome ones.

src/cider/nrepl/middleware/debug.clj
```diff
@@ -438,12 +438,28 @@
             (eval form)
             (alter-meta! v assoc ::instrumented instrumented)))))))
 
+(def step-in-blacklist
+  "A set of names or regexes specifying namespaces that should not be stepped in
+  to. By default, this contains only \"clojure.core\"."
+  (atom #{"clojure.core"}))
```
Malabarba: Should this be a regexp?

grammati: I made it so either a string (exact namespace name) or a regexp works. But in light of your earlier comment, do you think I should just switch back to a simpler implementation that hard-codes a list of namespaces to skip (initially containing only "clojure.core")?

Malabarba: Oh sorry, I see you have a `string?` clause below.

src/cider/nrepl/middleware/debug.clj
```diff
@@ -438,12 +438,28 @@
...
+
+(defn blacklisted-for-step-in?
```
Malabarba: I don't think this is specific to `step-in` and I prefer positive predicates when possible. So rename this to `instrumentable-namespace?` and negate the return value.

Malabarba: Actually, call it `debuggable-namespace?`.

grammati: As implemented right now, this only applies to step-in. If you want to manually navigate into "clojure/core.clj" and instrument a function, we don't prevent that. Do you think we should?

src/cider/nrepl/middleware/debug.clj
Malabarba: > But in light of your earlier comment, do you think I should just switch back to a simpler implementation that hard-codes a list of namespaces to skip (initially containing only "clojure.core")?

I think so. And I think `breakpoint-if-interesting` should also avoid instrumenting if the current `*ns*` is in  this list.

src/cider/nrepl/middleware/debug.clj
Malabarba: Do we know for sure that this problem only happens with step-in?

grammati: No, the crash can definitely happen when you manually instrument. But to me, the difference is that it's very easy to step in to a core function by accident. There are probably many functions in clojure.core that are perfectly safe to debug, if someone really wants do explicitly do so, so I'm not sure we should make that impossible.

grammati:
OK, I went back to a simpler, non-configurable implementation.

It still only prevents step-in, but I'll add code to prevent manually instrumenting functions in these namespaces if you think that's the right thing to do, @Malabarba .

src/cider/nrepl/middleware/debug.clj
Malabarba: It can be pretty frustrating to get a crash in the middle of a work session. Firstly because could lose work, secondly because CIDER takes almost a minute to jack-in again.

Malabarba:
Let's keep it like this for now. I might block it for manual instrumentation soon, but there's no hury.

bbatsov:
> If that's the case, we don't need to make this customizable, we only need to exclude all namespaces that debug.clj depends on (be it directly or indeirectly).

Sure. I originally thought the purpose of this PR was to just skip stepping into boring functions, which definitely seems useful to me. 

Guess we should mention those peculiarities in the manual, as some people might wonder why they can't step into certain functions.

PR #343, opened 2016-04-15
Fix type error in merge-used-aliases

### LINKED ISSUE #338: Apparent type error in track-state/update-and-send-cache
2016-04-13-141010_563x340_scrot

stacktrace

2016-04-13-141057_772x459_scrot

This got exposed in [Jaunt](https://github.com/jaunt-lang/jaunt), where `RT.get` on non-associative types throws an exception instead of returning nil. Looks like the `new-ns-map` isn't correctly formed, wherever it's coming from.

arrdem:
Fixes #338

### DIFF
--- a/src/cider/nrepl/middleware/track_state.clj
+++ b/src/cider/nrepl/middleware/track_state.clj
@@ -96,6 +96,12 @@
                     [the-ns-name data]))))
         new-map))
 
+(defn get-aliases [o]
+  (cond
+    (instance? Namespace o) (.getAliases ^Namespace o)
+    (associative? o)        (get o :aliases)
+    :else                   nil))
+
 ;;; State management
 (defn merge-used-aliases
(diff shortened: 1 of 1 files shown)


arrdem:
As an aside, it looks like all this code is insane in the context of a Clojure repl and is targeted entirely at ClojureScript. It seems that the repl's nominal namespace in a Clojure REPL doesn't correspond to the namespace being analyzed here but I'm not entirely sure what the purpose of this code is so just expressing a concern.

Malabarba:
It's not insane. That namespace object was supposed to be a namespace map created by `ns-as-map`. There was a recent refactor in that code which reordered some things and caused the problem you discovered.

Malabarba:
I'm sorry I hadn't fixed this yet. I say your issue and fixed it locally, then forgot to push.
I've pushed the fix now, let me know if it works for your fork. :+1:

PR #349, opened 2016-05-03
Return class name as ns for Java interop methods

cskksc:

- [x] All tests are passing

Thanks!

For clojure-emacs/cider#1725

### DIFF
--- a/src/cider/nrepl/middleware/info.clj
+++ b/src/cider/nrepl/middleware/info.clj
@@ -267,7 +267,8 @@
     var-info
     {:status :no-info}))
 
-(defn extract-eldoc [info]
+(defn extract-arglists
+  [info]
   (cond
     (:special-form info) (->> (:forms info)
                               ;; :forms contains a vector of sequences or symbols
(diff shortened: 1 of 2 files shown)


src/cider/nrepl/middleware/info.clj
```diff
@@ -280,23 +280,34 @@
...
+(defn format-arglists [raw-arglists]
+  (map #(mapv str %) raw-arglists))
+
+(defn extract-ns
+  [info]
+  (cond
+    (:ns info) (:ns info)
+    (:candidates info) (->> (:candidates info)
+                            keys)
+    :else (:class info)))
```
cskksc: This method treats classes as arrays. We could also return a `:class` key. Would that be better ?

src/cider/nrepl/middleware/info.clj
```diff
@@ -280,23 +280,34 @@
...
+    (str raw-ns)))
 
 (defn eldoc
   [info]
-  (if-let [raw-eldoc (extract-eldoc info)]
-    (format-eldoc raw-eldoc)))
+  (if-let [arglists (extract-arglists info)]
+    {:eldoc (format-arglists arglists)
+     :name (:name info)
+     :ns (-> info extract-ns format-ns)}
```
cskksc: We could also return a `:class` key, to differentiate between classes and ns'. Would that be better ?

src/cider/nrepl/middleware/info.clj
```diff
@@ -280,22 +281,27 @@
...
-        var-eldoc (eldoc info)]
-    (if var-eldoc
-      {:eldoc var-eldoc
+  (let [info (info msg)]
+    (if-let [arglists (extract-arglists info)]
+      {:eldoc (format-arglists arglists)
        :name (:name info)
-       :ns (str (:ns info))}
+       :ns (str (:ns info))
+       :class (extract-class info)}
```
bbatsov: Guess it'd be best if clojure symbols didn't have the `class` key and java stuff didn't have `ns`. Btw, `class` can also be an interface, but I guess this is not very important.

test/clj/cider/nrepl/middleware/info_test.clj
```diff
@@ -424,6 +424,17 @@
       (let [response (session/message {:op "eldoc" :symbol "." :ns "user"})]
         (is (= (:status response) #{"done"}))))
 
+    (testing "java interop method with multiple classes"
+      (let [response (session/message {:op "eldoc" :symbol ".length" :ns "cider.nrepl.middleware.info-test"})]
+        (is (= (:class response)
+               "java.lang.String, java.lang.StringBuffer, java.lang.CharSequence, java.lang.StringBuilder"))
```
bbatsov: We should probably return an array.

cskksc: Yeah .. That's what I realised later. It would allow us to handle this `join` in emacs and also keep the separator configuration away from the middleware. I guess I wanted to keep the interface between emacs and middleware constant. Always a string. But it needs to be a list. Do you think we should also return an array when there is a single match ?

bbatsov: Yeah. This would be simplest.

cskksc: Done. Some unrelated tests failed, `cider.nrepl.print-method-test/print-idrefs` and `cider.nrepl.middleware.debug-integration-test`

PR #357, opened 2016-05-11
Make sure see-also.edn can be loaded

plexus:
A recent change is causing a regression, see https://github.com/clojure-emacs/cider/issues/1745

```
java.io.FileNotFoundException: resources/see-also.edn (No such file or directory),
 compiling:(info.clj:143:38)
```

I am not sure this is the best way to fix it, but it does get things working again. Opening this in the spirit of "the best way to get an answer on the internet is not to post a question, but to post a wrong answer"

----

Add `resources` to the class path, and load `see-also.edn` as a
resource, because otherwise it's causing a FileNotFoundException.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -24,6 +24,8 @@
 
   :test-paths ["test/common"] ;; See `test-clj` and `test-cljs` profiles below.
 
+  :resource-paths ["resources"]
+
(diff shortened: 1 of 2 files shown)


bbatsov:
Thanks, @plexus! @arrdem submitted a fix for the problem a little bit before you, so I've opted to merge his PR.

PR #360, opened 2016-05-17
Support variables in eldoc op

cskksc:

- [ ] All tests are passing

Thanks!

### DIFF
--- a/src/cider/nrepl/middleware/info.clj
+++ b/src/cider/nrepl/middleware/info.clj
@@ -306,14 +306,20 @@
     member {:member (str member)}
     candidates {:member (->> candidates vals (map :member) first str)}))
 
+(defn extract-eldoc
+  [info]
+  (if-let [arglists (seq (-> info extract-arglists format-arglists))]
+    {:eldoc arglists :type "function"}
+    {:type "variable"}))
+
 (defn eldoc-reply
   [msg]
-  (let [info (info msg)]
-    (if-let [arglists (extract-arglists info)]
-      (merge (extract-ns-or-class info)
-             (extract-name-or-member info)
-             {:eldoc (format-arglists arglists)})
-      {:status :no-eldoc})))
+  (if-let [info (info msg)]
+    (merge (extract-ns-or-class info)
+           (extract-name-or-member info)
+           (extract-eldoc info)
+           {:docstring (:doc info)})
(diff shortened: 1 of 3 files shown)


src/cider/nrepl/middleware/info.clj
```diff
@@ -308,12 +308,12 @@
...
-  (let [info (info msg)]
-    (if-let [arglists (extract-arglists info)]
-      (merge (extract-ns-or-class info)
-             (extract-name-or-member info)
-             {:eldoc (format-arglists arglists)})
-      {:status :no-eldoc})))
+  (if-let [info (info msg)]
+    (merge (extract-ns-or-class info)
+           (extract-name-or-member info)
+           {:doc (:doc info)}
```
bbatsov: Rename `doc` to `docstring`.

src/cider/nrepl/middleware/info.clj
```diff
@@ -308,12 +308,12 @@
...
+           {:eldoc (-> info extract-arglists format-arglists)})
```
bbatsov: Btw, we might always return to the docstring and some `:type` to the response as well, so it would be easier to check what exactly did we end up resolving in the UI. The extra dostring for functions might also be displayed somehow in the eldoc if we figure out a creative way to do so.

src/cider/nrepl/middleware/info.clj
```diff
@@ -308,12 +308,12 @@
...
+    {:status :no-eldoc}))
```
Malabarba: Combine these 3 maps into one

Malabarba: Sorry. Ignore this comment.

src/cider/nrepl/middleware/info.clj
Malabarba: Combine this map with the one below.

src/cider/nrepl/middleware/info.clj
cskksc: Yep, got it :-) So the docstring will always be returned, even for functions. We could certainly use it somewhere in the eldoc. I'll add the `:type` key here.

src/cider/nrepl/middleware/info.clj
cskksc: Done.

PR #362, opened 2016-07-02
make inspector print nil when value is nil

pdenno:
The inspector currently prints nothing when a value is nil. The 3 lines added ensures it prints “nil” (without the quotes).


Malabarba:
LGTM. Though it would be nice if someone could confirm the buggy behaviour.

pdenno:
The bug can be demonstrated by just inspecting {:foo nil}. The patch is not on the head however. I'll start over.

PR #365, opened 2016-07-16
Allow after fn's that can be called with no args

dpsutton:

- [ ] All tests are passing

Thanks!

Not sure how to run all tests actually. I was able to run the clj ones with `lein with-profile +test-clj test` but i'm not sure if there's an easy way to run them all.

I'm also not sure how to ensure that I am not generating any reflection warnings but I don't think that I've added anything that would, but I'm happy to check this before merging.

### DIFF
--- a/test/clj/cider/nrepl/middleware/refresh_test.clj
+++ b/test/clj/cider/nrepl/middleware/refresh_test.clj
@@ -16,8 +16,11 @@
 (defn- after-fn []
   (println "after-fn invoked"))
 
+(defn- after-fn-optional-arg [& a]
+  (when a (throw (IllegalArgumentException. "should not have been called with arg")))
+  (println "after with optional argument works"))
 
-(deftest refresh-op
+(deftest refresh-op-test
   (testing "refresh op works"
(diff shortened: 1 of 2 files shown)


Malabarba:
I don't think we have a test for reflection warnings, so don't worry about it. 

The cljs test was failing for you, but I think it was a false failure, so I restarted the test for you.

bbatsov:
You'll have to rebase this on top of the current `master`.

test/clj/cider/nrepl/middleware/refresh_test.clj
```diff
@@ -61,14 +65,22 @@
       (is (:error response)))))
 
 (deftest after-arg
```
bbatsov: Maybe we should use `after-arg-fn` as our term in the tests. "After arg" sounds a bit odd.

bbatsov:
@dpsutton ping :-)

dpsutton:
Thanks for the ping, sorry for the necessity :) Rebased and test names updated.

bbatsov:
👍

PR #371, opened 2016-08-20
[Fix clojure-emacs/cider#1825] Capture test input

sebastianpoeplau:
We now capture test input as generated by `test.check` and pass it on to
cider_test.el for display.


sebastianpoeplau:
It seems that the test that is failing here also fails on the latest master...

bbatsov:
LGTM. Thanks!

@Malabarba can you push to clojars? I don't have my workhorse with me. :-)

Malabarba:
I can push on Monday. I don't have my workhorse during the weekends. :-/

Malabarba:
Done :+1:

PR #373, opened 2016-08-25
fixes debugger failing for go blocks (or any macro that requires &env

### LINKED ISSUE #372: Debugger breaks on macros that require &env (like async.core/go)
```
>> (defn [foo] (go [] foo))
```

This will crash 0.14.0-SNAPSHOT

An issue is open in cider repo that provides more information. https://github.com/clojure-emacs/cider/issues/1775

The basic fix is to use https://github.com/ztellman/riddley as the macroexpander instead of walk/macroexpand-all

I have already written a patch. It seems to work on my machine. There is one problem, which is that the original form does not get saved into metadata as `::original-form`. I will submit a pull request, but perhaps this part will need to be fixed.

SlightlyCyborg:
fixes #372

### DIFF
--- a/src/cider/nrepl/middleware/util/instrument.clj
+++ b/src/cider/nrepl/middleware/util/instrument.clj
@@ -2,6 +2,7 @@
   "Generic instrumentation for clojure code"
   {:author "Artur Malabarba"}
   (:require [cider.nrepl.middleware.util.meta :as m]
+            [riddley.walk :as r]
             [clojure.walk :as walk]))
 
 ;;;; # Instrumentation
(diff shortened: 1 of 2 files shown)


SlightlyCyborg:
I forgot that the riddley library needed a patch. I locally installed the patch on my machine, but we need to wait for the riddley clojars repo to be in sync with teh github repo.

SlightlyCyborg:
I forgot

src/cider/nrepl/middleware/util/instrument.clj
```diff
@@ -285,7 +286,7 @@
                                  {::extras (assoc line-col :coor i)}))
                       form)
         ;; Expand so we don't have to deal with macros.
-        (m/macroexpand-all ::original-form)
+        (r/macroexpand-all)
```
Malabarba: Doesn't this get rid of the original-form metadata? Some functionality relies on it.

SlightlyCyborg: Yes. It does.

PR #374, opened 2016-08-26
Display Clojure.spec in the doc buffer

cskksc:

- [x] All tests are passing

Thanks!

This is not ready for a merge. I'm not really sure how do we make the nREPL use Clojure 1.9, and still maintain backwards compatibility with older versions. While developing this, I just bumped the version for the `:provided` profile, but I'm sure that's not the right way to do it. Please help me figure out the correct `project.clj` for this feature.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -53,6 +53,8 @@
 
              :1.7 {:dependencies [[org.clojure/clojure "1.7.0"]]}
              :1.8 {:dependencies [[org.clojure/clojure "1.8.0"]]}
+             :1.9 {:dependencies [[org.clojure/clojure "1.9.0-alpha12"]]
+                   :test-paths ["test/spec"]}
              :master {:repositories [["snapshots" "https://oss.sonatype.org/content/repositories/snapshots"]]
                       :dependencies [[org.clojure/clojure "1.8.0-master-SNAPSHOT"]]}
 
--- a/src/cider/nrepl/middleware/info.clj
+++ b/src/cider/nrepl/middleware/info.clj
@@ -10,7 +10,8 @@
             [cider.nrepl.middleware.util.misc :as u]
             [clojure.repl :as repl]
             [cljs-tooling.info :as cljs-info]
-            [clojure.tools.nrepl.middleware :refer [set-descriptor!]]))
+            [clojure.tools.nrepl.middleware :refer [set-descriptor!]]
+            [cider.nrepl.middleware.util.spec :as spec]))
 
 (defn- boot-project? []
   ;; fake.class.path under boot contains the original directories with source
@@ -63,9 +64,29 @@
         (assoc :file (ns/ns-path ns)))
     meta-map))
 
+;; Similar to `print-doc` from clojure.core
+;; https://github.com/clojure/clojure/blob/master/src/clj/clojure/repl.clj#L83
+(defn- format-spec
+  [fnspec]
+  (for [role [:args :ret :fn]
+        :let [spec' (get fnspec role)]
+        :when spec']
+    (str (name role) ": " (pr-str (spec/describe spec')))))
+
+(defn maybe-add-spec
+  "If the var `v` has a spec has associated with it, assoc that into meta-map.
+  The spec is formatted to avoid processing it in CIDER."
+  [v meta-map]
+  (if-let [fnspec (spec/get-spec v)]
+    (merge meta-map {:spec (format-spec fnspec)})
+    meta-map))
+
 (defn var-meta
   [v]
-  (-> v meta maybe-protocol (select-keys var-meta-whitelist) map-seq maybe-add-file))
+  (let [meta-map (-> v meta maybe-protocol
+                     (select-keys var-meta-whitelist)
+                     map-seq maybe-add-file)]
+    (maybe-add-spec v meta-map)))
 
 (defn ns-meta
   [ns]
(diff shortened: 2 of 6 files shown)


project.clj
```diff
@@ -24,7 +24,7 @@
 
   :test-paths ["test/common"] ;; See `test-clj` and `test-cljs` profiles below.
 
-  :profiles {:provided {:dependencies [[org.clojure/clojure "1.7.0"]]}
+  :profiles {:provided {:dependencies [[org.clojure/clojure "1.9.0-alpha10"]]}
```
Malabarba: I don't think we should depend on 1.9. We need a better way to use the spec functions.

src/cider/nrepl/middleware/info.clj
```diff
@@ -63,9 +75,27 @@
...
+  [fnspec]
+  (for [role [:args :ret :fn]
+        :let [spec' (get fnspec role)]
+        :when spec']
+    (str (name role) ": " (pr-str (spec/describe spec')))))
+
+(defn maybe-add-spec
+  "If the var `nm` has a spec has associated with it, assoc that into meta-map.
+  The spec is formatted to avoid processing it in CIDER."
+  [{ns :ns nm :name :as meta-map :or {nm "" ns ""}}]
```
Malabarba: `{ns :ns, nm :name, :as meta-map, :or {nm "", ns ""}}`

src/cider/nrepl/middleware/info.clj
```diff
@@ -63,9 +75,27 @@
...
+  (if-let [fnspec (spec/get-spec (resolve-var (symbol (str ns)) (symbol nm)))]
```
Malabarba: This nested calls are a little hard to read, split it into something like

    (spec/get-spec (resolve-var (symbol (str ns)) 
                                (symbol nm)))

(or use a `->`)

Malabarba: Also, what is this doing? Isn't `(resolve-var (symbol (str ns))` the same as `(find-ns ns)`?

Malabarba: Thirdly, what happens here if `resolve-var` returns nil?

test/clj/cider/nrepl/middleware/info_test.clj
```diff
@@ -416,7 +416,19 @@
...
+      (let [response (session/message {:op "info" :symbol "ranged-rand" :ns "cider.test-ns.first-test-ns"})]
+        (is (= (:status response) #{"done"}))
+        (is (= (:ns response) "cider.test-ns.first-test-ns"))
+        (is (= (:name response) "ranged-rand"))
+        (is (= (:arglists-str response) "([start end])"))
+        (is (nil? (:macro response)))
+        (is (= (:doc response) "Returns random int in range start <= rand < end."))
+        (is (= (:spec response) ["args: (and (cat :start int? :end int?) (< (:start %) (:end %)))"
+                                 "ret: int?"
+                                 "fn: (and (>= (:ret %) (-> % :args :start)) (< (:ret %) (-> % :args :end)))"])))))
```
Malabarba: also add a test for when there's no spec.

src/cider/nrepl/middleware/info.clj
cskksc: > Also, what is this doing? Isn't (resolve-var (symbol (str ns)) the same as (find-ns ns)?

We were basically reconstructing the argument `v` passed to `var-meta`, which is not included in the meta-map. It should be a var of the form, `#'some-function`. We do that with `(resolve-var ns var)` here. And `resolve-var` expects its arguments to be symbols, whereas the `ns` in this function is a `clojure.lang.Namespace` var. So we have to do`ns -> str -> symbol` conversion.

> Thirdly, what happens here if resolve-var returns nil?

`(spec/get-spec nil)` just returns nil, which is handled by the` if-let`.

cskksc: Just changed the structure a little to pass `v` into `maybe-add-spec`, to avoid the awkward conversions. Looks a little better now :-)

project.clj
cskksc: Yep. But I'm not sure how to do that. I'm gonna read up on some leiningen documentation to see if I can find something. How did we handle this before, during the `v1.8.0` transition ?

bbatsov: Well, we do run the tests on all supported clojure versions, not just the provided one. And you can always override the provided `clojure` when running `cider-nrepl` locally.

cskksc:
Update: I'm travelling for a few days. I'll update this PR in about 2 days. Sorry for the delay.

project.clj
Malabarba: I think it's not that much about leiningen. The project.clj should just keep saying `1.7.0`. The tricky part is requiring the spec namespace. We need a separate namespace in cider-nrepl that provides the spec functions we use, and then this namespace takes care to only require `clojure.spec` if it is available.

src/cider/nrepl/middleware/info.clj
Malabarba: Cool. I agree it looks a bit better.
FTR, next time you need to turn a namespace object into a symbol, I think `ns-name` should do the trick.

bbatsov:
No problem.

bbatsov:
@cskksc where are we with this?

cskksc:
@bbatsov I've managed to implement it using @Malabarba's approach. We also have to change the tests to run spec related tests only under profile +1.9. I'm working on it right now. Should be done in sometime :-)

cskksc:
I'm done with all the changes. We now have a separate `spec` directory which contains the tests  requiring Clojure 1.9. I'll try to submit a PR for `cljs-tooling` by the end of this week, if I have some free time.  I'm sorry about keeping this PR open for so long.

Also, the `debug-integration-test` tests have failed. Doesn't seem related.

bbatsov:
👍 The changes look good to me.

PR #386, opened 2016-12-18
Add path tracking functionality to inspector

alexander-yakushev:
In the inspector a path is a series of functions to be applied to the inspected data structure to arrive at the current destination. The idea is to render the path in the inspector buffer, and to give ability to copy the path to kill-ring by pressing `w`.

Path is represented as if the inspected value was threaded with `->`.

For example, in a structure:

```clojure
{:a [{:b 1} {:c 2}] :d 42}
```

after navigating to `2` the path shall be `:a (nth 1) :c`. Navigating to `:b` yields `:a (nth 0) (find :b) first`.

Haven't modified the README yet, waiting for the green light for the feature in general.

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,7 +135,7 @@ Middleware        | Op(s)      | Description
 `wrap-debug`      | `init-debugger/debug-input` | Establish a channel for `cider-debug` commands, use it to get debug input, and also wrap the eval op.
 `wrap-format`     | `format-(code/edn)` | Code and data formatting.
 `wrap-info`       | `info/eldoc` | File/line, arglists, docstrings and other metadata for vars.
-`wrap-inspect`    |`inspect-(start/refresh/pop/push/reset)` | Inspect a Clojure expression.
+`wrap-inspect`    |`inspect-(start/refresh/pop/push/reset/get-path)` | Inspect a Clojure expression.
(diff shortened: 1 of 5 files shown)


bbatsov:
Looks like an interesting feature to me! Update the README, I'll accept it.

alexander-yakushev:
Added mention in README and middleware doc.

alexander-yakushev:
Also removed `(press w to copy)` from the buffer. It might be better for discoverability, but certainly shouldn't be rendered by the backend.

alexander-yakushev:
@bbatsov: Another idea. The path can be clickable, and on click it will be copied to the kill-ring. Solves both discoverability and redundant verbosity. To make the path clickable, does something have to added on the backend, or it's solely a frontend task?

bbatsov:
> Also removed (press w to copy) from the buffer. It might be better for discoverability, but certainly shouldn't be rendered by the backend.

True. I forgot to remark on this. Obviously the clients should decide how to make the best use of this data.

> @bbatsov: Another idea. The path can be clickable, and on click it will be copied to the kill-ring. Solves both discoverability and redundant verbosity. To make the path clickable, does something have to added on the backend, or it's solely a frontend task?

Solely a front-end task.

alexander-yakushev:
Thanks! I will submit a sister pull request to CIDER soon.

bbatsov:
@alexander-yakushev Roger that! :-)

PR #387, opened 2017-01-01
Output to each session's *err* in the same manner as *out*

bhagany:
This builds on @Malabarba's work in #304 in order to fix https://github.com/clojure-emacs/cider/issues/1588 and https://github.com/clojure-emacs/cider/issues/1896.  In addition, this also redirects Java's `System/out` and `System/err` to `*out*` and `*err*` (they're confusingly not the same) so that everything you'd expect gets printed at the repl.

I didn't add tests because I'm not sure how to go about doing that without actually spinning up an nrepl server.  I'm glad to accept guidance on this, though.  Existing tests do pass.

### DIFF
--- a/src/cider/nrepl/middleware/out.clj
+++ b/src/cider/nrepl/middleware/out.clj
@@ -80,7 +80,9 @@
                       ([bytes]
                        (.write @(resolve printer) (String. bytes)))
                       ([bytes ^Integer off ^Integer len]
-                       (.write @(resolve printer) (String. bytes) off len)))
+                       (let [byte-range (byte-array
+                                         (take len (drop off bytes)))]
+                        (.write @(resolve printer) (String. byte-range)))))
                     (flush []
                       (.flush @(resolve printer))))
                   true)))
--- a/src/cider/nrepl/middleware/out.clj
+++ b/src/cider/nrepl/middleware/out.clj
@@ -14,13 +14,12 @@
             [clojure.tools.nrepl.middleware :refer [set-descriptor!]]
             [clojure.tools.nrepl.middleware.interruptible-eval :as ie]
             [clojure.tools.nrepl.middleware.session :as session])
-  (:import [java.io PrintWriter Writer]))
+  (:import [java.io PrintWriter Writer PrintStream OutputStream]))
 
 ;;; OutStream
 (defonce original-out *out*)
 (defonce original-err *err*)
 
-(declare tracked-sessions-map)
 (declare unsubscribe-session)
 
 (defmacro with-out-binding
@@ -64,6 +63,27 @@
                       (.flush printer))))
                 true))
 
+(defn print-stream
+  "Returns a PrintStream suitable for binding as java.lang.System/out
+  or java.lang.System/err. All operations are forwarded to all output
+  bindings in the sessions of messages in addition to the server's
+  usual PrintWriter (saved in `original-out` or `original-err`).
+  type is either :out or :err."
+  [type]
+  (let [printer (case type
+                  :out '*out*
+                  :err '*err*)]
+    (PrintStream. (proxy [OutputStream] []
+                    (close [] (.flush ^OutputStream this))
+                    (write
+                      ([b]
+                       (.write @(resolve printer) (String. b)))
+                      ([b ^Integer off ^Integer len]
+                       (.write @(resolve printer) (String. b) off len)))
+                    (flush []
+                      (.flush @(resolve printer))))
+                  true)))
+
 ;;; Known eval sessions
 (def tracked-sessions-map
   "Map from session ids to eval `*msg*`s.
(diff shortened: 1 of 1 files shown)


bhagany:
I don't quite understand the Travis failure, but if there's anything I should do about that, let me know.

src/cider/nrepl/middleware/out.clj
```diff
@@ -64,19 +63,40 @@
...
   (atom {}))
 
 (defn tracked-sessions-map-watch [_ _ _ new-state]
-  (let [o (forking-printer (vals new-state) :out)]
-    ;; FIXME: This won't apply to Java loggers unless we also
-    ;; `setOut`, but for that we need to convert a `PrintWriter` to a
-    ;; `PrintStream` (or maybe just not use a `PrintWriter` above).
-    ;; (System/setOut (PrintStream. o))
-    (alter-var-root #'*out* (constantly o))))
+  (let [ow (forking-printer (vals new-state) :out)
```
bbatsov: Guess we can come up with better names than `ow` and `ew`. :-)

src/cider/nrepl/middleware/out.clj
```diff
@@ -64,19 +63,40 @@
...
+  usual PrintWriter (saved in `original-out` or `original-err`).
+  type is either :out or :err."
+  [type]
+  (let [printer (case type
+                  :out '*out*
+                  :err '*err*)]
+    (PrintStream. (proxy [OutputStream] []
+                    (close [] (.flush ^OutputStream this))
+                    (write
+                      ([b]
```
bbatsov: Does `b` stand for bytes?

bbatsov:
You should also update the namespace's description to reflect the current behaviour of the code. 

I'm assuming you've tested those changes locally and they work, right?

src/cider/nrepl/middleware/out.clj
bhagany: haha, okay.  I was trying to follow the existing conventions, which seemed to favor short names in this file.  Plus, `ow` and `ew` made me chuckle.  I use more descriptive names.

src/cider/nrepl/middleware/out.clj
bhagany: Yes - I can change that too

bhagany:
Okay, will do.  And yes, it works locally :)

bhagany:
I fixed the things you requested, but I also discovered an encoding issue.  Trying to get to the bottom of that.

bhagany:
And, fixed!  I'm unaware of any remaining issues.

oliyh:
Hi @bhagany,

I think this has resulted in some undesired behaviour for us. We’re using `clojure.tools.logging` with `logback` and the `ConsoleAppender`.

Our logs, which used to go to `*nrepl-server*` only, now go to `*cider-repl*` *as well*.

Here are our logging deps:
```clojure
[org.clojure/tools.logging "0.3.1"]
[org.slf4j/jul-to-slf4j "1.7.22"]
[ch.qos.logback/logback-classic "1.1.8" :exclusions [org.slf4j/slf4j-api]]
```

logback.xml:

```
<configuration scan="true" scanPeriod="10 seconds">

  <!-- Console output -->
  <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">

    <!-- encoder defaults to ch.qos.logback.classic.encoder.PatternLayoutEncoder -->
    <encoder>
      <pattern>%date{yyyy-MM-dd HH:mm:ss.SSS} [%thread] %-5level %logger{36} - %msg severity=%level, namespace=%logger, %mdc%n</pattern>
    </encoder>
  </appender>

  <root level="INFO">
    <appender-ref ref="STDOUT" />
  </root>

</configuration>
```

And the behaviour we see is:
```clojure
user> (clojure.tools.logging/warn "hi")
2017-01-03 13:38:23.432 [nREPL-worker-16] WARN  user - hi severity=WARN, namespace=user,
nil

user> (println "hi")
hi
nil
```

We only expect `println` to go to the REPL and all log statements to go to `*nrepl-server*` only. We’re being flooded by our own log messages now in the REPL.

Can you suggest a workaround or change to accommodate our use case?

Thanks

bbatsov:
> We only expect println to go to the REPL and all log statements to go to *nrepl-server* only. We’re being flooded by our own log messages now in the REPL.

This was always a bug and I cautioned people not to rely on this. There's a ticket on CIDER's side about redirecting certain type of output to a specific buffer, but it's been on hold for a while. Guess now we'll finally have to implement it. This is the ticket - https://github.com/clojure-emacs/cider/issues/1525 It's client-side work only and should be trivial to implement. Unfortunately I'm kinda busy right now, so ideally we'd find some volunteer to help.

bhagany:
Hi @oliyh,

The fastest thing you can do is probably downgrade CIDER... in the meantime though, the things that come to mind are logging to a file instead of stdout, or configuring logging to be quieter on dev.

bbatsov:
@oliyh See also https://github.com/clojure-emacs/cider/pull/1907

There's a strong chance I'll reject this functionality in the end given some of the comments, but you can join the discussion. I think that it's best to just log to file and tail it within Emacs.

PR #388, opened 2017-01-06
Avoid printing to original `*out*` and `*err*`

bhagany:
Original `*out*` and `*err*` are still bound to the original
System/out and System/err, which causes output sent to them to end
up in the server buffer. This was undesired.

As requested here: https://github.com/clojure-emacs/cider/pull/1907#issuecomment-270955541


bbatsov:
Unfortunately we had to revert this in https://github.com/clojure-emacs/cider-nrepl/pull/474

Turns out we didn't account for people connecting to remote their running apps and messing up their logging. Down the road we should make this smarter - e.g. restoring the output streams on disconnect or just adding some commands to toggle the complete redirect on/off.

PR #395, opened 2017-02-03
Add eldoc datomic query operation

mallt:
- [x] All tests are passing

This PR adds an operation to supply eldoc information for a datomic query. The operation returns a list of the expected inputs of the supplied query.

The basic idea here is that if in the client the regular eldoc operation has determined the function call at point resolves to [datomic.api/q](http://docs.datomic.com/clojure/#datomic.api/q), the query is passed to this new operation that will return the expected inputs so these can be shown as eldoc info (as described in the cider feature request [1832](https://github.com/clojure-emacs/cider/issues/1832)).

I've been testing calls from cider already, but before finalizing the cider code I thought it would be useful to already submit this pull request for review.

Thanks!


bbatsov:
The code looks good to me.

bbatsov:
Just one small remark, I'm not sure we really need a separate eldoc op for this, but I'm fine with the current approach. (alternatively we could have just made the existing op a bit smarter)

mallt:
Thanks!

bbatsov:
@mallt Any progress on the client-side of things?

mallt:
Yes, I hope to have a PR ready within the next couple of days.

PR #396, opened 2017-02-15
Upgrade to tools.namespace:0.3.0-alpha3

### LINKED ISSUE #385: Refresh doesn't work on boot projects
## Expected behavior
operation refresh would refresh namespaces
## Actual behavior
operation refresh refreshes no namespaces

```
boot -d seancorfield/boot-new new -t app -n myapp
cd myapp
boot repl # (or boot cider repl, depending on your setup)
```
_run refresh in your preferred way_

## Environment & Version information
Boot 2.6.0

### cider-nrepl version

0.14.0

### Java version

1.8

### Operating system

Linux

---

This is caused by how c.t.n gets the classpath, solutions are to:
1. Upgrade to 0.3.0-alpha3 of c.t.n
2. https://github.com/Deraen/boot-ctn/blob/master/src/deraen/boot_ctn.clj run something like this when detecting inside a boot project

SevereOverfl0w:
Fixes #385

- [x] All tests are passing

I've also tested this locally following the reproduction steps in #385. I don't believe the remaining checkboxes apply to this PR, please correct me if I'm wrong.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -14,7 +14,7 @@
                  ^:source-dep [cljs-tooling "0.2.0"]
                  ^:source-dep [cljfmt "0.5.6" :exclusions [org.clojure/clojurescript]]
                  ^:source-dep [org.clojure/java.classpath "0.2.3"]
-                 ^:source-dep [org.clojure/tools.namespace "0.2.11"]
+                 ^:source-dep [org.clojure/tools.namespace "0.3.0-alpha3"]
                  ^:source-dep [org.clojure/tools.trace "0.7.9"]
                  ^:source-dep [org.clojure/tools.reader "0.10.0"]]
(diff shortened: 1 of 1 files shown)


bbatsov:
Generally I don't like the idea of having alpha-grade deps. Any ideas why this didn't get a stable release for ages?

Deraen:
Probably because of the open issue mentioned in the changelog: http://dev.clojure.org/jira/browse/TNS-42

There is also a other problem that causes problems with Cljc libs and Cljs: http://dev.clojure.org/jira/browse/TNS-45

bbatsov:
Well, if we merge this we'd then be replacing one set of problems for another set of problems...

bbatsov:
On the other hand - this might actually motivate someone to continue the work on `tools.namespace`. :-)

PR #400, opened 2017-03-19
Check test namespaces exist

phillord:
Addresses clojure-emacs/cider#1940

- [x ] All tests are passing
- [x ] The new code is not generating reflection warnings

Thanks!

### DIFF
--- a/src/cider/nrepl/middleware/test.clj
+++ b/src/cider/nrepl/middleware/test.clj
@@ -202,9 +202,12 @@
 (defn has-tests?
   "Return a truthy value if the namespace has any `:test` metadata."
   [ns]
-  (seq (filter (comp :test meta val)
-               (ns-interns (the-ns ns)))))
-
+  (when-let [ns-obj
+             (if (instance? clojure.lang.Namespace ns)
+               ns
+               (find-ns ns))]
+    (seq (filter (comp :test meta val)
+                 (ns-interns ns-obj)))))
 
 ;;; ## Middleware
 
--- a/test/clj/cider/nrepl/middleware/test_test.clj
+++ b/test/clj/cider/nrepl/middleware/test_test.clj
@@ -9,3 +9,8 @@
                            10))))
   (is (= (class @test/default-executor)
          java.util.concurrent.ThreadPoolExecutor)))
+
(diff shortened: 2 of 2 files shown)


bbatsov:
The fix looks good, but I'd also some unit test.

bbatsov:
It also seems that the middleware op calling this method is not wrapped in `try/catch` which should also be fixed.

bbatsov:
@phillord ping :-)

phillord:
Running a little bit behind schedule....

phillord:
I'm not sure how to handle the exceptions further up. Most of the middleware handlers seem to use `with-safe-transport` which looks like the thing, although some (`wrap-refresh`) do not. I don't really understand this code in any depth. Am I on the right track with with-safe-transport? And if so, does `wrap-refresh` and some of the others need exception handling as well?

bbatsov:
> I'm not sure how to handle the exceptions further up. Most of the middleware handlers seem to use with-safe-transport which looks like the thing, although some (wrap-refresh) do not. I don't really understand this code in any depth. Am I on the right track with with-safe-transport? And if so, does wrap-refresh and some of the others need exception handling as well?

You're on the right path - if you take a look at the code of `with-safe-transport` it simply adds some exception handling logic. If something goes wrong it would send an error reply which CIDER can present to the user. Probably it's an oversight that some middleware don't use it.  

@sanjayl added this to simplify the code repetition we had before. Details here https://github.com/clojure-emacs/cider-nrepl/commit/9a81e27dcea44457b1670f4dfd8895e1e772f658

sanjayl:
I'm not sure if `with-safe-transport` will work with this as currently written because the `msg` map is being altered [during the creation of the handler](https://github.com/clojure-emacs/cider-nrepl/blob/master/src/cider/nrepl/middleware/test.clj#L299-L309). The `with-safe-transport` macro doesn't have a mechanism to accept modifications to the `msg` at handler-creation time.

bbatsov:
OK, guess I'll merge as is for now. A fix is better than no fix. :-)

phillord:
Oh, okay. I hadn't finished, but never mind.

PR #401, opened 2017-03-22
[Fix #399] Catch ns-path exceptions propagated from tools.reader

### LINKED ISSUE #399: clojure.lang.ExceptionInfo: No reader function for tag id when requesting eldoc info from private repo
When requesting eldoc info for a function of a private repo (datomic in my case), cider-nrepl throws a `clojure.lang.ExceptionInfo: No reader function for tag id` exception.

This seems to be caused by the upgrade to `tools.namespace:0.3.0-alpha3`. Downgrading locally to `0.2.11` solves the issue.

Just mentioning it here in case someone else would encounter the same issue.

mallt:
This PR adds error handling to the `ns-path` function for catching exceptions propagated from `tools.reader` (breaking change going from 0.2.11 to 0.3.0-alpha1).

- [x] All tests are passing

### DIFF
--- a/src/cider/nrepl/middleware/util/namespace.clj
+++ b/src/cider/nrepl/middleware/util/namespace.clj
@@ -112,11 +112,13 @@
   "Return the path to a file containing namespace `ns`.
   `ns` can be a Namespace object or the name of a namespace."
   [ns]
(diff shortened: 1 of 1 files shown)


bbatsov:
👍

PR #403, opened 2017-04-10
Feature/mark project in stacktrace

dpsutton:
https://github.com/clojure-emacs/cider/issues/1929

In reference to the above. The feature is an easy way to filter down to just the project frames in a stack trace.

An idea given in that request was from a project called Prone which would read the project.clj file. This is deficient for two reasons: 1) it ties the implementation to lein when we are already a little lein heavy and 2) it doesn't work: to wit, cider-nrepl's project name is cider/cider-nrepl which would match 0 namespaces. 

In order to work around this, I look at the class path, determine which are directories, and then call tools.namespace/find-namespaces, which conveniently finds all namespaces in a seq of directories. But I ran into some non-classpath namespaces, since cider-nrepl uses profiles for testing. To accommodate this, I reduce these found namespaces down to a common prefix that I use to match against any namespaces in the stack trace.

Thus the test to see if a namespace is in a project is a) are you a namespace in a directory on the classpath or b) does your namespace begin with the common prefix of those namespaces mentioned in a).

### DIFF
--- a/src/cider/nrepl/middleware/stacktrace.clj
+++ b/src/cider/nrepl/middleware/stacktrace.clj
@@ -8,7 +8,9 @@
             [clojure.tools.nrepl.middleware :refer [set-descriptor!]]
             [clojure.tools.nrepl.middleware.session :refer [session]]
             [clojure.tools.nrepl.misc :refer [response-for]]
-            [clojure.tools.nrepl.transport :as t])
+            [clojure.tools.nrepl.transport :as t]
+            [clojure.java.io :as io]
+            [clojure.tools.namespace.find :as nsfind])
   (:import (clojure.lang Compiler$CompilerException)))
 
 ;;; ## Stacktraces
@@ -22,6 +24,11 @@
    :line   (.getLineNumber frame)
    :class  (.getClassName frame)
    :method (.getMethodName frame)})
+
+(defn flag-frame
+  "Update frame's flags vector to include the new flag."
+  [frame flag]
+  (update-in frame [:flags] (comp set conj) flag))
 
 (defn analyze-fn
   "Add namespace, fn, and var to the frame map when the source is a Clojure
@@ -49,7 +56,7 @@
                     :else (last (.split ^String file "\\."))))]
     (-> frame
         (assoc :type type)
-        (update-in [:flags] (comp set conj) type))))
+        (flag-frame type))))
 
 (defn flag-repl
   "Flag the frame if its source is a REPL eval."
@@ -57,7 +64,7 @@
   (if (and file
            (or (= file "NO_SOURCE_FILE")
                (.startsWith ^String file "form-init")))
-    (update-in frame [:flags] (comp set conj) :repl)
+    (flag-frame frame :repl)
     frame))
 
 (defn flag-tooling
@@ -68,9 +75,56 @@
   (let [tool-regex #"^clojure\.lang\.Compiler|^clojure\.tools\.nrepl|^cider\."
         tool? #(re-find tool-regex (or (:name %) ""))
         flag  #(if (tool? %)
-                 (update-in % [:flags] (comp set conj) :tooling)
+                 (flag-frame % :tooling)
                  %)]
     (map flag frames)))
+
+(def directory-namespaces
+  "This looks for all namespaces inside of directories on the class
+  path, ignoring jars."
+  (let [paths (-> (System/getProperty "java.class.path")
+                  (str/split #":"))]
+    (->> paths
+         (filter (complement #(.endsWith % ".jar")))
+         (map io/file)
+         (filter #(.isDirectory %))
+         (nsfind/find-namespaces)
+         (into #{}))))
+
+(def ns-common-prefix
+  "In order to match more namespaces, we look for a common namespace
+  prefix across the ones we have identified."
+  (let [common
+        (try (reduce
+              (fn [common ns]
+                (let [ns (str/lower-case ns)
+                      matched (map vector common ns)
+                      coincident (take-while (fn [[a b]] (= a b)) matched)]
+                  (apply str (map first coincident))))
+              (str/lower-case (first directory-namespaces))
+              directory-namespaces)
+             (catch Throwable e :error))]
+    (condp = common
+      ""
+      {:valid false :common :missing}
+
+      :error
+      {:valid false :common :error}
+
+      ;;default
+      {:valid true :common common})))
+
+(defn flag-project
+  "Flag the frame if it is from the users project. From a users
+  project means that the namespace is one we have identified or it
+  begins with the identified common prefix."
+  [{:keys [ns] :as frame}]
+  (if (and directory-namespaces ns
+           (or (contains? directory-namespaces (symbol ns))
+               (when (:valid ns-common-prefix)
+                 (.startsWith ^String ns (:common ns-common-prefix)))))
+    (flag-frame frame :project)
+    frame))
 
 (defn flag-duplicates
   "Where a parent and child frame represent substantially the same source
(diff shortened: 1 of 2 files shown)


dpsutton:
So for some reason clojure.tools.namespace isn't being found in the mr. anderson inlined deps on any java version but 1.8 it seems.

dpsutton:
@benedekfazekas is there a chance you could look at the build logs? It looks like mr anderson isn't locating some of the things correctly on jvm versions other than 1.8

benedekfazekas:
will have a look

test/clj/cider/nrepl/middleware/stacktrace_test.clj
```diff
@@ -1,7 +1,8 @@
 (ns cider.nrepl.middleware.stacktrace-test
   (:require [cider.nrepl.middleware.stacktrace :refer :all]
             [clojure.pprint :refer [pprint]]
-            [clojure.test :refer :all]))
+            [clojure.test :refer :all]
+            [clojure.tools.namespace.find :as nsfind]))
```
benedekfazekas: is this needed here? might fix the build too... ;)

dpsutton: thanks. sorry about pinging you on unnecessary stuff then :)

benedekfazekas:
haha no worries. did it fix the build then?

dpsutton:
well, that's the problem. the build worked on all of the 1.8 builds and locally on my machine ... waiting to see what the travis ci says

dpsutton:
Cleaned it right up, thanks!

benedekfazekas:
coolio :+1:

src/cider/nrepl/middleware/stacktrace.clj
```diff
@@ -8,9 +8,14 @@
...
             [clojure.tools.nrepl.middleware.session :refer [session]]
             [clojure.tools.nrepl.misc :refer [response-for]]
-            [clojure.tools.nrepl.transport :as t])
+            [clojure.tools.nrepl.transport :as t]
+            [clojure.java.io :as io]
+            [clojure.tools.namespace.find :as nsfind])
   (:import (clojure.lang Compiler$CompilerException)))
 
+(def project-name
+  (System/getProperty "user.dir"))
```
bbatsov: What exactly is the purpose of this?

src/cider/nrepl/middleware/stacktrace.clj
```diff
@@ -23,6 +28,10 @@
    :class  (.getClassName frame)
    :method (.getMethodName frame)})
 
+(defn flag-frame
```
bbatsov: A docstring here would be useful.

src/cider/nrepl/middleware/stacktrace.clj
```diff
@@ -68,10 +77,28 @@
   (let [tool-regex #"^clojure\.lang\.Compiler|^clojure\.tools\.nrepl|^cider\."
         tool? #(re-find tool-regex (or (:name %) ""))
         flag  #(if (tool? %)
-                 (update-in % [:flags] (comp set conj) :tooling)
+                 (flag-frame % :tooling)
                  %)]
     (map flag frames)))
 
+(def directory-namespaces
```
bbatsov: Some docstring here would be useful.

src/cider/nrepl/middleware/stacktrace.clj
```diff
@@ -68,10 +77,28 @@
...
+  (let [paths (-> (System/getProperty "java.class.path")
```
bbatsov: `paths` is a bit misleading here - I think we call these `cp-entries` in other places of the code. Or something like this.

src/cider/nrepl/middleware/stacktrace.clj
```diff
@@ -68,10 +77,28 @@
...
+                  (str/split #":"))]
+    (->> paths
+         (filter (complement #(.endsWith % ".jar")))
+         (map io/file)
+         (filter #(.isDirectory %))
+         (nsfind/find-namespaces)
+         (into #{}))))
+
+(defn flag-project
+  "Flag the frame if it is from the users project."
```
bbatsov: user's

bbatsov:
I guess you can squash those commits and we'd be more or less good to go.

src/cider/nrepl/middleware/stacktrace.clj
```diff
@@ -68,10 +74,50 @@
...
+  (let [common
+        (try (reduce
+              (fn [common ns]
+                (let [ns (str/lower-case ns)
+                      matched (map vector common ns)
+                      coincident (take-while (fn [[a b]] (= a b)) matched)]
+                  (apply str (map first coincident))))
+              (str/lower-case (first directory-namespaces))
+              directory-namespaces)
+             (catch Throwable e "NO_PREFIX_INFO"))]
```
bbatsov: I find it a bit odd that we're returning some random string here instead of a keyword or a symbol (which is the lisp thing to do).

src/cider/nrepl/middleware/stacktrace.clj
```diff
@@ -68,10 +74,50 @@
...
+    (if (= common "")
```
bbatsov: Same here. One more thing - I'd use `clojure.string/blank?`.

src/cider/nrepl/middleware/stacktrace.clj
```diff
@@ -68,10 +74,50 @@
...
+      common)))
+
+(defn flag-project
+  "Flag the frame if it is from the users project. From a users
+  project means that the namespace is one we have identified or it
+  begins with the identified common prefix."
+  [{:keys [ns] :as frame}]
+  (if (and directory-namespaces ns
+           (or (contains? directory-namespaces (symbol ns))
+               (.startsWith ^String ns ns-common-prefix)))
```
bbatsov: I guess `starts-with?` and `ends-with?` were added to Clojure only in 1.8, right?

src/cider/nrepl/middleware/stacktrace.clj
dpsutton: an aborted first attempt that was accidentally committed. it is removed in a later commit and will disappear in the squash.

src/cider/nrepl/middleware/stacktrace.clj
dpsutton: I'm not sure. I used this after parroting others. I assumed it was because we would invoke string methods on a type hinted object rather than let it reflect. But its quite possible I'm wrong. But looking it up, those appear to be 1.8 additions so we must leave these for now.

src/cider/nrepl/middleware/stacktrace.clj
dpsutton: that's true, but that would add logic into whether we _should_ try to match the namespace instead of letting it just fail to match. I'm happy to change it, I'm just wondering what it would look like. Perhaps let it coalesce to nil and just add a `when` clause? I just didn't want to accidentally successfully match the empty string against every namespace and mark everything as the project, which I thought this method handled explicitly

src/cider/nrepl/middleware/stacktrace.clj
dpsutton: It is already there in a later commit.

src/cider/nrepl/middleware/stacktrace.clj
dpsutton: I think I'm getting a glimmer of an idea. I'll do it after work and make the data a little more explicit to follow more idiomatic Clojure. Good catch. Always appreciate your suggestions.

dpsutton:
missed your squash comment but we are squashed and should be good to go now.

dpsutton:
@bbatsov ping :)

PR #405, opened 2017-05-30
Refactor breakpoints system

vspinu:
This addresses clojure-emacs/cider#1699. The core of the change is to remove most of the global dynamic vars and rely on lexical binding on top of the main instrumented form for all the common metadata. Currently same metadata is replicated again and again for each sub-form. Another major change is to reduce the macros machinery to two basic function calls `break` and `apply-instrumented-maybe`. 

This pretty much reaches theoretical minimum for the size of generated clojure code, one call per instrumented form. Pushing this further would require understanding lower level details of what and how eats into the java's 64KB method limit space. 

The following code

```clojure
(let [i 1]
  (+ i
     (* i i)))
```
now expands into:

```clojure
(let*
 [META_
  {:code "#dbg\n(let [i 1]\n  (+ i\n     (* i i)))\n",
   :original-id "12",
   :file "/home/vspinu/tmp/tmp.clj",
   :line 10,
   :column 0}]
 (cider.nrepl.middleware.debug/break
  []
  (let*
   [i 1]
   (cider.nrepl.middleware.debug/break
    [2]
    (+
     (cider.nrepl.middleware.debug/break [2 1] i META_ {})
     (cider.nrepl.middleware.debug/break
      [2 2]
      (*
       (cider.nrepl.middleware.debug/break [2 2 1] i META_ {})
       (cider.nrepl.middleware.debug/break [2 2 2] i META_ {}))
      META_
      {}))
    META_
    {}))
  META_
  {}))
```

instead of the current:

```clojure
(let*
 []
 (clojure.core/push-thread-bindings
  (clojure.core/hash-map
   #'cider.nrepl.middleware.debug/*skip-breaks*
   (let*
    [or__4238__auto__ cider.nrepl.middleware.debug/*skip-breaks*]
    (if or__4238__auto__ or__4238__auto__ (clojure.core/atom nil)))
   #'cider.nrepl.middleware.debug/*locals*
   {}
   #'cider.nrepl.middleware.debug/*extras*
   {:code "#dbg\n(let [i 1]\n  (+ i\n     (* i i)))\n",
    :original-id "20",
    :file "/home/vspinu/tmp/tmp.clj",
    :line 20,
    :column 0,
    :coor []}
   #'cider.nrepl.middleware.debug/*pprint-fn*
   (:pprint-fn
    clojure.tools.nrepl.middleware.interruptible-eval/*msg*)))
 (try
  (if
   (clojure.core/not
    (clojure.core/seq @cider.nrepl.middleware.debug/debugger-message))
   (do (throw (new java.lang.Exception "Debugger not initialized"))))
  (let*
   [val__6618__auto__
    (let*
     [i 1]
     (let*
      []
      (clojure.core/push-thread-bindings
       (clojure.core/hash-map
        #'cider.nrepl.middleware.debug/*skip-breaks*
        (let*
         
[or__4238__auto__ cider.nrepl.middleware.debug/*skip-breaks*]
         (if
          or__4238__auto__
          or__4238__auto__
          (clojure.core/atom nil)))
        #'cider.nrepl.middleware.debug/*locals*
        {}
        #'cider.nrepl.middleware.debug/*extras*
        {:code "#dbg\n(let [i 1]\n  (+ i\n     (* i i)))\n",
         :original-id "20",
         :file "/home/vspinu/tmp/tmp.clj",
         :line 20,
         :column 0,
         :coor [2]}
        #'cider.nrepl.middleware.debug/*pprint-fn*
        (:pprint-fn
         clojure.tools.nrepl.middleware.interruptible-eval/*msg*)))

.............................................................................
.............................................................................
clip 282 lines
.............................................................................
.............................................................................

   (if
    (cider.nrepl.middleware.debug/skip-breaks? [])
    val__6618__auto__
    (if
     (clojure.core/=
      (:mode @cider.nrepl.middleware.debug/*skip-breaks*)
      :trace)
     (do
      (cider.nrepl.middleware.debug/print-step-indented
       0
       '(let* [i 1] (+ i (* i i)))
       val__6618__auto__)
      val__6618__auto__)
     (if
      :else
      (cider.nrepl.middleware.debug/read-debug-command
       val__6618__auto__
       (clojure.core/assoc
        cider.nrepl.middleware.debug/*extras*

        :debug-value
        (cider.nrepl.middleware.debug/pr-short val__6618__auto__)))
      nil))))
  (finally (clojure.core/pop-thread-bindings))))
```


bbatsov:
@vspinu Impressive work!

@Malabarba please, take a look at the changes.

src/cider/nrepl/middleware/util/instrument.clj
```diff
@@ -253,6 +253,15 @@
   ;; Don't use `postwalk` because it destroys previous metadata.
   (walk-indexed #(tag-form %2 breakfunction) form))
 
+(defn print-form [form & [expand? meta?]]
```
bbatsov: You can add some docstring here.

src/cider/nrepl/middleware/util/instrument.clj
```diff
@@ -137,8 +135,7 @@
    (let [{extras ::extras,
           [_ orig] ::original-form,
           bf   ::breakfunction} (meta form)]
-     (when verbose-debug
-       (println "[DBG]" (not (not bf)) extras (or orig form)))
+     ;; (println "[DBG]" (not (not bf)) extras (or orig form))
```
bbatsov: I won't leave the commented out code here.

test/clj/cider/nrepl/middleware/debug_test.clj
```diff
@@ -203,27 +199,27 @@
...
-          :code        :code
-          :original-id :id))
-      (reset! d/debugger-message [:fake])
-      ;; Locals capturing
-      (is (= (:value (eval `(let [~'x 10] (d/breakpoint d/*locals* {:coor [1]} nil))))
-             '{x 10}))
-      ;; Top-level sexps are not debugged, just returned.
-      (is (= (eval `(let [~'x 10] (d/breakpoint d/*locals* {:coor []} nil)))
-             '{x 10})))))
+;; (deftest breakpoint
```
expez: Leaving this commented out was probably not intended

expez:
Thank you so much for digging into this @vspinu!  

I hit this all the time, but I haven't had to time to do anything about it.

test/clj/cider/nrepl/middleware/debug_test.clj
vspinu: Right. Good catch.

vspinu:
One thing that I am not very clear about is why `eval-with-locals` uses the dynamic `*ppprint-fn*` ([here](https://github.com/vspinu/cider-nrepl/blob/master/src/cider/nrepl/middleware/debug.clj#L251)). This doesn't seem necessary so I changed it to a more direct `(:pprint-fn *msg*)`.

vspinu:
For the record. Two more factors contribute to the size of the method - original forms that are inlined into the final code and locals (`&env` expansion). If any of these two is large you can get "method too large" even with a simple `#break`.  

The original forms could be saved in a separate atom with a simple macro trick, but locals really need to be inlined. So I am "fixing" the "locals issue" by simply recompiling without locals when "method code too large!" is caught. You won't be able to use command that depend on locals (`l,e,j,p`) but all others will work as expected.  It's a good tradeoff.

dpsutton:
do you think it would be possible for you to put together a markdown document giving a general overview of how the debugger works? I'd imagine no one knows more about it that you right now.

Malabarba:
Thanks indeed @vspinu . I'll try to have a look as soon as possible.

vspinu:
> a markdown document giving a general overview of how the debugger works?

I am not currently in the position to do that as I don't have a good understanding of the emacs-debuger communication. Some design choices are still not clear to me (for instance why compulsory initialization?, why blocking debugger?). This PR mostly concerns with the breakpoint logic. 

Only @Malabarba could make such a write-up but I am not sure it would be worth it at this stage. The debugger is still a moving target and there are already quite a few documentation notes in the code.

grammati:
For what it's worth, I have been trying out this change and it works perfectly, as far as I can tell.

I have tried out all the debug operations, and have not seen any regressions. I have also tried debugging functions that used to fail with "Method code too large", and they can be debugged now. Yay!

👍

Malabarba:
WRT a markdown guide: there's an issue on this repo (or maybe on the cider repo) about supporting cljs in the debugger. On that issue I made a *very* short guise on how the debugger works. 
I could write a more completed one if I ever find the time.

src/cider/nrepl/middleware/util/misc.clj
```diff
@@ -83,3 +83,19 @@
...
+    (let [rhex (format "%x" (bit-or 0x8 (bit-and 0x3 (rand-int 14))))]
+      (str (nhex 8) "-" (nhex 4) "-4" (nhex 3)
+           "-" rhex (nhex 3) "-" (nhex 12)))))
+
+(defn seq=
+  "To deal with, eg: (= () nil) => true"
+  [a b]
+  (= (seq a) (seq b)))
+
+
```
Malabarba: Whitespace

src/cider/nrepl/middleware/util/instrument.clj
```diff
@@ -253,50 +267,74 @@
...
+  [form]
+  (walk-indexed
+    (fn [_ f]
+      (if (instance? clojure.lang.IObj f)
+        (let [keys [::original-form ::extras ::breakfunction ::def-symbol]
+              f    (if (::def-symbol (meta f))
+                     (let [br (::breakfunction (meta f))
+                           f1 (m/strip-meta f keys)]
+                       (if br
+                         (vary-meta f1 assoc :instrumented (name (:name (meta br))))
```
Malabarba: Why use `(name (:name (meta br)))` instead of just `br`?

Malabarba: Also, we should use `::instrumented`.

src/cider/nrepl/middleware/enlighten.clj
```diff
@@ -49,10 +50,11 @@
                           (let [out# (apply ~val %&)]
                             ;; The only non-symbol form we enlighten
                             ;; is the `defn`.
-                            (d/with-debug-bindings ~extras
-                              (->> (pr-very-short out#)
-                                   (assoc d/*extras* :status :enlighten :debug-value)
-                                   d/debugger-send))
+                            (->> (assoc (:msg_ ~'META_)
```
Malabarba: Is the _ in :msg_ intentional?

Malabarba:
Besides a few comments, I think it looks good. And it must have taken a lot of work. Thanks a lot for the effort.

Malabarba:
Also what's that META_ symbol you are using? Is that some syntax I never heard of?

bbatsov:
@vspinu I'll merge this once you address @Malabarba's comments. And you might also want to combine some of the commits together, but that's not particularly important to me.

src/cider/nrepl/middleware/util/instrument.clj
vspinu: Final form has only `:instrumented` string tag in it. Keeping full function around leads to huge clutter when you print the form with meta and generally makes it much harder to understand what is what. Each meta key also adds 5 extra bytecode ops to the final object size on every breakpoint invocation. After all it's just a tag and should be a string or a symbol, especially when you send it to emacs.

The non-qualified `:instruented` is intentional. I am not bound to it but it looks to me that qualification is not warranted in this context. The`:instrumented` tag is more generic than `:cider.nrepl.middleware.util.instrument/instrumented`. Different middleware or different implementation of instrumentation should be able to use it without sending that sausage around. It also rather annoying in printouts as doesn't bear extra informational content there. On emacs  side it feels like spillage of internals as well. See [associated changes](https://github.com/clojure-emacs/cider/pull/2002).

src/cider/nrepl/middleware/enlighten.clj
vspinu: Typo indeed.

vspinu:
> Also what's that META_ symbol you are using? 

It's just anaphoric local variable. It needs to be accessed from all breakpoint macros and as clojure autoqualifies symbols in macros you need to inhibit the auto-qualification by quoting with `'~` everywhere. I appended `_` at the end to signify that it's an internal symbol which is automatically cleaned up in `sanitize-env` along all clojure temporaries with _ in them. Also to avoid potential conflicts with user variables named `META`.

BTW, that cleaning of local variables will break if users use locals with `_`, but that's not really a regression.

vspinu:
> cleaning of local variables will break if users use locals with _,

Taht could be robustified by removing symbols with `__` only. AFAICS all clojure temporaries use that. So I guess I will also use `META__` then.

bbatsov:
> Taht could be robustified by removing symbols with __ only. AFAICS all clojure temporaries use that. So I guess I will also use META__ then.

Sounds good. Also add some explanations about this in the code itself.

bbatsov:
@vspinu ping :-)

vspinu:
Yes. I will be back on it this weekend. Got a rather tight two weeks.

bbatsov:
OK. You'll also have to rebase this on top of the current `master`.

xiongtx:
I don't understand the changes, but the new expansion seems much more pleasing than the old :smile:.

vspinu:
I am done with addressing all comments and rebasing & squashing it. Added one more "side" commit relating to #410. 

One cell in travis matrix fails for non-sider reason.

bbatsov:
👍

PR #406, opened 2017-06-01
Enable stacktrace support for piggieback.

alessandrod:
As of https://github.com/cemerick/piggieback/pull/72 piggieback supports *e so this change enables the stacktrace middlware for piggieback.

The change is backwards compatible with older versions of piggieback which don't set *e. A :status :no-error response is sent as before.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -62,7 +62,7 @@
                         :java-source-paths ["test/java"]
                         :resource-paths ["test/resources"]}
              :test-cljs {:test-paths ["test/cljs"]
-                         :dependencies [[com.cemerick/piggieback "0.2.1"]
+                         :dependencies [[com.cemerick/piggieback "0.2.2"]
                                         [org.clojure/clojurescript "1.7.189"]]}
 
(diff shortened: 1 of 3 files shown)


alessandrod:
I've updated the PR to include tests. They pass locally with https://github.com/cemerick/piggieback/commit/404663b2f29dbecba04991789cd4d2fc932fbd88 but travis fails as we still depend on piggieback 0.2.1 (my change hasn't made it to a release yet).

alessandrod:
piggieback 0.2.2 was released including my change so I've bumped the dep and updated the PR.

Now all the test-cljs jobs are green except https://travis-ci.org/clojure-emacs/cider-nrepl/jobs/240737696 which fails silently. Does anyone have any clue?

bbatsov:
Hmm, perhaps it's something random. Looks pretty odd to me.

Just drop the `.` from the commit messages and I'll merge this PR.

alessandrod:
Done!

alessandrod:
Thanks a lot for merging this so quickly! I would appreciate if when you get a chance you could publish a SNAPSHOT with this so I can start depending on it in the clojure vscode extension.

PR #408, opened 2017-06-02
[Fix #407] Add support for user level messages and warnings

### LINKED ISSUE #407: Message ops?
Is there a way to trigger emacs to display a message or a warning?

vspinu:


src/cider/nrepl/middleware/util/nrep.clj
```diff
@@ -0,0 +1,26 @@
+(ns cider.nrepl.middleware.util.nrepl
+  "Common utilities for interaction with the client."
+  (:require
+   [clojure.tools.nrepl.middleware.interruptible-eval :refer [*msg*]]
+   [clojure.tools.nrepl.misc :refer [response-for]]
+   [clojure.tools.nrepl.transport :as transport]))
+
+
+(defn send-user-message
```
bbatsov: Maybe those two functions can be folded into one with one extra param (the type of the message)?

bbatsov:
How exactly is this supposed to be used, btw?

vspinu:
You are asking for exact code or for motivation?

It's similar to what you would do with `transport/send -> response-for` mechanism. 

```clj
;; use last *msg* of `interuptable-eval` for transport and response-for
(send-user-message "bla-bla")`  ;; -> emacs shows message "bla-bla", 

 ;; same but uses *msg* to build a response.
(send-user-message *msg* "bla-bla")
```

src/cider/nrepl/middleware/util/nrep.clj
vspinu: It could be better indeed. I separated them because on emacs side `display-warning` is a very different mechanism than simple `message`, but I doublt anyone would want to use that for  nrep warnings.  So yes, let me change that.

bbatsov:
> You are asking for exact code or for motivation?

I asked for the code, as I see the utility functions, but no actual usages of them, so I guess I should have probably asked about the motivation instead. :-)

vspinu:
Sometimes you just want to inform the user about various conditions or changes in behavior which are not critical to the execution. For example [this code](https://github.com/vspinu/cider-nrepl/blob/master/src/cider/nrepl/middleware/debug.clj#L459-L461) in the current debugger.  In the break branch there is another [example](https://github.com/vspinu/cider-nrepl/blob/break/src/cider/nrepl/middleware/debug.clj#L526-L530) where compilation is retried without locals.

Malabarba:
So the point is printing an Emacs message like println, but without printing to Clojure's stdout like println would?

vspinu:
Well, currently on the emacs side it does both message and println. The idea is to inform the user even when the repl is hidden.

bbatsov:
I'm fine with merging this and giving @vspinu the chance to show us the usefulness of those messages.

PR #410, opened 2017-06-16
Include special forms in apropos

### LINKED ISSUE #409: `find-symbols` does not include special forms
## Expected behavior

`cider.nrepl.middleware.apropos/find-symbols` should include special forms when no `search-ns` is specified.

Functions like `var-name` and `var-doc` should handle special forms.

## Actual behavior

`cider.nrepl.middleware.apropos/find-symbols` does not include special forms. 

This is because it uses `ns-publics` or `ns-interns`, which only return vars. Special forms needs to be retrieved with [`clojure.lang.Compiler/specials`](https://github.com/clojure/clojure/blob/clojure-1.8.0/src/jvm/clojure/lang/Compiler.java#L105), as can be seen from the implementation of [`special-symbol?`](https://github.com/clojure/clojure/blob/clojure-1.8.0/src/clj/clojure/core.clj#L4786).

`var-name` and `var-doc` only handle vars. Maybe they can leverage () [`clojure.repl/doc`](https://clojure.github.io/clojure/clojure.repl-api.html#clojure.repl/doc)?

1. Call `(cider.nrepl.middleware.apropos/find-symbols nil "if" nil false false false "")`.
2. See that the special form `if` is not among the results.

### cider-nrepl version

`v0.13.0`

### Java version

`1.8`

### Operating system

`Fedora 24`

xiongtx:
Fixes #409.

Some special forms (e.g. `if`) are included in apropos if the `search-ns` is nil or
`clojure.core`.

Included special forms are those those with documentation in `clojure.repl`,
as well as `&`, `catch`, and `finally`, the first of which uses the
documentation of `fn` and the others that of `try`.

**Update**:

With Lein 2.7.1, this passes:

```
$ lein with-profile +master,+test-clj test

Ran 181 tests containing 1118 assertions.
0 failures, 0 errors.
```

Whereas this does not:

```
$ lein with-profile +master,+plugin.mranderson/config,+test-clj test

Exception in thread "main" java.io.FileNotFoundException: Could not locate fipp/edn__init.class or fipp/edn.clj on classpath., compiling:(cider/nrepl/middleware/pprint.clj:1:1)
	at clojure.lang.Compiler.load(Compiler.java:7391)
...
```

Maybe the `thomasa/mranderson` plugin needs to be updated?

- [ ] All tests are passing

Thanks!

### DIFF
--- a/src/cider/nrepl/middleware/apropos.clj
+++ b/src/cider/nrepl/middleware/apropos.clj
@@ -4,6 +4,7 @@
   (:require [cider.nrepl.middleware.util.error-handling :refer [with-safe-transport]]
             [clojure.string :as str]
             [clojure.tools.nrepl.middleware :refer [set-descriptor!]]
+            [cider.nrepl.middleware.info :as info]
             [cider.nrepl.middleware.util.namespace :as ns]))
 
 ;;; ## Overview
@@ -17,17 +18,26 @@
 ;; abbreviated version (i.e. first sentence only) may be returned for
 ;; symbol-only searches.
 
+(def special-forms
+  "Special forms that can be apropo'ed."
+  (concat (keys (var-get #'clojure.repl/special-doc-map))
+          '[& catch finally]))
+
 (defn var-name
-  "Return a var's namespace-qualified name as a string."
+  "Return a special form's name or var's namespace-qualified name as a string."
   [v]
-  (str/join "/" ((juxt (comp ns-name :ns) :name)
-                 (meta v))))
+  (if (special-symbol? v)
+    (str (:name (info/resolve-special v)))
+    (str/join "/" ((juxt (comp ns-name :ns) :name) (meta v)))))
 
 (defn var-doc
-  "Return a var's docstring, optionally limiting the number of sentences
-  returned."
+  "Return a special form or var's docstring, optionally limiting the number of
+  sentences returned."
   ([v]
-   (or (:doc (meta v)) "(not documented)"))
+   (or (if (special-symbol? v)
+         (:doc (info/resolve-special v))
+         (:doc (meta v)))
+       "(not documented)"))
   ([n v]
    (->> (-> (var-doc v)
             (str/replace #"\s+" " ") ; normalize whitespace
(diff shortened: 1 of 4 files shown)


xiongtx:
@bbatsov Seems like some of the tests are not passing, which I suspect is due to  `thomasa/mranderson`. Who can look more into this?

bbatsov:
Probably @benedekfazekas

benedekfazekas:
will have a look

src/cider/nrepl/middleware/apropos.clj
```diff
@@ -17,17 +18,26 @@
...
+          '[& catch finally]))
+
 (defn var-name
-  "Return a var's namespace-qualified name as a string."
+  "Return a special form's name or var's namespace-qualified name as a string."
   [v]
-  (str/join "/" ((juxt (comp ns-name :ns) :name)
-                 (meta v))))
+  (if (special-symbol? v)
+    (str (:name (info/resolve-special v)))
```
vspinu: So all these functions conflate symbols and vars now? 

If that's the way, then I think instead of inserting `(special-symbol? ...)` checks everywhere it's better to have all these internals in one place - `(var-meta ..)`.  Also IMO including one middleware into another is not ideal. All these small utilities could easily live in a separate utility namespace (precisely what has been done as an aside of #405 btw).

@xiongtx do you plan for more imediate changes? Could you please give me a day or so to merge #405 as any changes to info and apropos will likely introduce hard conflicts with that branch. Thanks.

xiongtx:
`info.clj` already conflates symbols and vars. How would you separate them?

vspinu:
Not sure. How and where is it conflating them? 

I guess my point is that those small utility functions are useful outside of info.clj and better have precise semantics. So `var-xyz` better operate only on `vars` and not on anything else. But on the second though,  `(var-doc 'foo)` could be defined as `(:doc (:meta #'foo))` and not `(:doc (:meta foo))`. That would be quite consistent. Special symbols are not held in vars and cannot hold meta so there is probably no harm to treat them as pseudo vars.

benedekfazekas:
had a quick look on the build failures. They don't seem to be mranderson problems. The fact that only some java/clojure version variations fail in the matrix in itself is a good indication that it is not the inlining where something goes wrong.

hope this makes sense

PR #412, opened 2017-06-18
Change format of forms-str and arglists-str

### LINKED ISSUE #411: forms-str does not match arglists-str
## Expected behavior

The format of `:forms-str` from `cider.nrepl.middleware.info/format-response` should match that of `arglist-str`.

CIDER's `cider-docview-render-info`, for example, [treats them the same](https://github.com/clojure-emacs/cider/blob/v0.14.0/cider-doc.el#L417).

## Actual behavior

The value of [`:forms-str`](https://github.com/clojure-emacs/cider-nrepl/blob/v0.14.0/src/cider/nrepl/middleware/info.clj#L283) does not match that of [`:arglist-str`](https://github.com/clojure-emacs/cider-nrepl/blob/v0.14.0/src/cider/nrepl/middleware/info.clj#L281).

For `:forms-str`,

- There are two leading spaces
- Forms are joined with newlines

This causes forms to be rendered incorrectly in the doc buffer by `cider-docview-render-info`.

It's not clear why this is the case. Where are the leading spaces and newlines used?

Not necessary.

### cider-nrepl version

`0.14.0`

### Java version

`openjdk version "1.8.0_111"`

### Operating system

`Fedora 24`

xiongtx:
Fixes #411.

forms-str now looks like:

    "(.instanceMember instance args*)\n(.instanceMember Classname args*)"

arglists-str now looks like:

    "[x]\n[x y]\n[x y z]"

Changes include:

- No extraneous whitespace
- Newlines

Previously, the two had inconsistent formats.

arglists-str looked like:

    "([x] [x y] [x y z])"

forms-str looked like:

    "  [(.instanceMember instance args*)  (.instanceMember Classname args*)]"

- [ ] All tests are passing

Thanks!

### DIFF
--- a/src/cider/nrepl/middleware/info.clj
+++ b/src/cider/nrepl/middleware/info.clj
@@ -288,22 +288,24 @@
 
 (defn format-response
   [info]
-  (when info
-    (-> info
-        (merge (when-let [ns (:ns info)]
-                 {:ns (str ns)})
-               (when-let [args (:arglists info)]
-                 {:arglists-str (pr-str args)})
-               (when-let [forms (:forms info)]
-                 {:forms-str (->> (map #(str "  " (pr-str %)) forms)
-                                  (str/join \newline))})
-               (when-let [file (:file info)]
-                 (file-info file))
-               (when-let [path (:javadoc info)]
-                 (javadoc-info path)))
-        format-nested
-        blacklist
-        u/transform-value)))
+  (letfn [(forms-join [forms]
+            (->> (map pr-str forms)
+                 (str/join \newline)))]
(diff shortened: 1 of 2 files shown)


bbatsov:
👍

bbatsov:
> "[x]\n[x y]\n[x y z]"

It's not a big deal, but it might be even better to return just a vector of all signatures and leave it to the client to join those however they see fit.

xiongtx:
Vector as in Clojure vector? Doesn't everything that's moved b/t `cider-nrepl` and CIDER have to be a string?

PR #414, opened 2017-06-24
wait until a printer is used before requiring it.

ryfow:
fipp and puget aren't used by default but together appear to take
over half a second to load. This change just delays loading of
the printers until use.

This makes some progress toward dealing with clojure-emacs/cider#1717 and is completely contained. I've got some other changes to help with the slow nrepl middleware namespaces, but those will come in another PR.

### DIFF
--- a/src/cider/nrepl/middleware/pprint.clj
+++ b/src/cider/nrepl/middleware/pprint.clj
@@ -7,19 +7,34 @@
             [clojure.tools.nrepl.middleware :refer [set-descriptor!]]
             [clojure.tools.nrepl.middleware.session :as session]
             [clojure.tools.nrepl.misc :refer [response-for]]
-            [clojure.tools.nrepl.transport :as transport]
-            [fipp.edn :as fipp]
-            [puget.printer :as puget])
+            [clojure.tools.nrepl.transport :as transport])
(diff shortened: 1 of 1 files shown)


src/cider/nrepl/middleware/pprint.clj
```diff
@@ -7,19 +7,24 @@
...
             [clojure.tools.nrepl.misc :refer [response-for]]
-            [clojure.tools.nrepl.transport :as transport]
-            [fipp.edn :as fipp]
-            [puget.printer :as puget])
+            [clojure.tools.nrepl.transport :as transport])
   (:import clojure.tools.nrepl.transport.Transport))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+(def ^:private fipp-printer
+  (delay (do (require 'fipp.edn)
```
bbatsov: I'd mention here why we need to delay the `require`.

bbatsov: And I'd add a newline after `delay` and `do`.

ryfow:
ok, ready for another review.

PR #415, opened 2017-06-24
delay cljfmt.core require to decrease initial load time

ryfow:
This delays loading of cljfmt until it's used. It drops about 1 second of initial load time for me.

Reference: clojure-emacs/cider#1717


bbatsov:
I've got exactly the same tiny remarks as for your other PR.

Great job overall, btw! I think we should apply this for all non-essential functionality. It's amazing how nobody thought of doing this for so long.

ryfow:
@bbatsov ok, all done. By the way, I just published the tool I'm using to find the biggest issues. https://github.com/ryfow/flame-du-jour

PR #417, opened 2017-06-27
Rename `nrep.clj` to `nrepl.clj`

xiongtx:
Fixes #416.

PR #419, opened 2017-06-29
Use lein2 on travis

benedekfazekas:
to fix hack to use newer version of lein when only lein1 was available
on travis. no confusing exception anymore during environment setup.

trigged by a conversation on #417

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,28 +1,23 @@
 language: clojure
+lein: 2.7.1
 sudo: false
 cache:
   directories:
     - $HOME/.m2
-install:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,28 +1,23 @@
 language: clojure
+lein: 2.6.1
```
bbatsov: Why not 2.7.1?

And why is it `lein` here and `lein2` everywhere else?

benedekfazekas: fair points. fixed.

Issue #421, opened 2017-07-04
Info `:file` returns temporary file in boot for cljs


SevereOverfl0w:
## Expected behavior
`{:op :info :symbol "some.cljs/symbol" :session "cljs-session"}` contains a "file" like: `~/project/src/some/cljs.cljs`
## Actual behavior
`{:op :info :symbol "some.cljs/symbol" :session "cljs-session"}` contains a "file" like: `~/.boot/cache/project/-adfjdkj/k1j/src/some/cljs.cljs`
## Steps to reproduce the problem

1. `$ boot -d boot/new new -t tenzing -n your-app`
2. `$ boot cider dev` (or however you run `dev` with cider!)
3. Connect to clojurescript repl & open http://localhost:3000
4. Open src/cljs/your_app/app.cljs
5. Jump to source on `init` (I did it on the definition, this is potentially editor specific!)
6. You're in a temporary file!

### cider-nrepl version

0.15.0-SNAPSHOT

### Java version

1.8

### Operating system

Linux

bbatsov:
That's a side-effect of how boot works. We have discussed this a few times, but I don't remember what was decided. Obviously translating the filenames is not hard and can be done.

SevereOverfl0w:
This seems to be working for me locally:

```
diff --git a/src/cider/nrepl/middleware/info.clj b/src/cider/nrepl/middleware/info.clj
index 57efbb2..5ce9d6e 100644
--- a/src/cider/nrepl/middleware/info.clj
+++ b/src/cider/nrepl/middleware/info.clj
@@ -79,7 +79,23 @@
 (defn info-cljs
   [env symbol ns]
   (some-> (cljs-info/info env symbol ns)
-          (select-keys [:file :line :ns :doc :column :name :arglists])))
+          (select-keys [:file :line :ns :doc :column :name :arglists])
+          (update
+            :file
+            (fn [f]
+              (let [path (java.nio.file.Paths/get f (into-array String []))
+                    path-count (.getNameCount path)]
+                (first
+                  (sequence
+                    (comp (map #(.subpath path % path-count))
+                          (map str)
+                          (filter
+                            (fn [subpath]
+                              (let [r (clojure.java.io/resource subpath)]
+                                (and r
+                                     (= (.getAbsolutePath (clojure.java.io/file r))
+                                        f))))))
+                    (range path-count))))))))
 
 (defn info-java
   [class member]
```

[...]

PR #422, opened 2017-07-04
Find relative path for cljs & boot in info [#421]

SevereOverfl0w:
This fixes a case where in clojurescript the `:file` is stored as an
absolute path, meaning it points to a temporary file in boot.

It uses java.nio.file.Path to find all paths with the prefixes deleted,
and filters them until it finds one on the classpath. This implicitly
finds the longest one.

If it cannot find a contracted version of the file on the classpath, it
will pass through the original absolute path which points to a temporary
file.

A small optimisation here is doing this only for boot. But that does
mean that the results will be different for boot & lein as opposed to
previously where it was clj & cljs. I've not tested if there's a
performance improvement, but it will likely reduce unexpected behaviour.

| | CLJ | CLJS |
|-|-----|----------
| Boot | `:resource` |`:resource` most of the time |
| Lein | `:resource` | `:resource` unset |

Currently cljs never has `:resource`. I have no idea if this will matter
to any consumers, only future ones if at all, and it can be enabled by
disabling the check.

- [X] The commits are consistent with our [contribution guidelines](CONTRIBUTING.md). *NB: I'm skipping the testing because it would require launching a real version of boot, which is a major extension to the testing*
- [X] All tests are passing
- [X] The new code is not generating reflection warnings *NB: Doesn't seem to be*

### DIFF
--- a/src/cider/nrepl/middleware/info.clj
+++ b/src/cider/nrepl/middleware/info.clj
@@ -79,7 +79,28 @@
 (defn info-cljs
   [env symbol ns]
   (some-> (cljs-info/info env symbol ns)
-          (select-keys [:file :line :ns :doc :column :name :arglists])))
+          (select-keys [:file :line :ns :doc :column :name :arglists])
+          (update
+            :file
+            (fn [f]
+              (if (boot-project?)
+                ;; Boot stores files in a temporary directory & clojurescript
+                ;; stores the :file metadata location absolutely instead of
+                ;; relatively to the classpath. This means when doing jump to
+                ;; source in Boot & Clojurescript, you end up at the temp file.
+                ;; This code attempts to find the classpath-relative location
+                ;; of the file, so that it can be opened correctly.
+                (let [path (java.nio.file.Paths/get f (into-array String []))
+                      path-count (.getNameCount path)]
+                  (or
+                    (first
+                      (sequence
+                        (comp (map #(.subpath path % path-count))
+                              (map str)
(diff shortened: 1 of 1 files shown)


src/cider/nrepl/middleware/info.clj
```diff
@@ -79,7 +79,22 @@
 (defn info-cljs
   [env symbol ns]
   (some-> (cljs-info/info env symbol ns)
-          (select-keys [:file :line :ns :doc :column :name :arglists])))
+          (select-keys [:file :line :ns :doc :column :name :arglists])
+          (update
+            :file
+            (fn [f]
+              (if (boot-project?)
```
bbatsov: Might be a good idea to add some explanations here why are we doing this.

bbatsov: And perhaps some unit test.

SevereOverfl0w: I can add a unit test for this change, but for the most part it won't be an accurate representation of what boot does. I can emulate a _similar scenario_, but I wasn't sure if this was preferable.

SevereOverfl0w: Added the explanation. Let me know about the tests.

SevereOverfl0w:
When adding explanation, I noticed that there was an extraneous function define, so I've added an additional commit which removes it.

bbatsov:
Looks good! Thanks!

PR #423, opened 2017-07-05
Add a spec middleware 

jpmonettas:
Added a spec middleware that provides this new ops : 

*spec-list* - return a sorted list of all specs in the registry that can be filtered by prefix.
*spec-form* - return the form as in (s/form) for a spec but enhanced in the case when the resulting form contains multi-specs.
*spec-example* - generates a spec examples and returns it pprinted.

This will be used in the implementation of a spec browser to navigate the spec registry and provide data examples generated by the specs.

Some functionality that has been asked for in issues like: 

https://github.com/clojure-emacs/cider/issues/1918
https://github.com/clojure-emacs/cider/issues/1919
https://github.com/clojure-emacs/cider/issues/1901

Can also be implemented in this middleware things like:

https://github.com/clojure-emacs/cider/issues/1924

It needs at least clojure 1.9.0-alpha16 to work because of the s/form output format, anyway I don't think this is a problem since people using spec are moving fast to latest alphas.

Haven't figured out yet how to implement tests since the spec ns depends on clojure version. 
I've tested it with older clojure versions (previous to spec) and doesn't break anything, maybe someone can help me on how to implement "not supported" stuff.

It's the first time I create a pull request for Cider so probably I'm missing something.

Thanks for all the hard work on Cider btw! 

--------------------------

- [x] All tests are passing

Thanks!


expez:
> Haven't figured out yet how to implement tests since the spec ns depends on clojure version.

I think we can do that most elegantly with [test-selectors](https://github.com/technomancy/leiningen/blob/master/sample.project.clj#L345-L353).  That way we can just add version specific metadata to tests.

src/cider/nrepl/middleware/spec.clj
```diff
@@ -0,0 +1,129 @@
...
+  [form]
+  (walk/postwalk #(if (coll? %)
+                    %
+                    (str %))
+                 form))
+
+(defn spec-list
+  "Retrieves a list of all specs in the registry, sorted by ns/name.
+  If filter-prefix is not empty, keep only the specs with that prefix."
+  [filter-prefix]
```
expez: The `filter-prefix` should probably be a regexp instead, so it's maximally general.  That way it's more likely to fit the use-case the next guy comes up with without requiring a breaking change / new function.

expez:
Looks good! :+1:

bbatsov:
You'll have to rebase this on top of the current `master`.

src/cider/nrepl/middleware/spec.clj
```diff
@@ -0,0 +1,128 @@
...
+            [clojure.tools.nrepl
+             [middleware :refer [set-descriptor!]]
+             [misc :refer [response-for]]]
+            [cider.nrepl.middleware.util.cljs :as cljs]
+            [clojure.string :as str]
+            [clojure.walk :as walk]
+            [clojure.pprint :as pp]))
+
+(defn str-non-colls
+  "Given a form, convert all non collection childs to str"
```
bbatsov: All docstring should end with a `.`.

I'd also use full words like string instead of `str` - that's typically  reserved for a parameter name.

src/cider/nrepl/middleware/spec.clj
```diff
@@ -0,0 +1,128 @@
...
+                (let [checkable-part (if (.startsWith spec-symbol-str ":")
+                                       (subs spec-symbol-str 1)
+                                       spec-symbol-str)]
+                 (.startsWith checkable-part filter-prefix)))
+              sorted-specs)
+      sorted-specs)))
+
+
+(defn get-multi-spec-sub-specs
+  "Given a multi-spec form, call it's multi method methods to retrieve
```
bbatsov: it's -> its

src/cider/nrepl/middleware/spec.clj
```diff
@@ -0,0 +1,128 @@
...
+  [multi-spec-form]
+  (let [[_ mm-symbol & _] multi-spec-form]
+   (->> @(resolve mm-symbol)
+        methods 
+        (map (fn [[spec-k method]]
+               [spec-k (spec-utils/form (method nil))])))))
+
+(defn add-multi-specs
+  "Walk down a spec form and for every subform that is a multi-spec
+  add it's sub specs"
```
bbatsov: it's -> its

src/cider/nrepl/middleware/spec.clj
```diff
@@ -0,0 +1,128 @@
...
+                                       (subs spec-symbol-str 1)
+                                       spec-symbol-str)]
+                 (.startsWith checkable-part filter-prefix)))
+              sorted-specs)
+      sorted-specs)))
+
+
+(defn get-multi-spec-sub-specs
+  "Given a multi-spec form, call it's multi method methods to retrieve
+  it's subspecs"
```
bbatsov: its

src/cider/nrepl/middleware/spec.clj
```diff
@@ -0,0 +1,128 @@
...
+  [multi-spec-form]
+  (let [[_ mm-symbol & _] multi-spec-form]
```
bbatsov: mm-symbol?

src/cider/nrepl/middleware/spec.clj
```diff
@@ -0,0 +1,128 @@
...
+  (walk/postwalk (fn [sub-form]
+                   (if (and (coll? sub-form)
+                            (symbol? (first sub-form))
+                            (-> sub-form first name (= "multi-spec")))
+                     (concat sub-form (get-multi-spec-sub-specs sub-form))
+                     sub-form))
+                 form))
+
+(defn spec-from-string
+  "Given a string like `clojure.core/let` or `:user/email` returns
```
bbatsov: I'd put string examples in quotes, not backticks (those are typically reserved for references).

src/cider/nrepl/middleware/spec.clj
```diff
@@ -0,0 +1,128 @@
...
+
+(defn spec-form
+  "Given a spec symbol as a string, get the spec form and prepare it for
+  a response"
+  [spec-symb-str]
+  (-> (spec-utils/form (spec-from-string spec-symb-str))
+      add-multi-specs
+      str-non-colls))
+
+
```
bbatsov: Kill this extra blank line.

src/cider/nrepl/middleware/spec.clj
```diff
@@ -0,0 +1,128 @@
...
+(defn spec-example
+  "Given a spec symbol as a string, returns a string with a pretty printed
+  example generated by the spec"
+  [spec-symb-str]
```
bbatsov: I'd simply name this `spec-name` or something like this.

src/cider/nrepl/middleware/util/spec.clj
```diff
@@ -1,13 +1,35 @@
...
+
+(defn describe [s] (spec "describe" s))
+
+(defn registry [] (spec "registry"))
+
+(defn form [s] (spec "form" s))
+
+(defn generate [s] (spec-gen "generate" (spec "gen" s)))
+
+
```
bbatsov: Kill all trailing blank lines.

bbatsov:
> Haven't figured out yet how to implement tests since the spec ns depends on clojure version.
> I've tested it with older clojure versions (previous to spec) and doesn't break anything, maybe someone can help me on how to implement "not supported" stuff.

Can't you just do some version check?

I'd also squash those commits together.

jpmonettas:
@expez @bbatsov regarding tests I took suggestions from both and added a lein test-selectors that uses (clojure-version) so we can now add :min-clj-version to a test metadata if we need to run it only after some version.
Also followed all your added suggestions! I hope I didn't miss anything.
After we are ready I'll squash all commits into one.

src/cider/nrepl/middleware/spec.clj
```diff
@@ -21,34 +21,34 @@
 (defn spec-list
   "Retrieves a list of all specs in the registry, sorted by ns/name.
   If filter-prefix is not empty, keep only the specs with that prefix."
```
bbatsov: This should be just `filter-regexp`.

src/cider/nrepl/middleware/spec.clj
```diff
@@ -21,34 +21,34 @@
...
-  [filter-prefix]
+  [filter-prefix-regex]
```
bbatsov: Same here.

src/cider/nrepl/middleware/spec.clj
```diff
@@ -113,17 +111,16 @@
     {:doc "Return a sorted list of all specs in the registry"
      :returns {"status" "done"
                "spec-list" "The sorted list of all specs in the registry with their descriptions"}
-     :optional {"filter-prefix" "Only the specs with the filter prefix will be returned "}}
+     :optional {"filter-prefix-regex" "Only the specs that matches filter prefix regex will be returned "}}
```
bbatsov: Same here.

bbatsov:
> Also followed all your added suggestions! I hope I didn't miss anything.

I see you forgot to add final `.` to the docstrings. I've added a few more remarks, but overall things look pretty good.

Issue #418, opened 2017-06-27
`src/cider/tasks.clj` requires `boot`, but project does not depend on it


xiongtx:
## Expected behavior

[`src/cider/tasks.clj`](https://github.com/clojure-emacs/cider-nrepl/blob/6ace734fc22ba094ac3dbaf6afd3a032fa28eb24/src/cider/tasks.clj#L2) uses `boot.core`, etc., so `cider-nrepl` shoudl have a dependency on `boot`.

Otherwise, `cider-refresh` fails on `tasks.clj`.

## Actual behavior

There is [no dependency](https://github.com/clojure-emacs/cider-nrepl/blob/6ace734fc22ba094ac3dbaf6afd3a032fa28eb24/project.clj#L9) on `boot`.

## Steps to reproduce the problem

- Check out `cider-nrepl`
- `cider-jack-in`
- `cider-refresh`

### cider-nrepl version

Commit [6ace734fc22ba094ac3dbaf6afd3a032fa28eb24](https://github.com/clojure-emacs/cider-nrepl/tree/6ace734fc22ba094ac3dbaf6afd3a032fa28eb24)

### Java version

```
openjdk version "1.8.0_111"
```

### Operating system

```
Fedora 24
```

bbatsov:
Let's add @arichiardi to this conversation, as I don't know what he had in mind. In general it doesn't seem right to add a depedency to boot, just to address this.

arichiardi:
Uhm, this probably needs a dynamic loading then, I wonder why `cider-refresh` fails still. Maybe it should avoid refreshing `cider-nrepl` namespaces? They are tooling after all...

bbatsov:
> Uhm, this probably needs a dynamic loading then, I wonder why cider-refresh fails still. Maybe it should avoid refreshing cider-nrepl namespaces? They are tooling after all...

@xiongtx was working on `cider-nrepl`. :-) `cider-refresh` is pretty useful when you do this. 

Maybe this should be moved to some standalone plugin for `boot`? Not really sure what's the best course of action here...

arichiardi:
On June 27, 2017 7:29:13 AM PDT, Bozhidar Batsov <notifications@github.com> wrote:
>> Uhm, this probably needs a dynamic loading then, I wonder why
>cider-refresh fails still. Maybe it should avoid refreshing cider-nrepl
>namespaces? They are tooling after all...
>
[...]

PR #424, opened 2017-07-07
Fix [#418] Add boot as a dev dependency

SevereOverfl0w:
Both boot/base and boot/core seem necessary. For reasons I'm uncertain
of, neither depends on the other through their pom.xml. This could be a
mistake, or it could be due to how boot must isolate it's pods.

- [X] All tests are passing


bbatsov:
👍

Issue #345, opened 2016-04-21
Type error in complete


arrdem:
![2016-04-21-183638_742x834_scrot](https://cloud.githubusercontent.com/assets/704767/14727396/195fc262-07f0-11e6-9eeb-e8743a61b014.png)

Exposed in [Jaunt](https://github.com/arrdem/clojarr) which is no fun whatsoever and hates your code.

SevereOverfl0w:
I wasn't able to reproduce with a project.clj like this:

```
(defproject jaunty "0.1.0-SNAPSHOT"
  :exclusions [org.clojure/clojure]
  :dependencies [[org.jaunt-lang/jaunt "0.2.1"]])
```

and sending the raw op:

```
{:op: "complete"
 :ns "user"
 :symbol "-"}
```

---

However, I do notice that the complete middleware doesn't enforce that it comes _before_ the session middleware. It works because tools.nrepl's ordering is repeatable.

It looks like [eval enforces this by depending on `"clone"`](https://github.com/clojure/tools.nrepl/blob/c14eaa89dbb750e230530afd267cd53fdd486c51/src/main/clojure/clojure/tools/nrepl/middleware/interruptible_eval.clj#L250). We should od the same for complete and any other middleware that is also peeking into sessions.

bbatsov:
> However, I do notice that the complete middleware doesn't enforce that it comes before the session middleware. It works because tools.nrepl's ordering is repeatable.

Seems that's fixed now. I'll close this issue.

PR #427, opened 2017-07-08
Require tns/session where grab-cljs-env is used

SevereOverfl0w:
`grab-cljs-env` peeks into the message for the resolved `:session` key,
this will only have happened if the tns session middleware has been run
over the message.

TNS is consistent in how it orders middleware, so I wasn't able to
trigger this bug, but #345 seems like it could be caused by this.


bbatsov:
Take a look at the failing build.

SevereOverfl0w:
Strange. They ran locally. I suspect I may have failed to do something correctly with the source-deps…

bbatsov:
Thanks! 👍

PR #431, opened 2017-07-23
Bump version to 0.16.0-SNAPSHOT

dpsutton:


dpsutton:
this should probably happen before the CIDER one and a release made to clojars.

see https://github.com/clojure-emacs/cider/pull/2063

this happened when going to 0.15 where CIDER was updated first and then a day later cider-nrepl and there were lots of people impacted.

dpsutton:
looks like this won't be needed for a bit

PR #435, opened 2017-08-19
Fix broken instrumentation of forms with #? conditionals

vspinu:
Any forms that contain reader conditionals cannot be instrumented: 

```clj
(defn tt [a]
  (let [b (+ a 1)]
    #?(:clj (println "clj")
       :cljs (println "cljs"))))
```

The reason is that conditionals are disabled by default in clojure's readers.

### DIFF
--- a/src/cider/nrepl/middleware/debug.clj
+++ b/src/cider/nrepl/middleware/debug.clj
@@ -571,7 +571,7 @@ this map (identified by a key), and will `dissoc` it afterwards."}
                                   (interleave '[dbg break light])
                                   (apply assoc *data-readers*))]
       (try
-        (read-string code)
+        (read-string {:read-cond :allow} code)
         (catch Exception e)))
     (if @has-debug?
       ;; Technically, `instrument-and-eval` acts like a regular eval
--- a/src/cider/nrepl/middleware/debug.clj
+++ b/src/cider/nrepl/middleware/debug.clj
@@ -571,7 +571,9 @@ this map (identified by a key), and will `dissoc` it afterwards."}
                                   (interleave '[dbg break light])
                                   (apply assoc *data-readers*))]
       (try
-        (read-string {:read-cond :allow} code)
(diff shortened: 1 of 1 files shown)


bbatsov:
I guess this should also have some test and a changelog entry on CIDER's side ideally.

Malabarba:
Stepping through these forms probably won't work when they're slightly more complicated, but it's better than the current situation of not reading them at all.

vspinu:
Indeed, debugger breaks on more complex inputs. Opening a separate #437 for that.

@bbatsov there is little point of testing whether debugger enters the form or not. More specific tests of #437 should do that. I will add the log entry on the cider side for multiple fixes at once.

src/cider/nrepl/middleware/debug.clj
```diff
@@ -571,7 +571,9 @@ this map (identified by a key), and will `dissoc` it afterwards."}
                                   (interleave '[dbg break light])
                                   (apply assoc *data-readers*))]
       (try
-        (read-string {:read-cond :allow} code)
+        ;; new-line in REPL always throws; skip for debug convenience
+        (when (> (count code) 3)
```
bbatsov: Shouldn't we check for them explicitly? Seems a bit arbitrary to check that something is 3+ characters.

vspinu: I am not completely sure what you mean but we do check for all of them explicitly with that fake reader above. All of them have >= 4 chars this is why 3 characters threshold. It's indeed arbitrary but avoids entering that code in simple cases when there is obviously no reader there. It could be 6 there just as fine :).

As mentioned in the comment, new line in REPL will throw and trigger that catch over there. When debugging I am placing a print there to see exactly what is going on, and those useless throws interfere with the experience. Does this clarify your question?

bbatsov: Well, I guess you can just write more explanations here and that would be enough. I just assume that if some code is confusing to me it will probably be confusing to someone else as well.

PR #436, opened 2017-08-20
Fix *print-right-margin* binding in `wrap-pprint-fn`

vspinu:
None of the functionality currently relying on `wrap-pprint-fn` is pretty printing. The reason is that `*print-right-margin*` is always set to nil as `@session` is not binding it.

To reproduce 

```clj
(throw (ex-info "blabla" {:some.ns/bablabla.bla "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
                          :other.ns/bablabla.bla "bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb"}))
```

and you should see ex-info data printed in one long line within the stack trace.

### DIFF
--- a/src/cider/nrepl/middleware/pprint.clj
+++ b/src/cider/nrepl/middleware/pprint.clj
@@ -66,7 +66,8 @@
                                      (binding [*print-length* (or print-length (get @session #'*print-length*))
                                                *print-level* (or print-level (get @session #'*print-level*))
                                                *print-meta* (or print-meta (get @session #'*print-meta*))
-                                               *print-right-margin* (or print-right-margin (get @session #'*print-right-margin*))]
+                                               ;; pprint/*print-right-margin* is not bound by session middleware
+                                               *print-right-margin* (or print-right-margin *print-right-margin*)]
                                        ((resolve-pprint-fn pprint-fn) object)))))))
 
 (def wrap-pprint-fn-optional-arguments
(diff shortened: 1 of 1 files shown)


vspinu:
A side question. Does anyone know how to reset nrepl handler? For some wrappers like `wrap-pprint-fn` most of the logic is inside the wrapper, so I need to restart cider session after each modification to that function.

src/cider/nrepl/middleware/pprint.clj
```diff
@@ -66,7 +66,7 @@
                                      (binding [*print-length* (or print-length (get @session #'*print-length*))
                                                *print-level* (or print-level (get @session #'*print-level*))
                                                *print-meta* (or print-meta (get @session #'*print-meta*))
-                                               *print-right-margin* (or print-right-margin (get @session #'*print-right-margin*))]
+                                               *print-right-margin* (or print-right-margin *print-right-margin*)]
```
bbatsov: I'd add a remark here explaining the difference.

bbatsov:
> None of the functionality currently relying on wrap-pprint-fn is pretty printing. The reason is that *print-right-margin* is always set to nil as @session is not binding it.

Is this an nREPL bug then?

vspinu:
No, session is binding only a few essential things and current thread bindings returned by `get-thread-bindings`. `*print-right-margin*` is from `pprint` namespace. 

Alternatively, bindings could be added when nrepl session is created but I don't think this should be the case here.

bbatsov:
> No, session is binding only a few essential things and current thread bindings returned by get-thread-bindings. *print-right-margin* is from pprint namespace.

Ah, I totally forgot about this. 

> Alternatively, bindings could be added when nrepl session is created but I don't think this should be the case here.

I think your current fix is all that's needed.

PR #439, opened 2017-08-23
Fix invalid namespaced keywords

AdamFrey:
When I required the current master branch of this repo I ran into: `Invalid token: ::clojure.test/once-fixtures`. The double colon keyword prefix only works with namespace aliases that are set up in the current namespace. In this case there is not a namespace alias "clojure.test", so these keywords are invalid.

- [ ] All tests are passing


bbatsov:
Thanks!

PR #440, opened 2017-08-26
Implement dedicated clojure.spec/assert stacktrace handling

vspinu:
Accompanies clojure-emacs/cider#2082.

### DIFF
--- a/src/cider/nrepl/middleware/stacktrace.clj
+++ b/src/cider/nrepl/middleware/stacktrace.clj
@@ -10,6 +10,7 @@
             [clojure.tools.nrepl.misc :refer [response-for]]
             [clojure.tools.nrepl.transport :as t]
             [clojure.java.io :as io]
+            [clojure.set :as set]
             [clojure.tools.namespace.find :as nsfind])
   (:import (clojure.lang Compiler$CompilerException)))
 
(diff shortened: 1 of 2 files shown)


src/cider/nrepl/middleware/stacktrace.clj
```diff
@@ -187,6 +188,39 @@
...
+(def spec-abbrev (delay
+                   (if (try (require 'clojure.spec.alpha) true
+                            (catch Throwable _ false))
+                     (resolve 'clojure.spec.alpha/abbrev)
+                     #'identity)))
+
+(defn prepare-spec-data [ed pprint-fn]
+  (let [pp-str #(with-out-str (pprint-fn %))
+        problems (sort-by #(count (:path %)) (:clojure.spec.alpha/problems ed))]
+    ;; (prn {:ed ed
```
expez: Some debug statements leftover here.

expez:
I restarted the tests because the coverage was failing, but I'm not sure how they're generally green on 1.7 where spec can't even be required.  

To solve that, in a previous pr, I remember a new test-selector was added:

```clj
(deftest ^{:min-clj-version "1.9.0-alpha16"} spec-list-integration-test
  (let [filter-regex "clojure"
        filtered-spec-list (:spec-list (session/message {:op "spec-list"
                                                         :filter-regex filter-regex}))]    
    (testing "Filtered spec list retrieving nothing extra"
      (is (every? #(re-find (re-pattern (str ":?" filter-regex)) %)
                  filtered-spec-list)))
    (testing "Filtering with simple words regex"
      (is (= (count filtered-spec-list)
             (count (:spec-list (session/message {:op "spec-list"
                                                  :filter-regex (str filter-regex ".+")}))))))))
```

This looks like another great addition btw!  :+1:

vspinu:
One or two tests in a matrix usually failing for unrelated reasons. It's mostly 7th, but could be some other. I guess it's because travis chokes on resources on parallel builds or something like that. You should restart the failing test only.

> , but I'm not sure how they're generally green on 1.7 where spec can't even be required.

Spec tests are not run on 1.7. Those are in dedicated "src/spec" directory which is added to 1.9 build only. It's harder to do otherwise because one would need to require `clojure.spec.alpha` conditionally.

src/cider/nrepl/middleware/stacktrace.clj
```diff
@@ -187,6 +188,37 @@
   (when-let [data (ex-data e)]
     (into {} (filter (comp (complement ex-data-blacklist) key) data))))
 
+
+(def spec-abbrev (delay
+                   (if (try (require 'clojure.spec.alpha) true
```
bbatsov: Seems to me you should check for the non-alpha version of the ns as well, as it's eventually going to be released.

src/cider/nrepl/middleware/stacktrace.clj
```diff
@@ -187,6 +188,37 @@
...
+                            (catch Throwable _ false))
+                     (resolve 'clojure.spec.alpha/abbrev)
+                     #'identity)))
+
+(defn prepare-spec-data [ed pprint-fn]
```
bbatsov: Guess this could use a docstring.

src/cider/nrepl/middleware/stacktrace.clj
```diff
@@ -187,6 +188,37 @@
...
+  (let [pp-str #(with-out-str (pprint-fn %))
+        problems (sort-by #(count (:path %)) (:clojure.spec.alpha/problems ed))]
```
bbatsov: I'm a bit worried about the hardcoded `alpha` namespaces.

src/cider/nrepl/middleware/stacktrace.clj
```diff
@@ -187,6 +188,37 @@
...
+     :problems (for [{:keys [in val
+                             pred reason
+                             via path]
+                      :as prob} problems]
+                 (->> {:in (when-not (empty? in) (pr-str in))
+                       :val (pp-str val)
+                       :predicate (pr-str (@spec-abbrev pred))
+                       :reason reason ; <- always nil or string
+                       :spec (when-not (empty? via) (pr-str (last via)))
+                       :at  (when-not (empty? path) (pr-str path))
```
bbatsov: Seems you added an extra space here.

src/cider/nrepl/middleware/stacktrace.clj
```diff
@@ -187,6 +188,37 @@
...
+                                                  :clojure.spec.alpha/failure}
+                                                (set/difference (set (keys prob)))
+                                                (select-keys prob))]
+                                (when-not (empty? extras)
+                                  (pp-str extras)))}
+                      (filter clojure.core/val)
+                      (into {})))}))
+
 (defn analyze-cause
   "Return a map describing the exception cause. If `ex-data` exists, a `:data`
```
bbatsov: Guess you can mention this now does some spec-related stuff as well.

bbatsov:
Apart from my small remarks - everything looks good to me.

src/cider/nrepl/middleware/stacktrace.clj
vspinu: Good idea.

src/cider/nrepl/middleware/stacktrace.clj
vspinu: One way to fix this would be to create a correct namespace in try-catch from above and create an alias there. Will think of something.

PR #442, opened 2017-09-18
Fix fns defined with def not font locked as fns

oskarkv:
This fixes the problem that functions defined with `def`, e.g. `(def f (fn [] ...))` gets font locked as non-function vars, because they have no `:arglists` in their metadata. Needs https://github.com/clojure-emacs/cider/pull/2086

(I have not updated Cider's CHANGELOG.md, because I didn't know what to call the next version.)

### DIFF
--- a/src/cider/nrepl/middleware/track_state.clj
+++ b/src/cider/nrepl/middleware/track_state.clj
@@ -13,7 +13,7 @@
             [clojure.tools.nrepl.misc :refer [response-for]]
             [clojure.tools.nrepl.middleware.session :as session]
             [clojure.tools.nrepl.transport :as transport])
-  (:import clojure.lang.Namespace
+  (:import (clojure.lang Namespace MultiFn)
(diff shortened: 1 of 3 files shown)


src/cider/nrepl/middleware/track_state.clj
```diff
@@ -22,13 +22,19 @@
 
 ;;; Auxiliary
 
+(defn meta-with-fn
```
Malabarba: This is very specific to vars. I think it's best to rename the function to var-meta-with-fn.

Malabarba: And make it private

oskarkv: Ah, yes, you are right!

oskarkv:
I just realized that I broke things for ClojureScript by using "fn" instead of "arglists", will try to fix it...

oskarkv:
Fixed it for ClojureScript too.

PR #443, opened 2017-09-26
Fix crash when inspecting primitive arrays

AdamNiederer:
This type hint causes a crash on arrays of primitive types, like [I and [C,
because neither are subclasses of java.lang.Object. Resolves
https://github.com/clojure-emacs/cider/issues/2077.

- [ ] All tests are passing

I'm getting a crash when running the repl or test suite, both on master and with this change.


bbatsov:
👍

PR #444, opened 2017-10-02
Deferred middleware extra fixes

vspinu:
These are extra fixes related to #438.  

  - Better names.
  - Better docs.
  - Move pprint and version wrappers along all others for consistency.
  - Inhibit concurrent loading of deferred middleware. Should fix clojure-emacs/cider#2092.

### DIFF
--- a/src/cider/nrepl.clj
+++ b/src/cider/nrepl.clj
@@ -3,9 +3,9 @@
             [clojure.tools.nrepl.middleware.session :refer [session]]
             [clojure.tools.nrepl.middleware.pr-values :refer [pr-values]]
             [clojure.tools.nrepl.server :as nrepl-server]
+            [cider.nrepl.version :as version]
             [cider.nrepl.middleware.util.cljs :as cljs]
             [cider.nrepl.middleware.pprint :as pprint]
-            [cider.nrepl.middleware.version]
             [cider.nrepl.print-method]))
 
 (def DELAYS (atom nil))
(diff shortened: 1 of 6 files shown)


src/cider/nrepl.clj
```diff
@@ -8,52 +8,57 @@
             [cider.nrepl.middleware.pprint :as pprint]
             [cider.nrepl.print-method]))
 
-(def DELAYS (atom nil))
+(def DELAYS
```
bbatsov: Why not `deferred-middleware`? Seems like the best possible name to me.

src/cider/nrepl.clj
```diff
@@ -12,6 +12,10 @@
   "Map of `delay`s holding deferred middleware handlers."
   (atom nil))
 
+(def LOCK
```
bbatsov: I really think this should be named differently - LOCK is way too generic (and as it's not a constant I don't think it should all-caps either).

src/cider/nrepl.clj
```diff
@@ -25,8 +29,9 @@
         sym (symbol (name `~fn-name))]
     (swap! DELAYS assoc sym
            (delay
-             (require `~ns)
-             (resolve-or-fail `~fn-name)))
+             (locking LOCK
```
bbatsov: I wonder if we can't serialize this in some more elegant manner. This combination of `delay` + `locking` seems very unclojury to me.

vspinu: Not sure what is troubling you. Those are rarely used clojure functions, but they do precisely what's needed in that context. Anything else would likely imply more code. Do you have some concrete ideas?

src/cider/nrepl.clj
vspinu: Those are not middleware as such, but handlers which are deferred and that. It's also a map of `delay` objects, thus `DELAYS`.  `delayed-handlers` is probably a better name.

src/cider/nrepl.clj
vspinu: I use caps to emphasize that those a global variables.  Some form of marking is needed to differentiate them from locals. All-caps is a useful marking technique and restricting it to only constants is a too massive of a restriction IMO. 

I can make them dynamic and use `*lock*` if that's what you are more comfortable with. But we will never use the dynamic nature of the vars then. 

As to the names, how about `REQUIRE-LOCK`?

vspinu:
Ok, as per our discussion, I have improved further on docs and naming.

src/cider/nrepl.clj
```diff
@@ -3,92 +3,123 @@
...
             [clojure.tools.nrepl.middleware.pr-values :refer [pr-values]]
             [clojure.tools.nrepl.server :as nrepl-server]
+            [cider.nrepl.version :as version]
             [cider.nrepl.middleware.util.cljs :as cljs]
             [cider.nrepl.middleware.pprint :as pprint]
-            [cider.nrepl.middleware.version]
             [cider.nrepl.print-method]))
 
-(def DELAYS (atom nil))
+(def DELAYED-HANDLERS
```
xiongtx: ALL_CAPS is unidiomatic Clojure. Any reason we're doing that here?

vspinu: It's good to differentiate between local and global non-function vars. It enhances readability. Any form of marking down is fine with me, but ALL-CAPS seems nicest to me. Some use `xyz_` for this purpose.

bbatsov: Well, anything non-private is in effect global, so I'm fine with sticking with the standard notation. Anyways, that's something tiny.

PR #445, opened 2017-10-04
Handle additional overloaded version of write.

pjstadig:
The proxy for `OutputStream/write` was only handling the case that its single argument was a byte array, but the super class also allows for a single character (int) to be passed to `write`.

This was causing a problem for me when I turned on debugging output for SSL connections (using `-Djavax.net.debug=all`).


bbatsov:
👍

PR #448, opened 2017-10-29
Fix jdk source code lookup for boot projects

pdbrown:
Adds available jdk sources to the ClassLoader used to resolve boot project
resources by the 'info' op. This way, e.g. (cider-var-info  "java.lang.String"),
can return a "file" entry with an absolute url to the source, instead of just
the relative resource path.

- [x] All tests are passing

Thanks!

### DIFF
--- a/src/cider/nrepl/middleware/info.clj
+++ b/src/cider/nrepl/middleware/info.clj
@@ -23,8 +23,13 @@
                             (str "file:" path)
                             (str "file:" path dir-separator))]
                   (new java.net.URL url)))
-              paths)]
-    (new java.net.URLClassLoader (into-array java.net.URL urls))))
+              paths)
+        jdk-sources (->> [#_"see '## Classpath' notes at `cider.nrepl.middleware.util.java`"
+                          ["src.zip"]
+                          ["lib" "tools.jar"]]
+                         (map (partial apply java/jdk-resource-url))
+                         (remove nil?))]
+    (new java.net.URLClassLoader (into-array java.net.URL (concat urls jdk-sources)))))
 
 (defn- resource-full-path [relative-path]
   (if (u/boot-project?)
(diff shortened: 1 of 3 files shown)


pdbrown:
Hi, I'm pretty new here, and wanted to say thanks for publishing CIDER, it's great!
While converting a lein project to boot, I noticed couldn't jump to JDK sources, and that `(cider-var-info "java.lang.String")` was returning the relative resource path instead of the url into my JDK's `src.zip`, as it had been when I was using lein.
I tested this change locally and it seems to work, but I wasn't able to get `lein source-deps` running, so I ended up putting the `^:source-dep`s on my dependent test project's classpath manually. The issue was a
```
java.lang.ClassCastException: Cannot cast org.apache.maven.repository.internal.DefaultVersionResolver to org.sonatype.aether.impl.VersionResolver
        at java.lang.Class.cast(Class.java:3369)                                                                           
        at org.sonatype.aether.impl.internal.DefaultServiceLocator.getServices(DefaultServiceLocator.java:161)
        at org.sonatype.aether.impl.internal.DefaultServiceLocator.getService(DefaultServiceLocator.java:133)
        at org.sonatype.aether.impl.internal.DefaultRepositorySystem.initService(DefaultRepositorySystem.java:142)
        at org.sonatype.aether.impl.internal.DefaultServiceLocator.getServices(DefaultServiceLocator.java:181)           
        at org.sonatype.aether.impl.internal.DefaultServiceLocator.getService(DefaultServiceLocator.java:133)        
        at mranderson047.pomegranate.v0v3v0.cemerick.pomegranate.aether$repository_system.invokeStatic(aether.clj:110)
```
but I can file a separate issue for that.

bbatsov:
> but I can file a separate issue for that.

Thanks would be great. And I'd appreciate if you tried to look into this as well, as lately I've been super short on time for CIDER.

Thanks for your contribution!

pdbrown:
Sure thing! I took a look, but didn't get very far: https://github.com/clojure-emacs/cider-nrepl/issues/449.

PR #451, opened 2017-11-20
Find full path for stacktrace frames

SevereOverfl0w:
This makes it easier for tooling which has a native concept of error list to
integrate with the stacktrace op.

In this case, that tool is Vim.

Another change I made arbitrarily is to include cljc files as things upon which
ns & var are resolved for. I have tested this, and it works.

It would be get to some additional feedback on:
* Does this make things better for Emacs in terms of cljc?
* Is this an improvement that would be useful to emacs
* Does my additional step for invokeStatic's make sense? ( I wasn't getting files without it! )

I looked at adding line-shift to the server, but I found it to be grossly inaccurate in most cases. An explanation for it would be useful, as I suspect I was using it wrong given the massively incorrect offsets it caused for, e.g. `load$invoke`.

- [X] All tests are passing

### DIFF
--- a/src/cider/nrepl/middleware/stacktrace.clj
+++ b/src/cider/nrepl/middleware/stacktrace.clj
@@ -63,7 +63,8 @@
   "Add namespace, fn, and var to the frame map when the source is a Clojure
   function."
   [{:keys [file type class method] :as frame}]
-  (if (= :clj type)
+  (if (or (= :clj type)
+          (= :cljc type))
     (let [[ns fn & anons] (-> (repl/demunge class)
                               (str/replace #"--\d+" "")
                               (str/split #"/"))
--- a/src/cider/nrepl/middleware/stacktrace.clj
+++ b/src/cider/nrepl/middleware/stacktrace.clj
@@ -2,6 +2,7 @@
   "Cause and stacktrace analysis for exceptions"
   {:author "Jeff Valk"}
   (:require [cider.nrepl.middleware.pprint :as pprint]
+            [cider.nrepl.middleware.info :as info]
             [cider.nrepl.middleware.util.cljs :as cljs]
             [clojure.repl :as repl]
             [clojure.string :as str]
@@ -9,6 +10,8 @@
             [clojure.tools.nrepl.middleware.session :refer [session]]
             [clojure.tools.nrepl.misc :refer [response-for]]
             [clojure.tools.nrepl.transport :as t]
+            [cider.nrepl.middleware.util.misc :as u]
+            [cider.nrepl.middleware.util.java :as java]
             [clojure.java.io :as io]
             [clojure.set :as set]
             [clojure.tools.namespace.find :as nsfind])
@@ -31,6 +34,31 @@
   [frame flag]
   (update-in frame [:flags] (comp set conj) flag))
 
+(defn- source-path
+  "Return the relative source path for the class without extension."
+  [class]
+  (-> (str/replace (str class) #"\$.*" "")
+      (str/replace "." "/")))
+
+(defn- path->url
+  "Return a url for the path, either relative to classpath, or absolute."
+  [path]
+  (or (info/file-path path) (second (info/resource-path path))))
+
+(defn- frame->url
+  "Return a java.net.URL to the file referenced in the frame, if possible.
+  Useful for handling clojure vars which may not exist. Uncomprehensive list of
+  reasons for this:
+  * Failed refresh
+  * Top-level evaluation"
+  [frame]
+  (some-> (:name frame)
+          source-path
+          (str "." (last (.split ^String (:file frame)
+                                 "\\.")))
+          path->url
+          u/transform-value))
+
 (defn analyze-fn
   "Add namespace, fn, and var to the frame map when the source is a Clojure
   function."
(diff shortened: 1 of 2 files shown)


src/cider/nrepl/middleware/stacktrace.clj
```diff
@@ -31,20 +34,47 @@
   [frame flag]
   (update-in frame [:flags] (comp set conj) flag))
 
+(defn- source-path
+  "Return the relative source path for the top-level class without extension."
```
bbatsov: `top-level` class?

test/clj/cider/nrepl/middleware/stacktrace_test.clj
```diff
@@ -122,14 +132,18 @@
             :path "/foo/bar/baz.clj"
             :line 1
             :column 42}
-           (extract-location {:class "clojure.lang.Compiler$CompilerException"
-                              :message "java.lang.RuntimeException: Unable to resolve symbol: foo in this context, compiling:(/foo/bar/baz.clj:1:42)"})))
+           (select-keys
```
bbatsov: Won't it be better if you add `:full-file` to the result (even if it's stubbed)?

bbatsov:
> Does this make things better for Emacs in terms of cljc?

It probably does.

> Is this an improvement that would be useful to emacs

Potentially yes.

> Does my additional step for invokeStatic's make sense? ( I wasn't getting files without it! )

Comments inline. 

I'm not sure about the name `:full-file` though - at first I though this was some file contents. And I guess in the middleware op description there should be some explanation about the difference between the two keys.

src/cider/nrepl/middleware/stacktrace.clj
```diff
@@ -31,20 +34,47 @@
...
                               (str/split #"/"))
           fn (or fn method)] ; protocol functions are not munged
       (assoc frame
              :ns  ns
              :fn  (str/join "/" (cons fn anons))
-             :var (str ns "/" fn)))
-    frame))
+             :var (str ns "/" fn)
+             ;; Full file path
+             :full-file (or (some-> (info/info-clj 'user (symbol (str ns "/" fn)))
```
bbatsov: It seems this is so complex it deserves its own function.

src/cider/nrepl/middleware/stacktrace.clj
```diff
@@ -31,20 +34,47 @@
...
+                                    :file
+                                    info/file-info
+                                    :file
+                                    u/transform-value)
+                            ;; invoke/invokeStatic
```
bbatsov: I don't quite get why this second code branch is needed to be honest.

src/cider/nrepl/middleware/stacktrace.clj
```diff
@@ -31,20 +34,47 @@
...
+                            (when-let [[_ ext] (re-matches #".*(\.clj.?)" (:file frame))]
```
bbatsov: Do you need a regexp if know this is `clj` or `cljc`?

test/clj/cider/nrepl/middleware/stacktrace_test.clj
SevereOverfl0w: Can do. I was being lazy.

src/cider/nrepl/middleware/stacktrace.clj
SevereOverfl0w: I can improve the comment, but I noticed that invoke & invokeStatic don't get their var extracted from the frame correctly. So this uses a different approach to grab the value of the file.

src/cider/nrepl/middleware/stacktrace.clj
SevereOverfl0w: Nope, code branch was originally elsewhere.

bbatsov:
I also wonder why are builds failing. This exception seems extremely odd.

src/cider/nrepl/middleware/stacktrace.clj
bbatsov: Got it. I wonder if that's some Clojure bug, that should be reported upstream as well.

SevereOverfl0w:
I've rebased this on my fixci PR.

src/cider/nrepl/middleware/stacktrace.clj
SevereOverfl0w: Okay, I've dug in a little more, it actually serves an unexpected purpose purely during reload. As the ns & var is unloaded by refresh, getting a source location is not possible. However, you can figure out the file that should hold it, and that will be on the classpath.

invokeStatic was a distraction I accidentally bumped into, as I was doing lightweight tests with only invoke/invokeStatic in the frame. I now suspect all clojure frames are like that!

I'll update the comment accordingly.

SevereOverfl0w: Ah, it also handles a second case! Top-level evaluations have a class like `aero.core_foo$eval16976` or `aero.core-foo/eval16976` as a clojure var. This doesn't have any metadata though, as it isn't "real".

This code path is able to infer the correct filepath based on this fake var.

The reason to not always use this code path is in the case that you load in a function dynamically, although I wouldn't expect this to be commonplace.

src/cider/nrepl/middleware/stacktrace.clj
SevereOverfl0w: I've extracted out the bulk of the work here into it's own function, all that remains is a call to `info-clj` and symbolification of the ns&fn.

SevereOverfl0w:
This is ready for review again I think.

bbatsov:
`file-url` or `abs-file-path` perhaps?

SevereOverfl0w:
I think I like `file-url` better. I'll make that change now.

SevereOverfl0w:
After rebase, the tests are now failing. This seems unrelated to this PR though. The un-rebased form of the PR passed: https://travis-ci.org/SevereOverfl0w/cider-nrepl/builds/313915859

bbatsov:
Might be something to do with the other PR I just merged, seeing the that problem occurs in `classpath.clj`. That's odd, though. I recall it had green tests.

bbatsov:
Thanks for all your work on this! And sorry for the slow responses from me!

SevereOverfl0w:
The classpath.clj in the stacktrace is different to cider's. I can't seem to track down which one it means from the frames.

PR #454, opened 2017-11-25
Use ipv4 when ipv6 fails

SevereOverfl0w:
Add start-server wrapper to fix NREPL-87 in CI


bbatsov:
👍

bbatsov:
Note for the future - this is fixed in nREPL's new `master`.

PR #455, opened 2017-11-29
Use fake.class.path as classpath in Boot projects

grzm:
Boot manages the classpath dynamically through filesets, so
`java.class.path` does not include the source paths. Boot provides
access to the original directories via the `fake.class.path` system
property. Use this `fake.class.path` property instead of
`java.class.path` for Boot projects.

- [x] All tests are passing


grzm:
This resolves an issue in cider https://github.com/clojure-emacs/cider/issues/1651 where `cider-expected-ns` returns `nil` when running under boot.

bbatsov:
👍  Nicely done!

Please, follow this up with a PR updating the CIDER changelog.

grzm:
Thanks! I've submitted https://github.com/clojure-emacs/cider/pull/2124

PR #459, opened 2017-12-10
Fix boot-related classpath bugs & list resources

SevereOverfl0w:

- [ ] All tests are passing
- [X] You've updated the readme (if adding/changing middleware) **Nothing relevant to update**

---

Please note that this is a breaking change to the resources list op. It now returns a dictionary, this will allow for more extension in the future if needed.

### DIFF
--- a/src/cider/nrepl/middleware/resource.clj
+++ b/src/cider/nrepl/middleware/resource.clj
@@ -1,18 +1,48 @@
 (ns cider.nrepl.middleware.resource
-  (:require [clojure.java.io :as io]
-            [cider.nrepl.middleware.util.error-handling :refer [with-safe-transport]]
-            [compliment.sources.resources :as r]
-            [compliment.core :as jvm-complete]))
+  (:require
+    [clojure.java.io :as io]
+    [clojure.java.classpath :as classpath]
+    [cider.nrepl.middleware.util.error-handling :refer [with-safe-transport]]
+    [cider.nrepl.middleware.util.classloader :refer [class-loader]]))
+
+(defn- trim-leading-separator
+  [s]
+  (if (.startsWith s java.io.File/separator)
+    (subs s 1)
+    s))
+
+(defn- get-project-resources
+  []
+  (mapcat
+    (fn [directory]
+      (->>
+        directory
+        (file-seq)
+        (filter (memfn isFile))
+        (map (fn [file]
+               (let [relpath (-> file
+                                 (.getPath)
+                                 (.replaceFirst
+                                   (.getPath directory)
+                                   "")
+                                 (trim-leading-separator))]
+                 {:root directory
+                  :file file
+                  :relpath relpath
+                  :url (io/resource relpath)})))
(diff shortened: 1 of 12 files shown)


SevereOverfl0w:
I've found a few tests I broke with this, and I'm also trying to find the cause of misc tests breaking.

The mysterious failure is something to do with leiningen-core being required fwiw. And it has masked a test failure in the fake class path stuff. Fortunately it's just that the test is a bit wrong.

SevereOverfl0w:
This is the classpath.clj:407 that's blowing up fwiw: https://github.com/technomancy/leiningen/blob/0720e0ea8a4a1f1e4005fad377118606b7318e32/leiningen-core/src/leiningen/core/classpath.clj#L407

This is a path not loaded in by leiningen-core, even though leiningen-core is using it. So this seems to be a bug in leiningen-core 😥

SevereOverfl0w:
https://github.com/technomancy/leiningen/issues/2369 I've opened an issue, I will comment out the relevant test in order to get the tests passing again, as we're flying blind without them.

SevereOverfl0w:
Okay, the 1.9 tests are failing. I've pushed a few changes to help the situation:

* Invalid ns require in an older cljs causes test-cljs on 1.7 to fail, so 1.9 must be tested with a 1.9 version of clojurescript also
* Fixed a wrong require in an info-test for spec

However, I notice the info test for spec is broken already, I'm guessing changes in spec have caused them to break. I'll leave fixing those for someone who knows what that code is supposed to do.

src/cider/nrepl/middleware/resource.clj
```diff
@@ -1,18 +1,48 @@
...
-  (when-let [resource (io/resource name)]
+  (when-let [resource (io/resource name (class-loader))]
     (.getPath resource)))
 
-(defn resources-list [{:keys [prefix context] :as msg
-                       :or {context "(resource \"__prefix__\")" prefix ""}}]
-  (->> (jvm-complete/completions prefix {:context context
-                                         :sources [::r/resources]})
-       (map :candidate)))
+(defn resources-list [{:keys [prefix]}]
```
bbatsov: It's be nice if this got a docstring.

bbatsov:
Btw, you also have to update this https://github.com/clojure-emacs/cider-nrepl/blob/master/src/cider/nrepl.clj#L348 regarding the changing format of the return value. I do wander whether `resources-list` is a good name for something returning a dict, through. Maybe we can have 2 ops?

bbatsov:
This looks like a list of dicts now, so I guess that's fine. I was a bit confused because there were no tests checking for the nature of the updated data format. You might want to add some.

SevereOverfl0w:
Added a test for the new `:file` key.

I've also removed the prefix test, as it was passing falsely. I've removed the prefix filtering as it was unused, and doesn't make much sense to do on the server afaict, the user will prefer that filtering takes place on the frontend more likely using something like helm.

I've added a docstring to the newer form of the resources-list function.

PR #461, opened 2017-12-17
for java9's version checking

morinocat:
For java9's version checking.

There is problem at "java-api-version" var.
This is example. (tested in my pc)

example 1> (case of JAVA SE 9)
> (System/getProperty "java.version")
"9"
> java-api-version
nil
(we expect "9")

example 2> (case of JAVA SE 9.0.1)
> (System/getProperty "java.version")
"9.0.1"
> java-api-version
0
(we expect "9" also)

so, there was "java.lang.NumberFormatException : null"(java.clj:81) at example 1.
and "cider-javadoc" was not working at example 2 case cuz the url's java version is null.

I setted the default java version to 7 as you did (when exception).

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -62,7 +62,8 @@
 
              :1.7 {:dependencies [[org.clojure/clojure "1.7.0"]]}
              :1.8 {:dependencies [[org.clojure/clojure "1.8.0"]]}
-             :1.9 {:dependencies [[org.clojure/clojure "1.9.0"]]
+             :1.9 {:dependencies [[org.clojure/clojure "1.9.0"]
+                                  [org.clojure/clojurescript "1.9.946"]]
                    :test-paths ["test/spec"]}
              :master {:repositories [["snapshots" "https://oss.sonatype.org/content/repositories/snapshots"]]
                       :dependencies [[org.clojure/clojure "1.10.0-master-SNAPSHOT"]]}


bbatsov:
I'm a bit confused here - what does `java.version` return for Java 9? Is 9 or 9.x.x?

morinocat:
There is two versions of java 9.
(https://en.wikipedia.org/wiki/Java_version_history#Java_SE_9)

The one is "JAVA SE 9" released at 2017-09-21. this version returns just "9".
but "JAVA SE 9.0.1" released at 2017-10-17 returns "9.0.1".

The first example is for "JAVA SE 9" and second example is for "JAVA SE 9.0.1" :)

bbatsov:
Got it. I'll add a few comments to your fix, but it also needs some unit tests.

src/cider/nrepl/middleware/util/misc.clj
```diff
@@ -17,8 +17,15 @@
   (not (nil? (boot-fake-classpath))))
 
 (def java-api-version
-  (try (-> (System/getProperty "java.version") (str/split #"\.") second)
-       (catch Exception _ "7")))
+  (try
+    (let [java-ver (System/getProperty "java.version")
+          splited-ver (str/split java-ver #"\.")
```
bbatsov: Why don't you just using destructuring here? I'd also name `old-ver` and `new-ver` `major` and `minor`, as that's what they truly are.

src/cider/nrepl/middleware/util/misc.clj
```diff
@@ -17,8 +17,15 @@
...
+          old-ver (second splited-ver)
+          new-ver (first splited-ver)]
+      (if (< 1 (bigint new-ver))
```
bbatsov: And I'd check if major is bigger than 1 `(> major 1)` - I think this makes the code easier to understand..

src/cider/nrepl/middleware/util/misc.clj
morinocat: you are right. I did not think :), I will modify this. 
thanks for your checking.

morinocat: you are right, but versioning is changed from java9, so I use new and old :)
anyway, I will fix it as your comment. I will send new pull request. 
thanks for your checking.

PR #469, opened 2018-01-04
[Fix #468] Now special-sym-meta only takes care of special symbols...

### LINKED ISSUE #468: Fix info-clj in info middleware to correctly return information on different kinds of objects
Currently no spec is returned for clojure.core/let symbol given that there is a spec in the registry for it.

```clojure
user> (:spec (cider.nrepl.middleware.info/info-clj 'clojure.core 'let))
;; => nil
> (clojure.spec.alpha/spec 'clojure.core/let)
;; => #object[clojure.spec.alpha$fspec_impl$reify__1215 0x75789318 "clojure.spec.alpha$fspec_impl$reify__1215@75789318"]
```
Looking at cider.nrepl.middleware.info/info-clj
```clojure
(defn info-clj
  [ns sym]
  (or
     ;; it's a special (special-symbol? or :special-form)
    (m/special-sym-meta sym)
    ;; it's a var
    (m/var-meta (m/resolve-var ns sym))
    ;; sym is an alias for another ns
    (m/ns-meta (get (m/resolve-aliases ns) sym))
    ;; it's simply a full ns
    (m/ns-meta (find-ns sym))
    ;; it's a Java class/member symbol...or nil
    (java/resolve-symbol ns sym)))
``` 
we return different kind of information for different kinds of objects represented by the symbol.
For example `clojure.core/let` (a symbol that represents a var) is being treated as a "special form" instead of a macro, so no spec info.

What I find very confusing in Clojure documentation is the use of "special form" which is not a concept, just an annotation on vars (always macros) that are special forms in other lisps.

In Clojure, real "special forms" are called special symbols:

```clojure
user> (keys (. clojure.lang.Compiler specials))
(&
 let*
 monitor-exit
 case*
 fn*
 try
 reify*
...
)
``` 
and even if the oficial documentation says let is a special form, it isn't special at all, let* is, while let is just a macro.

Looking at Clojure code I couldn't find any use of :special-form annotation apart from printing a label at the repl doc.

I think the cleanest way to branch info-clj is to forget about the "special form" thing and say a symbol could be one of :
- special symbol 
- a var
- a namespace
- a java thing

The only things that are annotated under special-forms in Clojure are : 
clojure.core/let
clojure.core/fn
clojure.core/letfn
clojure.core/loop

jpmonettas:
Fix for https://github.com/clojure-emacs/cider-nrepl/issues/468

Now special-sym-meta only takes care of special symbols. 
Adds `maybe-add-url`.
Adds and fixes some clojure 1.9 spec tests.

- [X] All tests are passing

Thanks!

### DIFF
--- a/src/cider/nrepl/middleware/util/meta.clj
+++ b/src/cider/nrepl/middleware/util/meta.clj
@@ -42,6 +42,19 @@
     (merge meta-map {:spec (format-spec fnspec)})
     meta-map))
 
+(defn- maybe-add-url
+  "If `info` has a non blank :url or it's a :special-form build a :url
+  entry pointing to https://clojure.org/..."
+  [info]
+  (if-let [url (cond
+                 (not (str/blank? (:url info)))
+                 (str "https://clojure.org/" (:url info))
+                 
+                 (:special-form info)
+                 (str "https://clojure.org/special_forms#" (:name info)))]
+    (assoc info :url url)
+    info))
+
 (defn- maybe-add-file
   "If `meta-map` has no :file, assoc the :namespace file into it."
   [{:keys [file ns] :as meta-map}]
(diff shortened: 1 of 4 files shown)


src/cider/nrepl/middleware/util/meta.clj
```diff
@@ -80,49 +93,30 @@
...
-          (= '& orig-sym) (dissoc :ns)
-
-          true (assoc :url (if (contains? m :url)
-                             (when (:url m)
-                               (str "https://clojure.org/" (:url m)))
-                             (str "https://clojure.org/special_forms#" (:name m)))))))
-    (catch NoClassDefFoundError _)
-    (catch Exception _)))
+  (let [orig-sym sym
+        sym (get '{& fn*, catch try, finally try} sym sym)
```
bbatsov: Maybe this should be some constant?

src/cider/nrepl/middleware/util/meta.clj
```diff
@@ -80,49 +93,30 @@
   (if-let [ns (find-ns ns)]
     (ns-aliases ns)))
 
-;; This reproduces the behavior of the canonical `clojure.repl/doc` using its
-;; internals, but returns the metadata rather than just printing. Oddly, the
-;; only place in the Clojure API that special form metadata is available *as
-;; data* is a private function. Lame. Just call through the var.
```
bbatsov: You had some great rationale on what's true a special form, so you might put it here for future reference.

test/spec/cider/nrepl/middleware/info_test.clj
```diff
@@ -17,24 +17,31 @@
 
 (use-fixtures :each session/session-fixture)
 (deftest integration-test
-  (testing "spec"
+  (testing "spec info on a normal funcion with spec"
```
bbatsov: function

src/cider/nrepl/middleware/util/meta.clj
jpmonettas: Ratinale added!

test/spec/cider/nrepl/middleware/info_test.clj
jpmonettas: Corrected that!

src/cider/nrepl/middleware/util/meta.clj
jpmonettas: Now we have special-sub-symbs

PR #474, opened 2018-01-17
add stdout and stderr to forking-printer

hojinatmixpanel:
Instead of redirecting `*out*` and `*err*` to print on sessions, keep `*out*` and `*err*` in addition to printing on sessions.  

I don't know whether `*out*` and `*err*` are intentionally redirected to session outs/errs only. It sometimes causes a problem since log functions using stdout and stderr are muted after connecting/disconnecting sessions. 

Thanks!

- [x] All tests are passing

Thanks!


bbatsov:
> I don't know whether *out* and *err* are intentionally redirected to session outs/errs only. It sometimes causes a problem since logging functions using stdout and stderr are muted after connecting/disconnecting sessions.

Can you elaborate on this? They were redirected so they'd end up in the REPL buffer instead some nREPL server process buffer (for instance).

hojinatmixpanel:
Thanks for your prompt reply. 

We are using `cider-nrepl` to manage and debug a system remotely. And the system uses `slf4j` to log and dump logs to `stderr`. As you said, `cider-nrepl` overrides `stderr` in such way that all outputs go to the REPL buffers only, not nREPL server process buffer, i.e., the original `stderr`.

After `cider-connect` to the system, the system is not logged anymore. We thus inevitably reboot  the system to recover the `stderr` after `cider-connect`.

bbatsov:
Hmm, that's an interesting problem. Why don't you simply log to a file? Logging to `stderr` doesn't seem like a very wise decision to me in general. :-)

hojinatmixpanel:
Yeah, logging into a file is a good option. 

We are using `docker` and `kubernetes` to manage many different microservices, which turns out to be very convenient and efficient. Logging via `stdout` and `stderr` seems to be a preferable way to log microservices based on `kubernetes` and etc, see https://kubernetes.io/docs/concepts/cluster-administration/logging/

> Everything a containerized application writes to stdout and stderr is handled and redirected somewhere by a container engine.

bbatsov:
I see. It seems to me the best way to handle this would be to make it configurable, as I don't want unconditionally to dump the output and the error in both the REPL and the server process buffer. I guess we can add some var controlling this behavior and a trivial nREPL op which you can invoke from CIDER that sets it to true/false.

bbatsov:
@gonewest818 @xiongtx Ideas from you on how to best approach this would be welcome! :-)

gonewest818:
I agree with the motivation ... logging to stdout/stderr is one of the principles of the Twelve Factor App (["a twelve factor app never concerns itself with routing or storage of its output stream"](https://12factor.net/logs)).

So yes, I can see providing an option to fork the *out* and *err* as long as disconnecting the repl leaves the server process in a viable state, such that it can stream log events to the infrastructure.

Arguably if you're truly doing a 12 Factor App it shouldn't be a problem terminating the microservice when you're done debugging, because processes are [stateless and share-nothing](https://12factor.net/processes), and [disposable](https://12factor.net/disposability).  :)

dpsutton:
One issue to keep in mind that others have run into. When using CIDER on a production app, the way it took over `*out*` would break logging for some people. If you collect logs from a container from writing to std-out, this would prevent any more writing there, necessitating a restart of production instances to restore the correct out printstream.

just something to keep in mind and fix if possible in a new implementation.

hojinatmixpanel:
Thanks for bringing this up again.

I guess it is a good idea to have such options. I can look into it later this week or so.

bbatsov:
I guess I could have also just reverted the commit that introduced this change (https://github.com/clojure-emacs/cider-nrepl/commit/3630fea6b790283269a2a863cb4f33e9b50c927f#diff-ddde73de35e28688b18e9b4c4b8e2795), but whatever.

PR #475, opened 2018-01-19
Fix CLJS tests

xiongtx:
The `:test-cljs` profile uses ClojureScript `1.7.189`, which uses `(require ...)` instead of `(:require ...)` in the ns form of `base64_vlq.clj`. This is fixed in clojure/clojurescript@48b83d9.

b63b153 tries to correct this by specifying a higher ClojureScript version in the `:1.9` profile, but this fails because the `:test-cljs` profile comes later.

We resolve this by matching each version of Clojure with a compatible version of Clojurescript in the version profiles.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,73 +6,68 @@ cache:
     - $HOME/.m2
 script:
   - lein source-deps
-  - $TEST_CMD
+  - |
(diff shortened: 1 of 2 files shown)


bbatsov:
Nicely done! 🙇

PR #476, opened 2018-01-19
parser.clj (parse-info): Return :type for fields and methods

abo-abo:
Allows e.g. to have the following sample of tags for
java.util.ArrayList.java:

    int MAX_ARRAY_SIZE
    void grow (int minCapacity)
    int hugeCapacity (int minCapacity)
    int size ()
    boolean isEmpty ()
    boolean contains (Object o)
    int indexOf (Object o)
    int lastIndexOf (Object o)
    Object clone ()
    Object[] toArray ()
    Object[] toArray (Object[] a)

### DIFF
--- a/src/cider/nrepl/middleware/util/java/parser.clj
+++ b/src/cider/nrepl/middleware/util/java/parser.clj
@@ -216,10 +216,12 @@
   MethodDoc
   (parse-info [m]
     {:argtypes (mapv #(-> ^Parameter % .type typesym) (.parameters m))
-     :argnames (mapv #(-> ^Parameter % .name symbol) (.parameters m))})
+     :argnames (mapv #(-> ^Parameter % .name symbol) (.parameters m))
+     :type (str (.returnType m))})
 
   FieldDoc
-  (parse-info [f])
+  (parse-info [f]
+    {:type (str (.type f))})
 
   ClassDoc
   (parse-info [c]


src/cider/nrepl/middleware/util/java/parser.clj
```diff
@@ -216,10 +216,12 @@
   MethodDoc
   (parse-info [m]
     {:argtypes (mapv #(-> ^Parameter % .type typesym) (.parameters m))
-     :argnames (mapv #(-> ^Parameter % .name symbol) (.parameters m))})
+     :argnames (mapv #(-> ^Parameter % .name symbol) (.parameters m))
+     :type (str (. m returnType))})
```
xiongtx: `(.returnType m)`

src/cider/nrepl/middleware/util/java/parser.clj
```diff
@@ -216,10 +216,12 @@
...
 
   FieldDoc
-  (parse-info [f])
+  (parse-info [f]
+    {:type (str (. f type))})
```
xiongtx: `(.type f)`

src/cider/nrepl/middleware/util/java/parser.clj
abo-abo: I prefer `(. m returnType)`. In line with Java being a language designed for an IDE, you never type out:

```java
m.returnType
```

You type `m.` and get a list of completions based on the class of `m`.

I do the same in Clojure: type `(. m )` and get a list of completions for the class of `m`.

bbatsov: But it's not like this in Clojure - as you don't really know the type of the receiver you get completions for possible classes anyways. At any rate - consistency in a codebase is important and we've always used the syntax suggested by @xiongtx so far, so please change this.

bbatsov:
How do you plan to use this on the client side?

abo-abo:
See commit message, to generate semantic tags.

bbatsov:
Got it. I guess you meant to write "semantic", but wrote "sample" instead in the commit message.

abo-abo:
> Got it. I guess you meant to write "semantic", but wrote "sample" instead in the commit message.

By "semantic" I mean tags that are retrieved from the AST, not just extracted from the source with elaborate regexes.

By "sample" I mean a sample of `imenu`-like strings for java.util.ArrayList.java; I don't include them all :)

bbatsov:
Now I finally got it. :D Can you move this PR here https://github.com/clojure-emacs/orchard/ ?

I'm now in the process of decoupling the middleware from the generic code and this file is now living here https://github.com/clojure-emacs/orchard/blob/master/src/orchard/java/parser.clj

PR #477, opened 2018-01-20
Update see-alsos from the latest ClojureDocs database dump

ckoparkar:

- [ ] All tests are passing

Thanks!


ckoparkar:
From what I can tell, Travis is using an older version of `orchard` and it's not able to find `as-sym` at `pprint.clj:40`. I fixed the failures on my local machine by bumping `orchard`'s version to `0.1.1-SNAPSHOT`. But it doesn't feel like that' s the right thing to do here.

bbatsov:
>  Maybe it just needs another deploy ?

Yeah, I just pushed the latest version to Clojars. My bad. 

Thanks for this quick response with this!

I just had one thought about this - probably we need to check whether all the functions that are related to the one we're checking for are actually bound, so on older Clojures we won't return something that doesn't exist yet. Not a big deal, but simple to implement.

ckoparkar:
No problem! Feels good to have a CIDER commit after a long long time :-)

bbatsov:
It's nice to have you around again! And there are plenty of tasks that you can help us with. 😉

PR #478, opened 2018-01-21
Add test extension to include expected / actual diffs

### LINKED ISSUE #336: diff expected and actual values in test result
While it's currently possible to diff the output of a test with ediff on emacs, i still find spotting the differences hard due to the output that clojure test generates. @pjstadig created a project called humane-test-output that uses `clojure.data/diff` to generate a diff between the actual and expected data structures. This comes in very handy spotting the differences. 

Whereas clojure.test outputs

```
   FAIL in (a-test) (humane_test_output_test.clj:7)
   FIXME, I fail.
    expected: (= {:foo :bar, :baz :quux, :something "a long string?", :another-key "and another value"} {:fo :bar, :baz :quux, :something "a long string?", :another-key "and another value"})
     actual: (not (= {:another-key "and another value", :foo :bar, :something "a long string?", :baz :quux} {:another-key "and another value", :something "a long string?", :fo :bar, :baz :quux}))

   FAIL in (a-test) (humane_test_output_test.clj:11)
   FIXME, I fail.
   expected: (= {:foo :bar, :baz :quux} {:foo :bar, :baz :quux} {:fo :bar, :baz :quux})
     actual: (not (= {:foo :bar, :baz :quux} {:foo :bar, :baz :quux} {:fo :bar, :baz :quux}))
     FAIL in (a-test) (humane_test_output_test.clj:14)

   FIXME, I fail.
   expected: (list? foo)
     actual: (not (list? {:another-key "and another value", :foo :bar, :something "a long string?", :baz :quux}))
```

humane-test-output gives

```
   FAIL in (a-test) (humane_test_output_test.clj:7)
   FIXME, I fail.
   expected: {:another-key "and another value",
              :foo :bar,
              :something "a long string?",
              :baz :quux}
     actual: {:fo :bar}
       diff: - {:baz :quux,
                :something "a long string?",
                :foo :bar,
                :another-key "and another value"}
             + {:fo :bar}

   FAIL in (a-test) (humane_test_output_test.clj:10)
   FIXME, I fail.
   expected: {:foo :bar, :baz :quux}
     actual: {:fo :bar, :baz :quux}
       diff: - {:foo :bar}
             + {:fo :bar}

   FAIL in (a-test) (humane_test_output_test.clj:13)
   FIXME, I fail.
   expected: (list? foo)
     actual: (not
              (list?
               {:another-key "and another value",
                :foo :bar,
                :something "a long string?",
                :baz :quux}))
```

Since humane-test-output has become incompatible (see #330) with cider-nrepl it might be useful the incorporate similar functionality into cider-nrepl.

xiongtx:
Fix #336

This basically brings [pjstadig/humane-test-output](https://github.com/pjstadig/humane-test-output)'s functionality into CIDER itself.

### DIFF
--- a/src/cider/nrepl/middleware/test/extensions.clj
+++ b/src/cider/nrepl/middleware/test/extensions.clj
@@ -1,0 +1,35 @@
+(ns cider.nrepl.middleware.test.extensions
+  "Extensions to `clojure.test` functionality.
+
+  These are kept in a separate namespace because they are, by definition,
+  opinionated."
+  (:require [clojure.data :as data]
+            [clojure.pprint :as pp]
+            [clojure.test :as test :refer [assert-expr]]))
+
+;; From pjstadig/humane-test-output
+;; https://github.com/pjstadig/humane-test-output
+(defmethod assert-expr '= [msg [_ expected & more]]
+  (if (seq more)
+    `(let [more# (list ~@more)
+           result# (apply = ~expected more#)]
+       (->> (if result#
+              {:type :pass}
+              {:type :fail
+               :diffs (->> (remove #(= ~expected %) more#)
+                           (map #(vector % (data/diff ~expected %))))})
+            (merge {:message ~msg,
+                    :expected ~expected
+                    :actual more#})
+            test/do-report)
+       result#)
+    `(throw (Exception. "= expects more than one argument"))))
+
+(defn diffs-result
+  "Convert diffs data to form appropriate for transport."
+  [diffs]
+  (let [pprint-str #(with-out-str (pp/pprint %))]
+    (map (fn [[a [removed added]]]
+           [(pprint-str a)
+            [(pprint-str removed) (pprint-str added)]])
+         diffs)))
--- a/src/cider/nrepl/middleware/test.clj
+++ b/src/cider/nrepl/middleware/test.clj
@@ -3,6 +3,7 @@
   {:author "Jeff Valk"}
   (:require [cider.nrepl.middleware.pprint :as pprint]
             [cider.nrepl.middleware.stacktrace :as st]
+            [cider.nrepl.middleware.test.extensions :as extensions]
             [orchard.misc :as u]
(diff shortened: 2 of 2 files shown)


src/cider/nrepl/middleware/test/extensions.clj
```diff
@@ -0,0 +1,32 @@
+(ns cider.nrepl.middleware.test.extensions
```
bbatsov: Btw, what's the idea behind having this in a separate namespace?

xiongtx: Just so we can put functionality that extends `clojure.test`'s behavior in a separate place, since these customizations are by definition opinionated.

bbatsov: Guess you can mention in the docstring this.

xiongtx: 👌

src/cider/nrepl/middleware/test/extensions.clj
```diff
@@ -0,0 +1,35 @@
...
+
+  These are kept in a separate namespace because they are, by definition,
+  opinionated."
+  (:require [clojure.data :as data]
+            [clojure.pprint :as pp]
+            [clojure.test :as test :refer [assert-expr]]))
+
+;; From pjstadig/humane-test-output
+;; https://github.com/pjstadig/humane-test-output
+(defmethod assert-expr '= [msg [_ expected & more]]
```
pjstadig: In cases where an `is` form is part of a macro expansion `assert-expr` will get called with the fully qualified name for `=` (`clojure.core/=`), so in humane-test-output I had to have a `defmethod` for that as well. If there isn't some other way to work around it, then you may want to do the same here.

xiongtx: Hmm, alright. Seems to me `assert-expr` should dispatch on the Var, not the symbol, but that's outside of our control.

PR #479, opened 2018-01-21
[Fix #457] Reinstate cloverage testing by omitting problem namespaces

### LINKED ISSUE #457: Fix code coverage reporting
Today I noticed that our code coverage reporting has been broken for ages for both coveralls.io and codecov.io. I didn't have time to look into this, but I guess something's wrong with our TravisCI config and we're not submitting any data there. 

On a related note - probably we should simply remove coveralls.io and use only codecov.io, as it seems like a nicer service. I don't even recall why were submitting data to both of them to begin with.

gonewest818:
This PR introduces the following:
 - adds a `Makefile` with targets to use in `travis.yml`, or can be used manually too
 - excludes some problematic namespaces and tests from `cloverage` instrumentation, where instrumentation would cause tests to throw exceptions and/or fail.
 - drops `cloverage --coveralls` completely, leaving only `cloverage --codecov`
 - adds Oracle JDK 9 to the test matrix

---

- [ ] [n/a] You've added tests to cover your change(s)
- [ ] [broken] All tests are passing
- [ ] [n/a] The new code is not generating reflection warnings
- [ ] [n/a] You've updated the readme (if adding/changing middleware)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ pom.xml.asc
 .lein-plugins
 .lein-repl-history
 .nrepl-port
+.source-deps
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,69 +5,52 @@ cache:
   directories:
     - $HOME/.m2
 script:
-  - lein source-deps
-  - |
-    if [ ! -z "$TEST_CMD" ]; then
-      $TEST_CMD;
-    else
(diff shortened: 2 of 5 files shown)


bbatsov:
💯 ❤️ 🙇

bbatsov:
Fantastic work!

PR #480, opened 2018-01-21
Only include bound vars in see-also response

ckoparkar:

- [ ] All tests are passing

Thanks!

Nice catch @bbatsov! It would prevent some odd `Symbol not resolved` errors from being displayed.

### DIFF
--- a/src/cider/nrepl/middleware/info.clj
+++ b/src/cider/nrepl/middleware/info.clj
@@ -70,8 +70,10 @@
                            (and class member) (info-java class member)
                            :else (throw (Exception.
                                          "Either \"symbol\", or (\"class\", \"member\") must be supplied")))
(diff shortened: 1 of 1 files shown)


src/cider/nrepl/middleware/info.clj
```diff
@@ -66,12 +66,15 @@
   [{:keys [ns symbol class member] :as msg}]
   (let [[ns symbol class member] (map u/as-sym [ns symbol class member])]
     (if-let [cljs-env (cljs/grab-cljs-env msg)]
+      (info-cljs cljs-env symbol ns)
```
ckoparkar: Did you intend to only remove the `(handle-cljx-sources ...)` call ?

bbatsov: Ops, my bad!

bbatsov:
👍

PR #485, opened 2018-01-23
formatting tweaks to satisfy cljfmt in CI

gonewest818:
This is the result of using `cljfmt fix` to adjust formatting that it doesn't agree with.

----

- [ ] All tests are passing

Thanks!

### DIFF
--- a/src/cider/nrepl/middleware/debug.clj
+++ b/src/cider/nrepl/middleware/debug.clj
@@ -211,9 +211,9 @@ this map (identified by a key), and will `dissoc` it afterwards."}
                   (let [root-ex# (#'clojure.main/root-cause e#)]
                     (when-not (instance? ThreadDeath root-ex#)
                       (debugger-send
-                        {:status :eval-error
-                         :causes [(let [causes# (stacktrace/analyze-causes e# (:pprint-fn *msg*))]
-                                    (when (coll? causes#) (last causes#)))]})))
+                       {:status :eval-error
+                        :causes [(let [causes# (stacktrace/analyze-causes e# (:pprint-fn *msg*))]
+                                   (when (coll? causes#) (last causes#)))]})))
                   error#))]
      (if (= error# ~sym)
        ~error-expr
@@ -284,12 +284,12 @@ this map (identified by a key), and will `dissoc` it afterwards."}
   "Create a dummy exception, send its stack."
   []
   (debugger-send
-    {:status :stack
-     :causes [{:class "StackTrace"
-               :message "Harmless user-requested stacktrace"
-               :stacktrace (-> (Exception. "Dummy")
-                               (stacktrace/analyze-causes (:pprint-fn *msg*))
-                               last :stacktrace)}]}))
+   {:status :stack
+    :causes [{:class "StackTrace"
+              :message "Harmless user-requested stacktrace"
+              :stacktrace (-> (Exception. "Dummy")
+                              (stacktrace/analyze-causes (:pprint-fn *msg*))
+                              last :stacktrace)}]}))
 
 (def debug-commands
   {"c" :continue
@@ -392,7 +392,7 @@ this map (identified by a key), and will `dissoc` it afterwards."}
                              (merge var-meta)
                              (assoc :file full-path))]
           (let [iform `(with-initial-debug-bindings
-                           ~(ins/instrument-tagged-code (debug-reader form)))]
+                         ~(ins/instrument-tagged-code (debug-reader form)))]
             ;; (ins/print-form iform true)
             (eval iform)
             (let [instrumented @v]
@@ -436,7 +436,6 @@ this map (identified by a key), and will `dissoc` it afterwards."}
            (not (:macro m))
            (not (:inline m))))))
 
-
 ;;;  ## Breakpoint logic
 
 (def ^:dynamic *tmp-forms* (atom {}))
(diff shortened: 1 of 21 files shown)


src/cider/nrepl/middleware/debug.clj
```diff
@@ -355,16 +355,16 @@ this map (identified by a key), and will `dissoc` it afterwards."}
                        (assoc dbg-state :inspect)
                        (recur value))
       :inspect    (try-if-let [val (read-eval-expression "Inspect value: " dbg-state code)]
-                    (->> (debug-inspect page-size val)
-                         (assoc dbg-state :inspect)
-                         (recur value))
-                    (recur value dbg-state))
+                              (->> (debug-inspect page-size val)
```
bbatsov: This looks like a macro, but it got indented as a function.

src/cider/nrepl/middleware/debug.clj
```diff
@@ -452,18 +451,18 @@ this map (identified by a key), and will `dissoc` it afterwards."}
...
   ;; NOTE: *msg* is the message that instrumented the function,
   `(let [~'STATE__ {:msg ~(let [{:keys [code id file line column ns]} *msg*]
-                           (-> {:code code,
+                            (-> {:code code
                                 ;; Passing clojure.lang.Namespace object
                                 ;; as :original-ns breaks nREPL in bewildering
                                 ;; ways.
-                                :original-id id, :original-ns (str (or ns *ns*)),
-                                :file file, :line line, :column column}
+                                 :original-id id, :original-ns (str (or ns *ns*))
```
bbatsov: I see here the code and the comments got misaligned.

src/cider/nrepl/middleware/info.clj
```diff
@@ -22,41 +22,41 @@
   [ns sym]
   (or
     ;; it's a special (special-symbol?)
-    (m/special-sym-meta sym)
+   (m/special-sym-meta sym)
```
bbatsov: Another case of misaligned code and comments.

bbatsov:
Most changes seem OK, but I also see a few questionable ones. I've left some notes inline, but you might want to go over that diff one more time, as I probably missed something.

src/cider/nrepl/middleware/debug.clj
xiongtx: I wonder if this is related to weavejester/cljfmt#111.

gonewest818:
Will do. This is verbatim output from `lein cljfmt fix` so I guess this is really about adding config to describe those cases.

bbatsov:
Somewhere there might also be mismatch between `clojure-mode` and `cljfmt`, although I doubt there are many such cases.

src/cider/nrepl/middleware/debug.clj
gonewest818: Yes, this looks like a bug, and I opened weavejester/cljfmt#112. Try editing the comments inside that hash-map to change their indentation any way you like, and `lein cljfmt check` doesn't complain about it nor does `lein cljfmt fix` apply a fix.

For now I'll just make the fix by hand.

src/cider/nrepl/middleware/info.clj
gonewest818: again, weavejester/cljfmt#112.

src/cider_nrepl/plugin.clj
```diff
@@ -27,15 +27,15 @@
                                   ;; In this case the onus is on the user. A warning will be emitted
                                   ;; later, but we assume that the user will provide an appropriate
                                   ;; implementation.
-                                  ,,true
+                                  true
```
gonewest818: This usage of `,,true` seems intentional... are the commas a convention of some sort that cljfmt doesn't know about?

xiongtx: No, they're just an @arrdem idiosyncrasy.

arrdem: weavejester/cljfmt#67 it me

gonewest818: Ok to remove them and make cljfmt happy? Otherwise I can leave them, and you'll have warnings in CI until/unless weavejester/cljfmt#67 gets addressed.  My vote would be remove the commas to avoid masking other formatting issues from creeping in.

xiongtx: Yes, remove them. 🔨  @arrdem's dreams of being a stylistic trendsetter.

gonewest818: Ok, then that's what merging this will do.

test/clj/cider/nrepl/middleware/debug_test.clj
```diff
@@ -194,10 +194,10 @@
   (with-redefs [d/read-debug-command (fn [v e] (assoc e :value v))
                 d/debugger-message   (atom [:fake])
                 d/*skip-breaks*      (atom nil)]
-    (binding [*msg* {:session (atom {}) :code :code, :id     :id,
+    (binding [*msg* {:session (atom {}) :code :code, :id     :id
                      :file    :file,    :line :line, :column :column}]
       (let [m (eval `(d/with-initial-debug-bindings
```
gonewest818: `cljfmt`'s inconsistency with `,` characters is irritating.

gonewest818:
I've looked through again, and this looks correct to me.  CI tests are still broken because of the unrelated `mranderson` issue (indeed, though, the cljfmt check passes!)

I suggest merging to get back to green and have meaningful feedback for the ongoing orchard work.

bbatsov:
Fair enough! 👍

PR #490, opened 2018-01-25
Fix classpath error handling redef arity

dpsutton:
`classpath-reply` takes a single arg, msg; is not a thunk

last failing test was an arity mismatch on classpath-repl


dpsutton:
`lein with-profile +test-clj test` is what i was running locally which showed two assertions in the same test failing.

bbatsov:
👍 Thanks for fixing my fuckup! I didn't see this between the missing namespace errors. One more argument in favour of not trying to fix problems in a hurry. 😆

PR #494, opened 2018-01-26
Disable failing tests

ckoparkar:
These failures are unrelated to the nREPL itself.

1. CLJS doesn't build on JDK9 because it depends on a package deprecated in JDK9 (javax.xml.bind). CLJS-2377 tracks it. It has been fixed on master, but not yet released. After that, we can enable CLJS-1.9 tests on JDK9, but not the others.

2. CLJS-1.9 also doesn't build on JDK7 anymore ? I think this is because of a dependency on the closure-compiler which requires JDK8 or higher (https://github.com/google/closure-compiler/wiki/Releases#september-10-2017-v20170910)

/cc @gonewest818 

- [ ] All tests are passing

Thanks!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,6 +44,13 @@ matrix:
       jdk: oraclejdk8
       after_success: bash <(curl -s https://codecov.io/bash) -f target/coverage/codecov.json
 
+  exclude:
+    # CLJS-1.9 doesn't build on JDK7. This is because of a dependency on the
+    # google/closure-compiler which requires JDK8 or higher.
+    # https://github.com/google/closure-compiler/wiki/Releases#september-10-2017-v20170910
+    - env: VERSION=1.9 TARGET=test-cljs
+      jdk: openjdk7
+
   fast_finish: true      # don't wait for allowed failures before build finish
   allow_failures:
     - env: VERSION=master TARGET=test-clj
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,9 @@
 VERSION ?= 1.9
 export CLOVERAGE_VERSION = 1.0.11-SNAPSHOT
 
+# The test-cljs target needs to be modified if working with JDK9
+JAVA_VERSION = $(shell lein version | cut -d " " -f 5 | cut -d "." -f 1-2)
+
 .source-deps:
 	lein source-deps
 	touch .source-deps
@@ -13,7 +16,13 @@ test-clj: .source-deps
 	lein with-profile +$(VERSION),+test-clj test
 
 test-cljs: .source-deps
-	lein with-profile +$(VERSION),+test-cljs test
+	if [ "$(JAVA_VERSION)" = "9.0" ] || [ "$(JAVA_VERSION)" = "9-internal" ]; then \
+            lein with-profile +$(VERSION),+test-cljs \
+                 update-in :jvm-opts concat '["--add-modules" "java.xml.bind"]' \
+                 -- test; \
+        else \
+            lein with-profile +$(VERSION),+test-cljs test; \
+	fi
 
 eastwood: .source-deps
 	lein with-profile +$(VERSION),+test-clj,+test-cljs,+eastwood eastwood


ckoparkar:
All tests pass! Yay!

.travis.yml
```diff
@@ -44,6 +44,16 @@ matrix:
       jdk: oraclejdk8
       after_success: bash <(curl -s https://codecov.io/bash) -f target/coverage/codecov.json
 
+  exclude:
+    - env: VERSION=1.9 TARGET=test-cljs
```
bbatsov: I think you should also put the explanations here, so people would be able to quickly understand why've setup the build this way.

ckoparkar: Good point. Added comments there.

.travis.yml
```diff
@@ -44,6 +44,22 @@ matrix:
       jdk: oraclejdk8
       after_success: bash <(curl -s https://codecov.io/bash) -f target/coverage/codecov.json
 
+  exclude:
+    # CLJS doesn't build on JDK9 because it depends on a package (javax.xml.bind)
+    # deprecated in JDK9. See (CLJS-2377).
+    # After a CLJS release, the VERSION=1.9 build can be re-enabled.
```
gonewest818: In https://github.com/bhauman/lein-figwheel/issues/612 another way to handle this was discussed.  Conditionally adding ` :jvm-opts ["--add-modules" "java.xml.bind"]` to the build.

ckoparkar: I considered doing this. So we could add something like:
```
:jvm-opts ~(try (do (Class/forName "javax.xml.bind.DatatypeConverter")
                     [])
                (catch ClassNotFoundException e  ["--add-modules" "java.xml.bind"]))
```

But this seemed like too much of a hack just to get the tests working. Should we  take this path ?

ckoparkar: On second thought, I should've mentioned this in the PR as well. Thanks for bringing it up.

gonewest818: No worries, I just wanted to put it out there. I don't have a strong opinion either way.

However, that said, another equally hacky approach would be to detect the java version in the Makefile and use `lein update-in` to inject the `:jvm-opts` as needed for cljs.

The Makefile has been the place where other transient toggling is being done, for instance filtering out some cloverage tests, and turning off mranderson.

ckoparkar: Ah, that's true. One advantage of this would be that all toggling would be in one place.

But would doing this in Clojure be a bit safer than grepping the version there ? (I have to admit that I have minimal experience with Makefiles, so I might be totally wrong. I used `Class/forName ` because of Alex Miller's  [comment](https://github.com/bhauman/lein-figwheel/issues/612#issuecomment-353169205) on the same figwheel issue.)

If we're not disabling the tests and adding java-opts instead, I'm leaning toward adding it in the `project.clj` file and using `Class/forName`. But I'm fine with the Makefile version too.

gonewest818: Well, for instance in orchard there's this
https://github.com/clojure-emacs/orchard/blob/master/Makefile#L5-L8

which detects Java 9 and constructs a test selector `:java9.0` which is then used to filter certain test cases based on metadata (`lein with-profile <+1.9,...etc.> test :java9.0`).  The usage would be different in this case, but we're already taking a similar risk elsewhere.

Makefile
```diff
@@ -13,7 +16,11 @@ test-clj: .source-deps
 	lein with-profile +$(VERSION),+test-clj test
 
 test-cljs: .source-deps
+ifeq ($(JAVA_VERSION),:java9)
+	lein with-profile +$(VERSION),+test-cljs,+jxb test
+else
 	lein with-profile +$(VERSION),+test-cljs test
+endif
```
ckoparkar: Oh well. Let's go ahead with the Makefile version then. Does this look ok ? I couldn't figure out how to use both `with-profile` and `update-in` at the same time. I'm not using Clojure/Leiningen for a while now. And I ran out of make-fu :P . Could you please help with this ?

Makefile
```diff
@@ -3,6 +3,9 @@
 VERSION ?= 1.9
 export CLOVERAGE_VERSION = 1.0.11-SNAPSHOT
 
+# The test-cljs target needs to be modified if working with JDK9
+JAVA_VERSION = :java$(shell lein version | cut -d " " -f 5 | cut -d "." -f 1-2)
```
gonewest818: This is going to produce a string like `:java9.0` (or `:java1.8`, `:java1.7`, it's annoying that the numbering convention changed).

Suggest you simplify by dropping the ":java" part:
```
JAVA_VERSION = $(shell lein version | cut -d " " -f 5 | cut -d "." -f 1-2)
```
which will produce a number like "#.#"

Makefile
gonewest818: Does that work? I've never tried expressing a recipe that way. I would have done the conditional test in bash syntax, ie. all the logic completely inside the recipe.

```
test-cljs: .source-deps
        if [ "$(JAVA_VERSION)" = "9.0" ]; then \
            lein with-profile +$(VERSION),+test-cljs,+jxb test; \
        else \
            lein with-profile +$(VERSION),+test-cljs test; \
        fi
```

Also, now that I'm thinking about it...   I wonder if it's more future-proof to name the Leiningen profile "+java9.0" rather than "+jxb"?  In the event the project requires other Java 9 modules down the road then this mechanism for adding modules has a general label instead of the specific one.

gonewest818: while I'm at my keyboard I'll take a look at the `update-in` usage also.

gonewest818: Doing it with `update-in` would look something like this:
```
test-cljs: .source-deps
        if [ "$(JAVA_VERSION)" = "9.0" ]; then \
            lein with-profile +$(VERSION),+test-cljs \
                 update-in :jvm-opts concat "[\"--add-module\" \"java.xml.bind\"]" \ 
                 -- test; \
        else \
            lein with-profile +$(VERSION),+test-cljs test; \
        fi
```

going this route you don't need the `:jxb` profile in `project.clj` at all.

Makefile
```diff
@@ -13,7 +16,13 @@ test-clj: .source-deps
 	lein with-profile +$(VERSION),+test-clj test
 
 test-cljs: .source-deps
-	lein with-profile +$(VERSION),+test-cljs test
+	if [ "$(JAVA_VERSION)" = "9.0" ] || [ "$(JAVA_VERSION)" = "9" ]; then \
+            lein with-profile +$(VERSION),+test-cljs \
+                 update-in :jvm-opts concat '["--add-modules" "java.xml.bindgn"]' \
```
ckoparkar: Thanks a lot @gonewest818! I've pretty much used your code verbatim, except for using single quotes here.  Escaping the double quotes didn't work.

The `jxb` profile sure was ugly. This looks so much better.

Makefile
```diff
@@ -13,7 +16,13 @@ test-clj: .source-deps
 	lein with-profile +$(VERSION),+test-clj test
 
 test-cljs: .source-deps
-	lein with-profile +$(VERSION),+test-cljs test
+	if [ "$(JAVA_VERSION)" = "9.0" ] || [ "$(JAVA_VERSION)" = "9" ]; then \
```
ckoparkar: Also, on my linux machine, JAVA_VERSION is simply "9" instead of "9.0". Is it different on yours ?

Makefile
```diff
@@ -13,7 +16,13 @@ test-clj: .source-deps
...
+            lein with-profile +$(VERSION),+test-cljs \
+                 update-in :jvm-opts concat '["--add-modules" "java.xml.bind"]' \
```
ckoparkar: Thanks a lot @gonewest818! I've pretty much used your code verbatim, except for using single quotes here. Escaping the double quotes didn't work.

The jxb profile sure was ugly. This looks so much better.

Makefile
ckoparkar: On Travis it returns "9.0"

gonewest818: right... I was in the middle of typing the same thing.  The `orchard` builds are all getting "9.0" on the Travis CI infrastructure running Ubuntu 14.04.

What distribution are you running?

ckoparkar: I'm using OpenJDK on Archlinux.

Should we leave the `= "9"` around,  just in case ?

gonewest818: I'm launching a Docker container with Ubuntu 16.04 and `apt-get install openjdk-9-jdk` to see. I assume that's what Travis CI will move to next.

gonewest818:
```
root@2735a24c8c08:~# ./lein version
Leiningen 2.8.1 on Java 9-internal OpenJDK 64-Bit Server VM
root@2735a24c8c08:~# ./lein version | cut -d " " -f 5
9-internal
root@2735a24c8c08:~# ./lein version | cut -d " " -f 5 | cut -d "." -f 1-2
9-internal
```

That `= "9"` doesn't look like it's going to work.  As of right now OpenJDK 9 on Ubuntu 16.04 calls itself `9-internal` and you could use that, but really we should open a ticket that we need a way to detect Oracle and OpenJDK versions and report back consistently. I'd be shocked if something better isn't available.

ckoparkar: Done.

bbatsov:
Nice! I was completely fine with the original solution, but you've managed to improve it quite a bit! Very useful (and also very educational for me)!

Thanks, guys! 🙇

ckoparkar:
Thanks a lot for all the help @gonewest818!

PR #496, opened 2018-01-26
stop `lein source-deps` since we're not using mranderson

gonewest818:
For now source-deps isn't relevant, is it?  It just makes the test jobs run longer.


gonewest818:
Closing this.  I'll include it in the "build stages" setup for auto-deploy.

PR #501, opened 2018-01-30
[Fix #500] simplify java version detection

### LINKED ISSUE #500: Use a standardized method to detect JDK version
To quote @gonewest818, "we need a way to detect Oracle and OpenJDK versions and report back consistently". See this [discussion](https://github.com/clojure-emacs/cider-nrepl/pull/494#discussion_r16426233) for more details.

The JAVA_VERSION is used to do some JDK specific things (eg. filter out certain tests in orchard, add `:jvm-opts` in the nREPL etc.). There are some inconsistencies with the current method;

`(lein version | cut -d " " -f 5 | cut -d "." -f 1-2)`

| Environment | Script result |
| ------------- |:-------------:|
| My local m/c running Archlinux and OpenJDK | `9` |
| Travis CI infrastructure running Ubuntu 14.04| `9.0` |
| @gonewest818's experiment with a Docker container<br> running Ubuntu 16.04 and OpenJDK 9 | `9-internal`|

This is not a high priority ticket at the moment. But if the Travis tests start failing, this might be one reason why.

gonewest818:
* `lein-sysutils` offloads the parsing of the JDK version number to the commons-lang3 library. The benefit is that we don't have to invent tricky regex syntax and anticipate new versioning conventions such as JEP 223 (http://openjdk.java.net/jeps/223). `lein-sysutils` uses reflection to parse the available fields in org.apache.commons.lang3.SystemUtils, which means this mechanism should be fairly future-proof as long as the underlying library is actively maintained.

* Also fixes two CI issues by updating the conditional test in the deploy stage in `.travis.yml`
  1. While `make release` creates a new version tag from which the
release should build and deploy, it turns out the resulting CI job
does not run the deploy task, because for those jobs the branch name
is not "master".  With this commit we allow deploys from branches
where the name matches "v#.#.#".
  2. Anyone running Travis CI builds on their own fork of this project
ended up with failed deploy jobs because the Clojars credentials could
not be parsed. With this commit we exclude deploy jobs from even
running if we're on a fork.

----

- [ ] You've added tests to cover your change(s) *N/A*
- [x] All tests are passing *EXCEPT eastwood which was broken before*
- [ ] The new code is not generating reflection warnings *N/A*
- [ ] You've updated the readme (if adding/changing middleware) *N/A*

Thanks!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,8 @@ VERSION ?= 1.9
 export CLOVERAGE_VERSION = 1.0.11-SNAPSHOT
 
 # The test-cljs target needs to be modified if working with JDK9
-JAVA_VERSION = $(shell lein version | cut -d " " -f 5 | cut -d "." -f 1-2)
+JAVA_VERSION = $(shell lein with-profile +sysutils \
+                       sysutils :java-version-simple | cut -d " " -f 2)
 
 .source-deps:
 	lein source-deps
@@ -16,7 +17,7 @@ test-clj: # .source-deps
 	lein with-profile +$(VERSION),+test-clj test
 
 test-cljs: # .source-deps
-	if [ "$(JAVA_VERSION)" = "9.0" ] || [ "$(JAVA_VERSION)" = "9-internal" ]; then \
+	if [ "$(JAVA_VERSION)" = "9" ]; then \
             lein with-profile +$(VERSION),+test-cljs \
                  update-in :jvm-opts concat '["--add-modules" "java.xml.bind"]' \
                  -- test; \
--- a/Makefile
+++ b/Makefile
@@ -55,19 +55,10 @@ release:
 
 # Deploying requires the caller to set environment variables as
 # specified in project.clj to provide a login and password to the
-# artifact repository.  We're setting TRAVIS_PULL_REQUEST to a default
-# value to avoid accidentally deploying from the command line. Inside
-# Travis CI this variable will be set to the pull request number, or
-# to "false" if it's not a pull request.
-
-TRAVIS_PULL_REQUEST ?= "true"
+# artifact repository.
 
 deploy: .source-deps
-	if [ "$(TRAVIS_PULL_REQUEST)" != "false" ]; then \
-	    echo "Pull request detected. Skipping deploy."; \
-	else \
-	    lein with-profile +$(VERSION),+plugin.mranderson/config deploy clojars; \
-	fi
+	lein with-profile +$(VERSION),+plugin.mranderson/config deploy clojars
 
 clean:
 	lein clean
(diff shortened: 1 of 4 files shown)


gonewest818:
The job https://travis-ci.org/clojure-emacs/cider-nrepl/jobs/335331901 failed but if you restart it, it should succeed. The same job runs successfully in my fork.

Makefile
```diff
@@ -4,7 +4,8 @@ VERSION ?= 1.9
 export CLOVERAGE_VERSION = 1.0.11-SNAPSHOT
 
 # The test-cljs target needs to be modified if working with JDK9
-JAVA_VERSION = $(shell lein version | cut -d " " -f 5 | cut -d "." -f 1-2)
+JAVA_VERSION = $(shell lein with-profile +sysutils \
+                       sysutils :java-version-simple | cut -d " " -f 2)
```
bbatsov: Btw, I would have named this "Major", not "Simple" if I were you.

gonewest818: Yep, I considered that, but thought people might object that "1.7" isn't "major" in the SemVer sense. Substituting another arbitrary semantic on top of the existing one. Do you feel strongly about it? This is all new, easy to change and re-release.

bbatsov: Not at all. Just random 2 cents.

src/cider_nrepl/main.clj
```diff
@@ -1,7 +1,7 @@
 (ns cider-nrepl.main
   (:require
-    [clojure.java.io :as io]
-    [clojure.tools.nrepl.server :as nrepl.server]))
+   [clojure.java.io :as io]
```
bbatsov: Btw, this seems unrelated to your changes - how was the build passing before?

bbatsov:
I guess the final commit should be applied to `orchard` as well, right?

> The job https://travis-ci.org/clojure-emacs/cider-nrepl/jobs/335331901 failed but if you restart it, it should succeed. The same job runs successfully in my fork.

Indeed.

src/cider_nrepl/main.clj
SevereOverfl0w: emacs vs vim indentation I think.

gonewest818:
I should have separated that final commit. Not strictly related to this work.  yes, needs to be done in orchard too.

src/cider_nrepl/main.clj
gonewest818: Cljfmt flagged the formatting of those lines. However cljfmt, Eastwood, and cloverage happen to be configured as allowed_failure in CI, so I assume someone’s editor is configured differently and that went unnoticed.

bbatsov: Got it!

PR #503, opened 2018-02-09
Honor clojure.tools.namespace config and conventions

plexus:
There are two things here: honor `set-refresh-dirs`, and honor `::load` metadata.

The directories to scan for a refresh can be configured with
`clojure.tools.namespace/set-refresh-dirs`. This makes CIDER honor this
configuration when set.

Specific namespaces can be excluded from being loaded or unloaded by setting
metadata on the namespace, for which there are convenience methods in
clojure.tools.namespace.repl. The docstrings of these methods say that setting
`:clojure.tools.namespace.repl/load false` should imply
`:clojure.tools.namespace.repl/unload false`. This makes CIDER honor that
convention.

- [X] All tests are passing

Thanks!

### DIFF
--- a/src/cider/nrepl/middleware/refresh.clj
+++ b/src/cider/nrepl/middleware/refresh.clj
@@ -17,9 +17,23 @@
 
 (defonce ^:private refresh-tracker (agent (track/tracker)))
 
+(defn- user-refresh-dirs
+  "Directories to watch and reload, as configured by the user.
+
+  See `clojure.tools.namespace.repl/set-refresh-dirs`.
+
+  The var is resolved at runtime to get the \"real\" clojure.tools.namespace,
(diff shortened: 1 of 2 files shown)


bbatsov:
You've found an interesting problem with the inlined deps. I wonder if we suffer from something similar with other libraries, but I don't see an universal solution for this. 

On a relate note - it might be even better if users could set this as some defcustom which can be dir-local. Anyways, thanks for looking into this and fixing the problem!

cichli:
Good catch regarding the semantics of `::load`. Thanks for the contribution @plexus! 

>On a relate note - it might be even better if users could set this as some defcustom which can be dir-local.

@bbatsov this was my original intention when including the `dirs` option to the `refresh` op, but I never got around to implementing the elisp side 😅.

PR #507, opened 2018-02-14
Fix #489 - reinstate mranderson on tests

### LINKED ISSUE #489: Eliminate all references to orchard in the tests
Basically ever since we started processing orchard with MrAnderson the Travis build as been broken as we can't refer to orchard from tests, because of how MrAnderson works (the test ns-es are not processed and can't find `orchard`). 

See this conversation as well

image

That's a simple task, but I don't have time for it now. I'd appreciate some help. //cc @xiongtx @gonewest818 @ckoparkar

gonewest818:
See discussion in #489 for an explanation what was done.

Kind of a lot going on here, but I made sure each "logical unit" of work is a distinct commit so you can view each piece separately.

-----

- [x] All tests are passing

Thanks!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,8 +22,8 @@ jdk:
   - oraclejdk8
   - oraclejdk9
 stages:
-  - name: test
   - name: check
+  - name: test
   # Deploy only from the home repo where the credentials can be
   # properly decrypted. Never deploy from a pull request job.
   # In addition, ensure we're on the master branch (snapshots)
@@ -78,8 +78,6 @@ jobs:
     - env: VERSION=master TARGET=test-clj
     - env: VERSION=master TARGET=test-cljs
     - env: VERSION=1.9 TARGET=cloverage
-    - env: VERSION=1.9 TARGET=eastwood
-    - env: VERSION=1.9 TARGET=cljfmt
 
 notifications:
   webhooks:
--- a/src/cider/nrepl/middleware/track_state.clj
+++ b/src/cider/nrepl/middleware/track_state.clj
@@ -3,6 +3,7 @@
   {:author "Artur Malabarba"}
   (:require [cider.nrepl.middleware.ns :as ns]
             [cider.nrepl.middleware.util.cljs :as cljs]
+            [cider.nrepl.middleware.util.meta :as um]
             [orchard.misc :as u]
             [orchard.meta :as m]
             [orchard.namespace :as namespace]
@@ -34,7 +35,7 @@
                         (when (var? the-var)
                           (let [{the-ns :ns :as the-meta} (var-meta-with-fn the-var)]
                             (when-not (identical? the-ns clojure-core)
-                              [sym (m/relevant-meta the-meta)]))))))))
+                              [sym (um/relevant-meta the-meta)]))))))))
 
 (defn- cljs-meta-with-fn
   "Like (:meta m) but adds {:fn true} if (:fn-var m) is true."
@@ -43,11 +44,6 @@
     (:fn-var m) (assoc :fn true)))
 
 ;;; Namespaces
-(defn- relevant-meta-keys
-  "Access orchard.meta/relevant-meta-keys. For use in test cases where
-  direct access to the symbol is not possible."
-  []
-  m/relevant-meta-keys)
 
 (defn ns-as-map [object]
   (cond
@@ -62,7 +58,7 @@
       {:aliases (merge require-macros requires)
        ;; For some reason, cljs (or piggieback) adds a :test key to the
        ;; var metadata stored in the namespace.
-       :interns (merge (u/update-vals #(dissoc (m/relevant-meta (cljs-meta-with-fn %)) :test)
+       :interns (merge (u/update-vals #(dissoc (um/relevant-meta (cljs-meta-with-fn %)) :test)
                                       defs)
                        ;; FIXME: `uses` and `use-macros` are maps from
                        ;; symbols to namespace names:
@@ -79,7 +75,7 @@
     {:aliases {}
      :interns (->> (ns-map clojure-core)
                    (filter #(var? (second %)))
-                   (u/update-vals #(m/relevant-meta (var-meta-with-fn %))))}))
+                   (u/update-vals #(um/relevant-meta (var-meta-with-fn %))))}))
 
 (defn calculate-changed-ns-map
   "Return a map of namespaces that changed between new-map and old-map.
--- a/src/cider/nrepl/middleware/track_state.clj
+++ b/src/cider/nrepl/middleware/track_state.clj
@@ -43,6 +43,12 @@
     (:fn-var m) (assoc :fn true)))
 
 ;;; Namespaces
+(defn- relevant-meta-keys
+  "Access orchard.meta/relevant-meta-keys. For use in test cases where
+  direct access to the symbol is not possible."
+  []
+  m/relevant-meta-keys)
+
 (defn ns-as-map [object]
   (cond
     ;; Clojure Namespaces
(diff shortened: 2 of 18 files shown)


src/cider/nrepl/middleware/info.clj
```diff
@@ -109,6 +109,14 @@
            (if resource-relative
              {:resource resource-relative}))))
 
+(defmacro with-java-api-version
```
bbatsov: Or we can just make some tests Java version specific, which might be better. This approach seems a bit ugly to me, although it makes sense.

src/cider/nrepl/middleware/track_state.clj
```diff
@@ -43,6 +43,12 @@
     (:fn-var m) (assoc :fn true)))
 
 ;;; Namespaces
+(defn- relevant-meta-keys
+  "Access orchard.meta/relevant-meta-keys. For use in test cases where
+  direct access to the symbol is not possible."
+  []
+  m/relevant-meta-keys)
```
bbatsov: I don't think this function is useful outside `cider-nrepl`, so we should probably bring it back. Might have missed a few of those when working in bulk.

bbatsov:
I'm a bit worried about some of the disabled tests and moving test code to the implementation ns, just to appease Mr. Anderson, but overall seems that's a step in the right direction.

@benedekfazekas @xiongtx Please, take a look as well.

test/clj/cider/nrepl/middleware/refresh_test.clj
```diff
@@ -129,9 +127,12 @@
     (with-redefs [resolve (constantly nil)]
       (is (nil? (#'r/user-refresh-dirs)))))
 
-  (testing "honors set-refresh-dirs"
-    (c.t.n.r/set-refresh-dirs "foo" "bar")
-    (is (= ["foo" "bar"] (#'r/user-refresh-dirs)))))
+  ;; Disabling the next test.
```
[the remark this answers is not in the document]
gonewest818: Right, I assumed the test was added while mranderson was disabled.  If I try to require `clojure.tools.namespace.repl`  (and state the dependency to leiningen under `+test-clj`) that dependency gets elided from that list ... I assume mranderson elides the duplicate entry knowing it's going to be localized anyway.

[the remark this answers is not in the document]
gonewest818: Confirmed. This test case was added a week ago. This is the commit: https://github.com/clojure-emacs/cider-nrepl/commit/b88365dfce85d18d6e8d28f17dd037ce25200e8d

src/cider/nrepl/middleware/track_state.clj
bbatsov: If a recall correctly at some point our bencode implementation was failing to encode certain complex metadata, so we decided to circumvent this by just selecting the metadata needed by CIDER instead of retrieving all metadata. This means that such a function would not be useful outside `cider-nrepl` most likely.

bbatsov:
> In particular though, on the "moving test code to the implementation ns" solely to appease mranderson, I would appreciate some more discussion what would be preferable.

Sure. My main concern is that we've done some questionable things in the past (like using `with-redefs` here and there) and that might be the real problem that needs fixing, not how to expose somehow something from orchard in the ns-rewritten tests. All of this might be an opportunity to fix a few test design problems on the side (e.g. like the one with the different java versions). Generally we should limit ourselves to testing some messages and their responses most of the time (apart for utility namespaces, of course).

src/cider/nrepl/middleware/track_state.clj
```diff
@@ -153,41 +159,53 @@
...
+  client.
+
+  The 2-arity call is the intended way to use this function.
+
+  The 4-arity call is provided for testing under mranderson.
+  Allows substitution of supporting fns in the implementation that
+  don't need to exposed otherwise. Be aware when the implementation
+  details change because this arity (and the tests) will need to
+  change also."
+  ([old-data msg]
```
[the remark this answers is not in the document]
bbatsov: Yeah, absolutely. I consider this the good thing coming from the split - we have to think more about good apis and proper separation of concerns.

.travis.yml
```diff
@@ -78,8 +78,6 @@ jobs:
     - env: VERSION=master TARGET=test-clj
     - env: VERSION=master TARGET=test-cljs
     - env: VERSION=1.9 TARGET=cloverage
-    - env: VERSION=1.9 TARGET=eastwood
-    - env: VERSION=1.9 TARGET=cljfmt
```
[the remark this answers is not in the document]
bbatsov: That's fine by me, as long as it's easy for the contributors to figure out what went wrong.

test/cljs/cider/nrepl/middleware/cljs_inspect_test.clj
```diff
@@ -1,18 +1,327 @@
...
+           (:value (session/message {:op "inspect-refresh"}))))))
+
+(deftest refresh-empty-idempotent-integration-test
+  (testing "refreshing an empty inspector renders nil"
+    (is (= nil-result
+           (:value (do
+                     (session/message {:op "inspect-refresh"})
+                     (session/message {:op "inspect-refresh"})))))))
+
+#_(deftest exception-integration-test
```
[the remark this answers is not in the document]
bbatsov: My bad. I didn't see this was the cljs test.

src/cider/nrepl/middleware/info.clj
gonewest818: Actually I don't think this is possible because `cloverage` doesn't support test selectors (see cloverage/cloverage#54). Cloverage has other mechanisms like filtering test namespaces by regex but that would also be ugly.

test/clj/cider/nrepl/middleware/util/instrument_test.clj
```diff
@@ -41,7 +22,7 @@
     '#{[(new java.lang.Integer 1) []]}))
 
 (deftest instrument-clauses-test
-  (are [exp res] (clojure.set/subset? res (breakpoint-tester exp))
+  (are [exp res] (clojure.set/subset? res (t/breakpoint-tester exp))
```
xiongtx: Should `(:require [clojure.set :as set])` in the `ns` form

src/cider/nrepl/middleware/util/instrument.clj
```diff
@@ -342,3 +342,27 @@
     (->> (ns-interns ns)
          (filter (comp :cider/instrumented meta second))
          (map first))))
+
+;;; Instrumentation test support
+;;;
+;;; This code migrated out of the test namespace to avoid a dependency
```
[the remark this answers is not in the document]
gonewest818: Right, like I said elsewhere, it becomes a discussion about the design of some of these components. Do they really need to expose an implementation detail (the use of orchard or other mranderson-ed libraries) to the caller and/or the test harness?

The orchard dependency is `breakpoint-tester` down below... calls `m/strip-meta` and `m/macroexpand-all`.

xiongtx:
Per a [previous post](https://github.com/clojure-emacs/cider-nrepl/issues/489#issuecomment-360571858) I'm not a fan of the constraint imposed on testing by `mranderson`.

> Philosophically, there should be nothing we cannot use in the tests that we use in the src. `mranderson`’s impact should purely be on the _distribution_ of the artifact--we apply it to prevent tooling deps from affecting user’s project deps. For `mranderson` to impact the way we _develop_ and _test_ the project is an anti-pattern.

It makes no sense for `cider-nrepl`'s tests to be barred from access to functions defined in its dependencies. It's one thing to _test_ `orchard`'s functionality, which we shouldn't do in `cider-nrepl`; it's another to _use_ it, which we absolutely _should_ be able to.

The ongoing work to port functionality from `cider-nrepl` to `orchard` is good. Bending over backwards to satisfy `mranderson`, a purely _distribution_ related concern, in our tests is not good--it's complecting two different concerns.

My vote would be to continue leaving `mranderson` out of testing until it supports rewriting test namespaces the way it does src. Focus on porting functionality from `cider-nrepl` to `orchard` w/out worrying about dividing the interface to satisfy `mranderson`.

test/clj/cider/nrepl/middleware/track_state_test.clj
```diff
@@ -23,39 +20,47 @@
...
-        ;; Sent message depends on the first arg.
-        (is (= (get-sent-value full-cache)
-               (get-sent-value full-cache)))
-        (is (= (get-sent-value (into {} (drop 3 full-cache)))
-               (get-sent-value (into {} (drop 3 full-cache))))))
+    (let [new-data (update-and-send-cache-tester nil msg sent-value)]
+      (is (map? new-data))
+      (is (> (count new-data) 100)))
+    (let [{:keys [repl-type changed-namespaces]} @sent-value]
+      (is (= repl-type :clj))
```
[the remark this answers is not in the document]
gonewest818: I'm pushing a fix for this.

test/clj/cider/nrepl/middleware/util/instrument_test.clj
gonewest818: I'm pushing a fix for this too.

gonewest818:
The new Travis inconsistency is occasional jobs running > 10 minutes and then failing. I know this build succeeds in my own branch so I can't imagine why it wouldn't succeed here too.

src/cider/nrepl/middleware/info.clj
gonewest818: One other possibility I'm looking at is `circleci/circleci.test` as a substitute for the default leiningen test runner which is a "massive and terrible monkeypatch" (their words, not mine).  The main thing is that circleci.test re-implements the test selector logic and is already integrated with cloverage. The missing piece is that cloverage can't/doesn't have a command line flag to pass in a selector keyword. Which would be a pull request in cloverage. If you want to wait for that, we could try for that route.

bbatsov:
> My vote would be to continue leaving mranderson out of testing until it supports rewriting test namespaces the way it does src. Focus on porting functionality from cider-nrepl to orchard w/out worrying about dividing the interface to satisfy mranderson.

Is @benedekfazekas even working on such a feature? 

The big problem right now is that without testing the real re-written build it's easy make some wrong assumptions (e.g. that you can use some var from one of the deps to configure them, as you would normally be able to do, etc). I think the `c.t.n.` problem fixed recently by @plexus is a classic example of this.

benedekfazekas:
not at the moment but can have a look. can you pls create an issue on mranderson. along the lines of making the source dirs configurable I guess...

xiongtx:
> it's easy make some wrong assumptions (e.g. that you can use some var from one of the deps to configure them, as you would normally be able to do...

Do you mean in `cider-nrepl` code? Or in user code? 😕

bbatsov:
In the user code, as things that worked when you were using the normal `cider-nrepl` while hacking on it, won't work with the ns-rewritten version.

xiongtx:
You mean if the user tries to e.g. `set!` a
var in one of the mangled deps?

bbatsov:
Yeah.

xiongtx:
1. That's on the user not to do that--`cider-nrepl`'s deps are `cider-nrepl`'s only.
2. Not sure how including `mranderson` in the testing process helps catch these errors.

src/cider/nrepl/middleware/info.clj
gonewest818: I pushed a change that uses commons-lang3, which is a dependency leiningen already uses, to mark tests that need to run in specific jvm versions.

```
(if (SystemUtils/IS_JAVA_1_8)
  (testing "a Java 8 specific test case"
    (is (= expected (run-test)))))
```

test/cljs/cider/nrepl/middleware/cljs_inspect_test.clj
```diff
@@ -6,18 +6,19 @@
 
 (def nil-result ["(\"nil\" (:newline))"])
 
-(def var-result ["(\"Class\" \": \" (:value \"clojure.lang.Var\" 0) (:newline) \"Meta Information: \" (:newline) \"  \" (:value \":ns\" 1) \" = \" (:value \"clojure.core\" 2) (:newline) \"  \" (:value \":name\" 3) \" = \" (:value \"*assert*\" 4) (:newline) \"Value: \" (:value \"true\" 5))"])
+(def var-result ["(\"Class\" \": \" (:value \"clojure.lang.Cons\" 0) (:newline) \"Contents: \" (:newline) \"  \" \"0\" \". \" (:value \"var\" 1) (:newline) \"  \" \"1\" \". \" (:value \"cljs.core/*assert*\" 2) (:newline))"])
```
gonewest818: When all is said and done, these fixes should be squashed (at least in groups, ie. all changes to cljs_inspect_test.clj) to make the overall commit history more sensible.

gonewest818:
I think I have completed all the code movement and restructuring called out in this thread.  I want to reorder and merge some commits to clean up the commit history a little bit before anything is merged. Travis CI is in some degraded mode today due to their message queues, and I need to wait until jobs have completed here and in Orchard.

Major items yet to address is @xiongtx's open question (why do this?). And the open question to @benedekfazekas if mranderson could instead process the test namespaces as well.

But in terms of the original goal of the ticket... I think it's all here.

bbatsov:
> I think I have completed all the code movement and restructuring called out in this thread. I want to reorder and squash some commits to clean up the commit history before anything is merged. Travis CI is in some degraded mode today due to their message queues, and I need to wait until jobs have completed here and in Orchard.

OK, let me know when you're ready.

> Major items yet to address is @xiongtx's open question (why do this?). And the open question to @benedekfazekas if mranderson could instead process the test namespaces as well.

Well, we all agreed that ideally tests should be self-contained and not refer to 3rd party libraries (when feasible), so the changes are in improvement even without the Mr. Anderson problems that prompted them. I guess the rest is a subject for another discussion.

> But in terms of the original goal of the ticket... I think it's all here.

That you certainly did.

gonewest818:
Those ☝️ commits are the sorted and squashed ones. I ran the full test suite locally for Clojure 1.9 but let's wait for Travis to confirm all the other JVMs and versions.

gonewest818:
LGTM.

benedekfazekas:
great job @gonewest818 !!

bbatsov:
Great job indeed! @gonewest818, your contributions lately have been invaluable and are extremely appreciated! 🙇

gonewest818:
Thank you!

PR #512, opened 2018-02-28
[Fix #367] Add a "smoke test" to check the middleware

### LINKED ISSUE #367: Middleware tests would build a jar and test it
CIDER faced some downtime on May 11, due to a bug in the middleware. It affected all users, as the middleware wouldn't start at all, and required a hotfix and a clojars deployment. We need to improve the tests, so that it would build the jar and test it against some application.

I guess we should tackle this before the 0.14 release. Creating a ticket here so that we don't forget :-)

Relevant slack conversation:

```
arrdem     [17:35:34] It would be nice if the tests actually built and ran there jar so this doesn't happen again. I may take a swing at that later.
bozhidar   [17:56:52] hehe
bozhidar   [17:56:58] I wouldn’t mind this
bozhidar   [17:57:20] getting travis involved in the deployment has been on my todo for a while
agile_geek [19:47:30] circleci is nice too
bozhidar   [20:46:34] guess so
bozhidar   [20:47:08] I just use travis by default, as a few years ago it was the only ci with the explicit goal of providing trivial github integration
malabarba  [21:09:22] We could set up an instance on AWS that builds and deploys the latest master. Travis already lets you ping a webhook when all builds pass, so they'd be trivial to integrate.
malabarba  [21:10:40] And AWS instances are pretty cheap AFAIK, so it wouldn't take a noticeable bite out of the donations.
dpsutton   [21:42:33] I'd be down with sponsoring a year of that
dpsutton   [21:42:49] if you can estimate the cost I can make a contribution of that amount
```

Full conversation [here](https://clojurians-log.clojureverse.org/cider/2016-05-11.html).

gonewest818:
In #367 an issue was reported where a buggy build was released, impacting users and requiring a quick fix.  The contents of `test/smoketest` is a sub-project that produces an uberjar solely for the purpose of launching an nrepl server with cider-nrepl middleware. This check runs along with the `make test-clj` in CI.

- [x] All tests are passing


test/smoketest/project.clj
```diff
@@ -0,0 +1,11 @@
+(defproject smoketest "0.1.0-SNAPSHOT"
+  :dependencies [[org.clojure/tools.nrepl "0.2.12"]
```
bbatsov: I think the current version is 0.2.13

test/smoketest/src/smoketest/core.clj
```diff
@@ -0,0 +1,48 @@
...
+      (-> (nrepl/client transport 1000)
+          (nrepl/message {:id 2 :op :cider-version})
+          doall
+          first
+          :cider-version
+          :version-string))))
+
+(defn project-version
+  "Get project version assuming first line of project.clj contains
+  the form (def VERSION \"0.17.0-SNAPSHOT\")."
```
bbatsov: Seems like some copy/paste mistake to me. :-)

test/smoketest/src/smoketest/core.clj
```diff
@@ -0,0 +1,48 @@
...
+;; see cider-nrepl issue #447
+(defn nrepl-handler []
+  (require 'cider.nrepl)
+  (ns-resolve 'cider.nrepl 'cider-nrepl-handler))
+
+(defn nrepl-version
+  "Launch nrepl server and send :cider-version message. This confirms
+  the repl launched with the middleware."
+  []
+  (with-open [serv (server/start-server :handler (nrepl-handler))]
```
bbatsov: Perhaps starting the server shouldn't be in this function in case we decide to extend the test down the road. Also - in the beginning I didn't even see where it was being started as I didn't expect it to be in this function.

test/smoketest/src/smoketest/core.clj
gonewest818: All I meant to say is I'm parsing this line
https://github.com/clojure-emacs/cider-nrepl/blob/master/project.clj#L1

bbatsov: Ah, for some reason I thought you were doing to parse the `project.clj` for the test project. Btw, might be better to use some middleware that won't require such an extra step - e.g. you can just check the eldoc of something or the classpath or whatever.

test/smoketest/src/smoketest/core.clj
gonewest818: yes ok, I'll refactor.

bbatsov:
Something like this was happening when the nREPL server was trying to bind itself to ipv6, when only ipv4 was available or vice versa.

gonewest818:
Ah, right, I've actually seen that too.

gonewest818:
Well, that's funny. I actually opened a ticket on that: https://dev.clojure.org/jira/browse/NREPL-87

Anyway, I patched that up by binding the nrepl server to localhost and I left a comment why. Tests passed across the version matrix. I just squashed the commits to clean up history so I guess the tests will run one more time, then this can be merged.

bbatsov:
@gonewest818 Btw, now that nREPL is out of clojure-contrib, you should probably file the ticket here https://github.com/cemerick/nREPL Now nREPL even accepts PRs. 😉

gonewest818:
Yep. cemerick copied my ticket over and fixed it for the next release.

https://github.com/cemerick/nREPL/issues/20 <https://github.com/cemerick/nREPL/issues/20>

PR #517, opened 2018-03-24
Add support for various image content-types

arrdem:
nREP side of fixing clojure-emacs/cider#1510

Implements a new middleware, modeled heavily after the pretty-printing middleware, which recognizes `URI`, `File` and `Image` instances as return values, and **removes** the `:value` key from the nREPL response, replacing it with the key `:content` and a `:content-type` indicating how the content can be interpreted.

Self-criticisms of this patchset:
1. Because the pretty-printing middleware also inspects and manipulates `:value`, this middleware just uses an alternate key so the two don't interact at all. The default middleware stacks have been modified so the content-type middleware wins. This also means that CIDER can't use the value handler (or the pprint handler) for fallback behavior if there isn't a content-type handler found.
2. These content-type handlings are VERY hard-coded, and there's not an obvious route through which users can participate by adding their own transport side content handlings.


stardiviner:
I asked an question here https://github.com/clojure-emacs/cider/pull/2248#issuecomment-375955426 After found your email on Clojure mailing list.

dpsutton:
One thing i've noticed. It looks like you unconditionally take over the :transport key on the message. But this already happens for tracking state: from track_state.clj:

    (defn handle-tracker [handler msg]
      (if (ops-that-can-eval (:op msg))
        (handler (assoc msg :transport (make-transport msg)))
        (handler msg)))

Can you see if this is still being handled?

src/cider/nrepl/middleware/content_type.clj
```diff
@@ -0,0 +1,49 @@
+(ns cider.nrepl.middleware.content-type
```
bbatsov: I'd add some high-level documentation of the ns here (and maybe list the shortcomings/design decisions listed in the PR description).

src/cider/nrepl/middleware/content_type.clj
```diff
@@ -0,0 +1,49 @@
...
+  (:import clojure.tools.nrepl.transport.Transport
+           java.awt.Image
+           [java.io ByteArrayOutputStream File OutputStream]
+           java.net.URI
+           java.util.Base64
+           javax.imageio.ImageIO))
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+(defn response+content-type [{:keys [session value] :as response}]
```
bbatsov: This function would certain benefit from some documentation.

bbatsov:
The code looks good to me overall. A few more things are needed before we can go live:

* mention the new middleware in the README
* Update the changelog
* Add some test for this

bbatsov:
> Because the pretty-printing middleware also inspects and manipulates :value, this middleware just uses an alternate key so the two don't interact at all. The default middleware stacks have been modified so the content-type middleware wins. This also means that CIDER can't use the value handler (or the pprint handler) for fallback behavior if there isn't a content-type handler found.

I don't quite get the last sentence. I guess I need to read the other part of the code. :-)

> These content-type handlings are VERY hard-coded, and there's not an obvious route through which users can participate by adding their own transport side content handlings.

That's fine.

arrdem:
Basically reworked this changeset top to bottom, based on how a couple RFCs describe handling MIME content, and representing a reference to other content via MIME. I'm not completely sold on this design, because it's gonna be a little involved on the emacs side and the RFC's choice of structure is pretty different from CIDER's prior art.

The advantage of this rework is that the handling of URL-like things (files, paths, uris, urls) is greatly simplified in that they're simply MIME typed as references to their own address. The new slurp middleware can be used to request that the server attempt to load and return data from a URL which will be MIME detected appropriately and encoded for transport. This makes loading an image from a file produced at the REPL a two step dance, albeit a simpler one, on the emacs side because you have one handler which recognizes any file URL and asks that it be slurped, and a second handler which sees the slurped content and displays it which passess off to CIDER and an additional network hop the previous hard-coding of the image loading machinery.

Docs, a pair of simple tests and an off-by-defualt guard flag added.

arrdem:
Argh. Base64 coding didn't land until Java 8. This changeset is going to have to wait until I have time to deal with that, or someone else wants to swap in an alternate base64 coder.

src/cider/nrepl/middleware/slurp.clj
```diff
@@ -0,0 +1,101 @@
+(ns cider.nrepl.middleware.slurp
```
bbatsov: I think this should be an `util` namespace, as it's not really a middleware itself.

src/cider/nrepl.clj
```diff
@@ -124,6 +125,23 @@
                :optional (merge wrap-pprint-fn-optional-arguments
                                 {"pprint" "If present and non-nil, pretty-print the result of evaluation."})}}}))
 
+(def-wrapper wrap-content-type cider.nrepl.middleware.content-type/handle-content-type
+  #{"eval"}
+  {:doc "Middleware that adds `content-type` annotoations to the result of the the eval op."
```
bbatsov: There's a typo in annotations

bbatsov:
Seems that eastwood and cljfmt are unhappy now.

arrdem:
false failure due to a corrupted jar.

bbatsov:
👍

stardiviner:
Today, I tried this new feature, seems does not work on my side.
Here is what I tried.

```
user> (do (println "foo")
    (binding [*out* *err*]
      (println "bar"))
    (println "baz")
    (clojure.java.io/file "~/Pictures/Beauty/2015-05-04 070354.jpg"))
foo
bar
baz
;; => #object[java.io.File 0x2f763da "~/Pictures/Beauty/2015-05-04 070354.jpg"]
user> (clojure.java.io/file "~/Pictures/Beauty/2015-05-04 070354.jpg")
;; => #object[java.io.File 0x1aa925f6 "~/Pictures/Beauty/2015-05-04 070354.jpg"]
```
I get this example code from https://github.com/clojure-emacs/cider/issues/1510#issuecomment-373988901

arrdem:
This probably means that you have the pretty printer enabled in your defcustoms. Pretty printing needs to be off by default - pretty printing and the content type machinery have some nasty and difficult to improve interactions.

arrdem:
Please open a separate issue so we can talk about this and I can keep track of it. Overloading a PR isn't great.

src/cider/nrepl.clj
```diff
@@ -488,8 +506,11 @@
     wrap-info
     wrap-inspect
     wrap-macroexpand
+    wrap-slurp
     wrap-ns
     wrap-out
+    wrap-content-type
+    wrap-slurp
```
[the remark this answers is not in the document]
arrdem: Hummm this is almost certainly an accident @plexus. Offhand I definitely can't think of a reason to have slurp in here twice - although are some difficult interactions between the slurping machinery and the pretty-printer so there may be a subtly here I'm forgetting.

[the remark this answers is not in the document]
plexus: I'll try taking this out. Given that nrepl re-orders these middlewares based on dependencies I find it hard to imagine that it does much.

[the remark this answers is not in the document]
arrdem: IIRC the core of the problem is that while nrepl re-orders middlewares, the pprint middleware rewrites the response to elide the eval result which creates an implicit data dependency between these two middlewares.

PR #519, opened 2018-03-31
Add test selector support

benedekfazekas:
Test all and test ns understand `leiningen` style test selectors, eg
metadata markers on `deftest`s. No support for namespace level metada
markers for now.

If test selector is specified only the test(s) run that have that
selector in their metadata.

- [x] All tests are passing

### DIFF
--- a/src/cider/nrepl/middleware/test.clj
+++ b/src/cider/nrepl/middleware/test.clj
@@ -199,14 +199,16 @@
 
 (defn test-vars
   "Call `test-var` on each var, with the fixtures defined for namespace
-  object `ns`."
-  [ns vars]
+  object `ns`. Filter vars with `selector` metadata marker when `selector` is present."
+  [ns selector vars]
   (let [once-fixture-fn (test/join-fixtures (::test/once-fixtures (meta ns)))
         each-fixture-fn (test/join-fixtures (::test/each-fixtures (meta ns)))]
     (try (once-fixture-fn
           (fn []
             (doseq [v vars]
-              (when (:test (meta v))
+              (when (and (:test (meta v))
+                         (or (not selector)
+                             ((keyword selector) (meta v))))
                 (each-fixture-fn (fn [] (test-var v)))))))
          (catch Throwable e
            (report-fixture-error ns e)))))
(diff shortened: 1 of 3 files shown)


test/clj/cider/nrepl/middleware/test_selector.clj
```diff
@@ -0,0 +1,29 @@
+(ns cider.nrepl.middleware.test-selector
```
bbatsov: Do we need separate ns for the selector tests? Seems odd to me.

src/cider/nrepl/middleware/test.clj
```diff
@@ -200,13 +200,15 @@
 (defn test-vars
   "Call `test-var` on each var, with the fixtures defined for namespace
   object `ns`."
-  [ns vars]
+  [ns test-selector vars]
```
bbatsov: I guess this should be just `selector`, as it can't really be a non test selector here. :-) You'll also have to update all the docstrings to mention this.

test/clj/cider/nrepl/middleware/test_selector.clj
benedekfazekas: happy to merge into `cider.nrepl.middleware.test-test`. The only reason to keep it separate that it is more integration like test eg. it needs `session/session-fixture` and sends `nREPL` messages while the `....test-test` ns is much simpler.

test/clj/cider/nrepl/middleware/test_selector.clj
```diff
@@ -0,0 +1,29 @@
...
+  (:require [cider.nrepl.test-session :as session]
+            [clojure.test :refer :all]))
+
+(use-fixtures :each session/session-fixture)
+
+(deftest only-smoke-test-run-test
+  (testing "only test marked as smoke is run when test-all is used"
```
bbatsov: The key alignment in all tests seems off. Maybe you used tabs or something?

test/clj/cider/nrepl/middleware/test_selector.clj
bbatsov: Hmm, I didn't notice it had virtually no tests. That's odd. I guess it can certainly use a few. In general I think that most of the middleware ns tests should be integration tests anyways.

bbatsov:
👍

PR #520, opened 2018-04-01
Fix test.extensions =-body: Only evaluate first argument once

niklasfasching:
`(is (= ...))` tests executed in cider currently fail if the first
argument to = is not pure as it is evaluated multiple times inside
`=-body`. By caching the value, this can be prevented.

This fixes clojure-emacs/cider/issues/2233

Example:
```
;; before
cider> (let [x (atom 0)]
         (is (= (swap! x inc) 1))
         @x)
;; => 2

;; after
cider> (let [x (atom 0)]
         (is (= (swap! x inc) 1))
         @x)
;; => 1
```

----

- [x] All tests are passing


xiongtx:
👏

niklasfasching:
sorry for the empty README line commit :see_no_evil: - should i open a PR to revert it or what do we do with this :D?

xiongtx:
Don't worry about it :wink:.

PR #522, opened 2018-04-16
Extend selectors to include/exclude lists

### LINKED ISSUE #521: Extended selector support
Compared with https://github.com/cognitect-labs/test-runner our selector support is basic. It would be great to support:

- multiple inclusions
- exclusions
- multiple exclusions

Matching the behaviour of test-runner is probably wise.

SevereOverfl0w:
Fix #521

### DIFF
--- a/src/cider/nrepl/middleware/test.clj
+++ b/src/cider/nrepl/middleware/test.clj
@@ -197,18 +197,34 @@
                               :message "Uncaught exception, not in assertion"})))
       (test/do-report {:type :end-test-var :var v}))))
 
+(defn- var-filter
+  [var include exclude]
+  (let [test-inclusion (if include
+                         #((apply some-fn include) (meta %))
+                         (constantly true))
+        test-exclusion (if exclude
+                         #((complement (apply some-fn exclude)) (meta %))
+                         (constantly true))]
+    (and (test-inclusion var)
+         (test-exclusion var))))
+
 (defn test-vars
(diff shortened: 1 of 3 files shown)


src/cider/nrepl/middleware/test.clj
```diff
@@ -197,18 +197,30 @@
...
+                         #((complement (apply some-fn exclude)) (meta %))
+                         (constantly true))]
+    (and (test-inclusion var)
+         (test-exclusion var))))
+
 (defn test-vars
   "Call `test-var` on each var, with the fixtures defined for namespace
   object `ns`. Filter vars with `selector` metadata marker when `selector` is present."
-  [ns selector vars]
+  [ns include exclude vars]
```
bbatsov: You forgot to update the docstring. I guess it should mention that those should be seqs.

src/cider/nrepl/middleware/test.clj
```diff
@@ -218,13 +230,13 @@
   Otherwise, test the specified vars. If no vars are specified, test all vars
   in the namespace. On completion, return a map of test results. If `selector`
   is present filter vars that have the `selector` metadata marker."
-  [ns selector vars]
+  [ns include exclude vars]
```
bbatsov: Same here.

bbatsov:
I've updated CIDER to support the new interface, so people can start playing with this. //cc @benedekfazekas

PR #526, opened 2018-05-06
Use orchard.query for apropos and tests

SevereOverfl0w:


src/cider/nrepl/middleware/util/nrepl.clj
```diff
@@ -19,3 +19,41 @@
                              :status :notification
                              :msg notification
                              (when type [:type type])))))
+
+(defn- update-some
```
bbatsov: I think you meant to put all of those in some of the other util namespaces. Those here are special middleware utils that result in sending messages to the clients.

src/cider/nrepl/middleware/apropos.clj
```diff
@@ -2,12 +2,36 @@
...
 
 ;;; ## Middleware
 
 (defn apropos [msg]
-  {:apropos-matches (apropos/find-symbols msg)})
+  {:apropos-matches
+   (apropos/find-symbols
+    (cond-> msg
+
+      (:privates? msg)
```
bbatsov: You might want to add some notes about compatibility code that will be removed down the road. Otherwise we'll likely forget to do this. :-)

src/cider/nrepl/middleware/util/nrepl.clj
SevereOverfl0w: I've added a util.coerce namespace for this. The `transform-value` stuff got moved into orchard, and I think it should be moved back eventually, perhaps even for this purpose.

bbatsov: Yeah, probably. I think it doesn't make much sense outside its usage with nREPL.

CHANGELOG.md
```diff
@@ -6,9 +6,13 @@
 
 * Extracted part of the nREPL-agnostic functionality to `orchard`.
 * Added a profiling middleware.
+* Support for orchard var-query in apropos
+* Support for orchard var-query in test
```
bbatsov: I guess here it should also be mentioned the new op `test-var-query`. We also list those in the README.

src/cider/nrepl/middleware/test.clj
```diff
@@ -300,29 +272,42 @@
...
-      (let [nss {ns (map u/as-sym tests)}
-            report (test-nss nss include exclude)]
+(defn handle-test-var-query-op
+  [{:keys [var-query transport] :as msg}]
+  (with-interruptible-eval
+    msg
+    (try
+      (let [report (test-var-query
+                    (-> var-query
+                         ;; TODO: Should this be done, or should the client specify it?
```
SevereOverfl0w: @bbatsov is this TODO a yes or no? Happy to remove it / act upon it.

bbatsov: Why would the client want to run tests in a ns without tests?

SevereOverfl0w: I don't think it would make sense to, but it wouldn't cause errors either. I can't see any harm in this right now. I shall remove the TODO.

PR #528, opened 2018-05-07
Accept non-namespaced tests in backwards compat

SevereOverfl0w:
This is the original API before the API was deprecated.


SevereOverfl0w:
I have no idea what codecov/patch is. I'm guessing it's because I haven't tested my changes.

SevereOverfl0w:
Added a regression test to cover the test op for this kind of usage too.

PR #532, opened 2018-05-23
Add a simple nrepl-server task in cider.tasks

arichiardi:
First step towards the new cider nrepl integration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,7 +155,7 @@ it on the command line through the `cider.tasks/add-middleware` task
 functionality):
 
 ```
-boot -d org.clojure/tools.nrepl:0.2.12 -d cider/cider-nrepl:0.x.y-SNAPSHOT -i "(require 'cider.tasks)" cider.tasks/add-middleware -m cider.nrepl.middleware.apropos/wrap-apropos -m cider.nrepl.middleware.version/wrap-version repl -s wait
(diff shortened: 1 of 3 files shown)


bbatsov:
That looks simple enough!

arichiardi:
If I understand correctly - if this is what we want for `boot` - we basically can handle the `cider-boot-parameters` in cider directly.

If `cider-jack-in-command` sees that  the project depends on:

 * [org.clojure/tools.nrepl "0.2.13"] -> `"repl -s -H :: wait"` (current behavior)
 * [nrepl "0.3.1"] -> `"repl -s -H :: wait"` (current behavior)
 * [nrepl "0.4.0"] -> `"nrepl-server -s -H :: wait"`

Problem is `cider-boot-parameters` is a `defcustom` so we probably need to take away the `repl` part because it suddenly becomes dynamic.

Another approach would be to always call `nrepl-server`:

```
(defcustom cider-boot-parameters
  "nrepl-server -s -H :: wait"
  "Params passed to boot to start an nREPL server via `cider-jack-in'."
  :type 'string
  :group 'cider
  :safe #'stringp
  :package-version '(cider . "0.9.0"))
```

And detect the `nrepl` dependency in the task.

bbatsov:
CIDER won't be able to work with both versions of nREPL anyways, so there's no need to complicate this. I'm ok with the second option.

arichiardi:
Ok so the task accepts `-p` and `-b` like boot's task so we could do this:

`nrepl-server -b :: wait`

The `-b` to localhost is the default too.

arichiardi:
I also made the whole thing a bit more robust - we should probably merge https://github.com/nrepl/nREPL/pull/34 before this just to make sure things are ok.

arichiardi:
The error in the tests is unrelated to the change. It is a zip error.

bbatsov:
Nice!

Now we just need the same time for lein and then we can update the the cider-jack-in commands and finally switch to our new nREPL.

PR #535, opened 2018-07-01
Check for cemerick/piggieback and fall back to cider/piggieback

dpsutton:
CIDER injects cider/piggieback automatically now while some users are
still on cemerick/piggieback. The result is that this will always
return true since cider/piggieback is on the classpath.

Notably, shadow-cljs fakes a cemerick/piggieback and places the
compiler environment there for cider-nrepl. But we see the injected
cider/piggieback and so look for the cljs compiler environment there
rather than in cemerick.

see https://github.com/clojure-emacs/cider/issues/2305

### DIFF
--- a/src/cider/nrepl/middleware/util/cljs.clj
+++ b/src/cider/nrepl/middleware/util/cljs.clj
@@ -41,9 +41,9 @@
   "Returns the path in the session map for the ClojureScript compiler
   environment used by piggieback."
   []
-  [(if cider-piggieback?
-     (resolve 'cider.piggieback/*cljs-compiler-env*)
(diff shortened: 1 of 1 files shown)


bbatsov:
I'll merge this, but @thheller should change it soon, as once I switch the middlewares to target the new nREPL server `cemerick/piggieback` will no longer work at all, and leaving this check would make things confusing, as no one would imagine it's here for `shadow-cljs`'s sake.

thheller:
Sorry about that. Totally forgot about this. Just pushed `shadow-cljs@2.4.19` which should properly support both or either namespace.

bbatsov:
Thanks for tackling this!

Issue #537, opened 2018-07-10
latest 0.18.0-SNAPSHOT broken:  No matching ctor found for class java.net.InetSocketAddress


hagmonk:
Looks like a new version of 0.18.0-SNAPSHOT was pushed today that breaks cider-nrepl. 0.17.0 isn't affected, nor was 0.18.0-SNAPSHOT prior to today.

```
java -version
java -version
java version "1.8.0_172"
Java(TM) SE Runtime Environment (build 1.8.0_172-b11)
Java HotSpot(TM) 64-Bit Server VM (build 25.172-b11, mixed mode)
```

```
/usr/local/bin/clojure -Sdeps '{:deps {org.clojure/tools.nrepl {:mvn/version "0.2.13"} refactor-nrepl {:mvn/version "2.4.0-SNAPSHOT"} cider/cider-nrepl {:mvn/version "0.18.0-SNAPSHOT"}}}'  -e '(require (quote cider-nrepl.main)) (cider-nrepl.main/init ["refactor-nrepl.middleware/wrap-refactor", "cider.nrepl/cider-middleware"])'
Exception in thread "main" java.lang.IllegalArgumentException: No matching ctor found for class java.net.InetSocketAddress
        at clojure.lang.Reflector.invokeConstructor(Reflector.java:183)
        at clojure.tools.nrepl.server$start_server.invokeStatic(server.clj:141)
        at clojure.tools.nrepl.server$start_server.doInvoke(server.clj:121)
        at clojure.lang.RestFn.invoke(RestFn.java:512)
        at cider_nrepl.main$start_nrepl.invokeStatic(main.clj:39)
        at cider_nrepl.main$start_nrepl.invoke(main.clj:31)
        at cider_nrepl.main$init.invokeStatic(main.clj:55)
        at cider_nrepl.main$init.invoke(main.clj:51)
        at user$eval1096.invokeStatic(NO_SOURCE_FILE:1)
        at user$eval1096.invoke(NO_SOURCE_FILE:1)
        at clojure.lang.Compiler.eval(Compiler.java:7062)
        at clojure.lang.Compiler.eval(Compiler.java:7025)
        at clojure.core$eval.invokeStatic(core.clj:3206)
        at clojure.main$eval_opt.invokeStatic(main.clj:291)
        at clojure.main$eval_opt.invoke(main.clj:285)
        at clojure.main$initialize.invokeStatic(main.clj:311)
[...]

PR #538, opened 2018-07-10
Proposed fix for #537

hagmonk:
- revert `cider.nrepl/init` such that it takes a vector of middlewares
- have `cider.nrepl/init` pass a map of arguments to `start-nrepl`
- tweak cider nrepl boot task to pass a map to `start-nrepl` directly
- set defaults for `port` and `bind` to keep `nrepl-server` happy

### DIFF
--- a/src/cider/tasks.clj
+++ b/src/cider/tasks.clj
@@ -30,6 +30,6 @@
     (util/dbug* "nREPL middleware: %s\n" (vec default-mws))
     (boot.core/with-pass-thru [_]
       (require 'cider-nrepl.main)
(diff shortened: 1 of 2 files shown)


hagmonk:
@arichiardi I changed the boot task in `src/cider/tasks.clj` to invoke `cider.nrepl/start-nrepl` - sorry if I'm missing something else here, I'm not a boot user

hagmonk:
no worries @arichiardi  👍

arichiardi:
Left a comment in the issue, probably the right fix is to update cider's side.

PR #539, opened 2018-07-11
Actually fix #537

### LINKED ISSUE #537: latest 0.18.0-SNAPSHOT broken:  No matching ctor found for class java.net.InetSocketAddress

hagmonk:
My fix yesterday was not properly tested - I started nREPL from the command line, but I didn't attempt to exercise it with messages. As a result, I missed the fact I was setting passing the middleware as the handler, which kept things just as broken.

This PR contains a fix for that problem, plus a new test ns `cider.nrepl.main-test` that exercises the `start-nrepl` function by passing it a variety of valid and invalid option maps. The tests go as far as checking that two `:op "clone"` messages can be exchanged before declaring success.

On my local machine I saw a test failure in a totally unrelated part of the code, `test-sum-types-is-base64`, which I'm assuming is an artifact of my local environment and not something that will come up in the PR tests.

### DIFF
--- a/src/cider_nrepl/main.clj
+++ b/src/cider_nrepl/main.clj
@@ -56,7 +56,7 @@
 
         {:keys [server-socket port] :as server}
         (nrepl.server/start-server :handler handler
-                                   :bind (or bind "::")
+                                   :bind (or bind "localhost")
                                    :port (or port 0))
 
         bind
--- a/src/cider_nrepl/main.clj
+++ b/src/cider_nrepl/main.clj
@@ -53,7 +53,7 @@
   [{:keys [bind port] :as opts}]
   (let [handler
         (build-handler opts)
-        
+
         {:keys [server-socket port] :as server}
         (nrepl.server/start-server :handler handler
                                    :bind (or bind "::")
--- a/src/cider_nrepl/main.clj
+++ b/src/cider_nrepl/main.clj
@@ -29,11 +29,8 @@
   (into [] mw-xf middleware-var-strs))
 
 (defn- build-handler
-  [{:keys [handler middleware]}]
-  (let [handler (or handler nrepl.server/default-handler)]
-    (if middleware
-      (apply handler (->mw-list middleware))
-      (handler))))
+  [middleware]
+  (apply nrepl.server/default-handler (->mw-list middleware)))
 
 (defn start-nrepl
   "Starts a socket-based nREPL server. Accepts a map with the following keys:
@@ -50,9 +47,11 @@
      representing middleware you wish to mix in to the nREPL handler. Vars can
      resolve to a sequence of vars, in which case they'll be flattened into the
      list of middleware."
-  [{:keys [bind port] :as opts}]
+  [{:keys [handler middleware bind port] :as opts}]
   (let [handler
-        (build-handler opts)
+        (if handler
+          (handler)
+          (build-handler middleware))
 
         {:keys [server-socket port] :as server}
         (nrepl.server/start-server :handler handler
--- a/src/cider_nrepl/main.clj
+++ b/src/cider_nrepl/main.clj
@@ -28,17 +28,36 @@
   [middleware-var-strs]
   (into [] mw-xf middleware-var-strs))
 
+(defn- build-handler
+  [{:keys [handler middleware]}]
+  (let [handler (or handler nrepl.server/default-handler)]
+    (if middleware
+      (apply handler (->mw-list middleware))
+      (handler))))
+
 (defn start-nrepl
-  [opts]
-  (let [{:keys [handler middleware bind port] :or {port 0 bind "::"}} opts
+  "Starts a socket-based nREPL server. Accepts a map with the following keys:
+ 
+   * :port — defaults to 0, which autoselects an open port
 
-        handler (cond-> (or handler nrepl.server/default-handler)
-                  middleware (apply (->mw-list middleware)))
+   * :bind — bind address, by default \"::\" (falling back to \"localhost\" if
+     \"::\" isn't resolved by the underlying network stack)
 
+   * :handler — the nREPL message handler to use for each incoming connection;
+     defaults to the result of `(nrepl.server/default-handler)`
+
+   * :middleware - a sequence of vars or string which can be resolved to vars,
+     representing middleware you wish to mix in to the nREPL handler. Vars can
+     resolve to a sequence of vars, in which case they'll be flattened into the
+     list of middleware."
+  [{:keys [bind port] :as opts}]
+  (let [handler
+        (build-handler opts)
+        
         {:keys [server-socket port] :as server}
         (nrepl.server/start-server :handler handler
-                                   :bind bind
-                                   :port port)
+                                   :bind (or bind "::")
+                                   :port (or port 0))
 
         bind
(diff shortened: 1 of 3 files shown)


test/clj/cider/nrepl/main_test.clj
```diff
@@ -0,0 +1,43 @@
+(ns cider.nrepl.main-test
+  (:require [cider.nrepl :refer [wrap-debug cider-middleware]]
+            [cider-nrepl.main :as m]
+            [clojure.test :refer :all]
+            [clojure.tools.nrepl :as nrepl]
+            [clojure.tools.nrepl.server :as nrepl.server]
+            [clojure.tools.nrepl.transport :as transport]))
+
+(defn start-stop-nrepl-session [opts]
```
bbatsov: I don't see where you stop the server in each test.

src/cider_nrepl/main.clj
```diff
@@ -28,17 +28,36 @@
   [middleware-var-strs]
   (into [] mw-xf middleware-var-strs))
 
+(defn- build-handler
```
bbatsov: I think that build handler should just take a list of middlewares and be invoked only if a handler wasn't passed to `start-nrepl`.

src/cider_nrepl/main.clj
```diff
@@ -28,17 +28,36 @@
...
 
-        handler (cond-> (or handler nrepl.server/default-handler)
-                  middleware (apply (->mw-list middleware)))
+   * :middleware - a sequence of vars or string which can be resolved to vars,
+     representing middleware you wish to mix in to the nREPL handler. Vars can
+     resolve to a sequence of vars, in which case they'll be flattened into the
+     list of middleware."
+  [{:keys [bind port] :as opts}]
+  (let [handler
+        (build-handler opts)
```
bbatsov: Here I'd `(or handler (build-handler)`. (provided it's extracted from the `ops` map above.

src/cider_nrepl/main.clj
```diff
@@ -28,17 +28,36 @@
...
 
         {:keys [server-socket port] :as server}
         (nrepl.server/start-server :handler handler
-                                   :bind bind
-                                   :port port)
+                                   :bind (or bind "localhost")
```
bbatsov: Doesn't bind work with `nil`?

test/clj/cider/nrepl/main_test.clj
hagmonk: My understanding was that the `with-open` macro would suffice to close the server. I borrowed much of this from `cider.nrepl.middleware.debug-integration-test/with-nrepl-session*`

src/cider_nrepl/main.clj
hagmonk: I see, so that implies that either a handler or middleware is expected, but not both. I can make that change.

src/cider_nrepl/main.clj
hagmonk: So, I think what will happen here is an exception when this goes through travis. From `clojure.tools.nrepl.server`:

```
 (let [bind-addr (if bind
                    (InetSocketAddress. ^String bind ^Integer port)
                    (let [local (InetSocketAddress. "::" port)]
                      (if (.isUnresolved local)
                        (InetSocketAddress. "localhost" port)
                        local)))
```

`nil` causes `(InetSocketAddress. "::" port)` to be invoked, which throws an exception in travis - I presume due to IPv6 not being enabled. If we default to localhost, we dodge that particular idiosyncrasy.

hagmonk: Sorry I should clarify: I made this change after experiencing the exception in travis. Not a thought experiment, but an observation :)

test/clj/cider/nrepl/main_test.clj
bbatsov: Ah, ok.

src/cider_nrepl/main.clj
bbatsov: Yeah, exactly. Normally you'd just use a predefined handler, but in some cases you might opt to build the handler yourself by supplying some list of middlewares instead.

bbatsov: (e.g. because some middleware is problematic in your setup or something like this)

src/cider_nrepl/main.clj
bbatsov: I see. I was just wondering if this won't message up something for people who have disabled IPv4. Probably not a big deal now, I'll fixed this once we finally move the project to https://github.com/nrepl/nREPL

bbatsov: (tools.nrepl is practically dead, so we've migrated the project out of clojure contrib to revive it)

PR #544, opened 2018-09-16
Bugfix: Debugging deftype Forms

cartesian-theatrics:
This is likely not mergable code as is, but hopefully gets the idea across. See https://github.com/clojure-emacs/cider/issues/2453 for more details.

### DIFF
--- a/src/cider/nrepl/middleware/debug.clj
+++ b/src/cider/nrepl/middleware/debug.clj
@@ -451,9 +451,11 @@ this map (identified by a key), and will `dissoc` it afterwards."}
 
 (def ^:dynamic *tmp-forms* (atom {}))
 (def ^:dynamic *do-locals* true)
+(def ^:dynamic *STATE__* nil)
+
 
 (defmacro with-initial-debug-bindings
-  "Let-wrap `body` with STATE__ map containing code, file, line, column etc.
+  "Let-wrap `body` with *STATE__* map containing code, file, line, column etc.
   STATE__ is an anaphoric variable available to all breakpoint macros. Ends with
   __ to avid conflicts with user locals and to signify that it's an internal
   variable which is cleaned in `sanitize-env' along other clojure's
@@ -461,20 +463,24 @@ this map (identified by a key), and will `dissoc` it afterwards."}
   {:style/indent 1}
   [& body]
   ;; NOTE: *msg* is the message that instrumented the function,
-  `(let [~'STATE__ {:msg ~(let [{:keys [code id file line column ns]} *msg*]
-                            (-> {:code code
-                                 ;; Passing clojure.lang.Namespace object
-                                 ;; as :original-ns breaks nREPL in bewildering
-                                 ;; ways.
-                                 :original-id id, :original-ns (str (or ns *ns*))
-                                 :file file, :line line, :column column}
-                                ;; There's an nrepl bug where the column starts counting
-                                ;; at 1 if it's after the first line. Since this is a
-                                ;; top-level sexp, a (= col 1) is much more likely to be
-                                ;; wrong than right.
-                                (update :column #(if (= % 1) 0 %))))
-                    :forms @*tmp-forms*}]
-     ~@body))
+  `(do
+     ;; Setting a root var because deftype does not respect locals.
+     (def *STATE__* {:msg ~(let [{:keys [code id file line column ns]} *msg*]
+                             (-> {:code code
+                                  ;; Passing clojure.lang.Namespace object
+                                  ;; as :original-ns breaks nREPL in bewildering
+                                  ;; ways.
+                                  :original-id id, :original-ns (str (or ns *ns*))
+                                  :file file, :line line, :column column}
+                                 ;; There's an nrepl bug where the column starts counting
+                                 ;; at 1 if it's after the first line. Since this is a
+                                 ;; top-level sexp, a (= col 1) is much more likely to be
+                                 ;; wrong than right.
+                                 (update :column #(if (= % 1) 0 %))))
+                     :forms @*tmp-forms*})
+     ~@body
+     (def *STATE__* nil)))
+
 
 (defn break
   "Breakpoint function.
@@ -509,14 +515,14 @@ this map (identified by a key), and will `dissoc` it afterwards."}
   [form {:keys [coor]} original-form]
   (let [val-form (if (looks-step-innable? form)
                    (let [[fn-sym & args] form]
-                     `(apply-instrumented-maybe (var ~fn-sym) [~@args] ~coor ~'STATE__))
+                     `(apply-instrumented-maybe (var ~fn-sym) [~@args] ~coor ~*STATE__*))
                    form)
         locals (when *do-locals*
                  (sanitize-env &env))]
     ;; Keep original forms in a separate atom to save some code
     ;; size. Unfortunately same trick wouldn't work for locals.
     (swap! *tmp-forms* assoc coor original-form)
-    `(break ~coor ~val-form ~locals ~'STATE__)))
+    `(break ~coor ~val-form ~locals ~*STATE__*)))
 
 (def irrelevant-return-value-forms
   "Set of special-forms whose return value we don't care about.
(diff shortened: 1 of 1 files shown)


bbatsov:
I haven't done much work on this part of the code, so I'll defer to @Malabarba, @vspinu and @grammati.

Malabarba:
I agree that this will probably break things. 

I'm also not sure if the added maintenance overhead of a proper solution would be worth the feature.
But I'm not the one who's being maintaining this code the most lately, so I'll let the others decide that.

cartesian-theatrics:
Yeah now that I think of it, I don't think this would even work if you instrumented more than one form at a time, so this definitely is not a workable solution.

An alternative approach is to modify the calls in the instrumentation code (e.g. ins/tag-form-recursively) to take the state map and flow it through, which would resolve the garbage issue at the cost of more change. I haven't yet investigated the details on any of this.

P.S. This is an awesome debugger! It has saved me tons of time, so thanks for all your work.

Malabarba:
> P.S. This is an awesome debugger! It has saved me tons of time, so thanks for all your work.

I'm glad it helped you. :-)

cartesian-theatrics:
@Malabarba @vspinu , this is so far the only approach that isn't turning into a big quagmire for me. Could something like this be made acceptable?

vspinu:
This might work actually. Does this produce nested state wraps?

vspinu:
I agree with @Malabarba that if deftype is the only form benefiting from this it's probably not worth it unless it's a really simple intervention. 

A simple solution would be to check if the top form is `deftype*` and do the wrap on each suitable form at level 1, along the lines that you proposed. 

An alternative solution would be to make `#dbg` and `#break` wrap state at the form where they are located. Then you could do something like:

```clj
(deftype Temp [a] 
  clojure.lang.IDeref
  (deref [this]
    #break
    (let [b 1]
      (+ a b))))
```

I am somewhat surprised that it's already not the the case. In the current implementation several    `#break`points will share the same state which surely results in unexpected `continue` behavior. There are a few debugger issues open touching on that. I think it would be wise to tackle all of these simultaneously. 

In any case, would you mind trying this approach? Basically moving STATE wrapping directly into `breakpoint-reader` and `debug-reader` and remove it from the `instrument-and-eval`.

cartesian-theatrics:
@vspinu , Yeah this only walks one level down. To me not supporting defrecord or deftype is a pretty critical short-cumming, so hopefully we can come to some solution.

It does look like continue might be problematic, but I haven't looked closely. Stepping is fine, though, as it wraps the next form in a new state before evaling it. So it looks like we aught to be able to generate all the state bindings ahead of time, which might actually be a cleaner solution overall. I'll give it a go when I get the chance -- probably this weekend.

vspinu:
> An alternative solution would be to make #dbg and #break wrap state at the form where they are located. 

I have looked at it myself and I think this is the way to go irrespective of this issue:

```diff
1 file changed, 12 insertions(+), 4 deletions(-)
src/cider/nrepl/middleware/debug.clj | 16 ++++++++++++----

modified   src/cider/nrepl/middleware/debug.clj
@@ -476,6 +476,13 @@ this map (identified by a key), and will `dissoc` it afterwards."}
                     :forms @*tmp-forms*}]
      ~@body))
 
+(defmacro breakpoint-with-initial-debug-bindings
+  {:style/indent 1}
+  [form dbg-state original-form]
+  `(with-initial-debug-bindings
+     (breakpoint-if-interesting
+       ~form ~dbg-state ~original-form)))
+
 (defn break
   "Breakpoint function.
   Send the result of form and its coordinates to the client and wait for
@@ -550,23 +557,24 @@ this map (identified by a key), and will `dissoc` it afterwards."}
              (expand-break ~form ~dbg-state ~original-form)
              (finally (reset! *skip-breaks* old-breaks#))))
         `(expand-break ~form ~dbg-state ~original-form)))))
+
 ;;; ## Data readers
 ;;
 ;; Set in `src/data_readers.clj`.
 (defn breakpoint-reader
   "#break reader. Mark `form` for breakpointing."
   [form]
-  (ins/tag-form form #'breakpoint-if-interesting))
+  (ins/tag-form form #'breakpoint-with-initial-debug-bindings))
 
 (defn debug-reader
   "#dbg reader. Mark all forms in `form` for breakpointing.
   `form` itself is also marked."
   [form]
-  (ins/tag-form-recursively form #'breakpoint-if-interesting))
+  (ins/tag-form (ins/tag-form-recursively form #'breakpoint-if-interesting)
+                #'breakpoint-with-initial-debug-bindings))
 
 (defn instrument-and-eval [form]
-  (let [form1 `(with-initial-debug-bindings
-                 ~(ins/instrument-tagged-code form))]
+  (let [form1 (ins/instrument-tagged-code form)]
     ;; (ins/print-form form1 true false)
     (try
       (binding [*tmp-forms* (atom {})]
```

I pushed this to `localized-debugger` branch. Unfortunately I will be able to look deeper only in the second half of October. For now let's sleep a bit on it. Would appreciate if you could play further on with it.

The following example works as expected:

```clj
(defprotocol MyInterface 
  (foo [this]))

(deftype Temp [a]
  MyInterface
  (foo [this]
    #dbg
    (let [b 1]
      (+ a b))))
(def temp (->Temp 3))
(foo temp)
```

But something strange happens with your deref example, or maybe I just did it wrongly. The debugger gets stuck in the last form:

```clj
(deftype Temp [a] 
  clojure.lang.IDeref
  (deref [this]
    #break
    (let [b 1]
      (+ a b))))
(def temp (->Temp 3))
@temp
```

cartesian-theatrics:
Looks great, I was just implementing something similar.  Only issue there is it looks like it'll wrap it even if it's not interesting? 

I think the last issue is something completely different. There's major weirdness with the built-in protocols. In this example, `IMeta` throws immediately, ILookup works fine, IFoo works fine.

```clojure
(deftype Temp []
  clojure.lang.IMeta
  (meta [this]
    (into {} (map vector (range 10) (range 10))))
  clojure.lang.Associative
  clojure.lang.ILookup
  (valAt [this k]
    (map list (range 10)))
  (valAt [this k not-found]
    (map list (range 10)))
  IFoo
  (foo [this]
    (map list (range 10))))
```

Maybe it has something to do with the type declarations on the Interfaces when you are implementing Interfaces? But then IMeta just returns an object, so idk what's going on there. This might be a major monkey wrench :(

cartesian-theatrics:
Okay, looks like instrumenting :static is at least one of the problems:
```clojure
(defprotocol IFoo (^:static foo [this]))
(deftype Temp []
  IFoo
  (foo [this]
    (into {} (map vector (range 10) (range 10)))))
(.foo (Temp.))
```

This will cause a stack overflow. looks like we may have to use reflection to determine if a method can be instrumented.

vspinu:
> Only issue there is it looks like it'll wrap it even if it's not interesting?

It's harmless and is not worth the code complication unless a meaningful message is conveyed to the user. Silently not-instrumenting or instrumenting and silently not running is the same thing. In any case it's not a big deal, it's up to the users to put debug readers in right place. 

> reflection to determine if a method can be instrumented.

I don't think so. The current code is oblivious of what form is instrumented and that is elegant and easy to maintain. To be able to deal with peculiarities of deftpe methods would require a pile of task specific hacks. We eventually want to make that debugger work in clojurescript, so any such complications  would be going in wrong direction. 

> This will cause a stack overflow

Works if you place an explicit reader:

```clj
(defprotocol IFoo (^:static foo [this]))
(deftype Temp []
  IFoo
  (foo [this]
    #dbg
    (into {} (map vector (range 10) (range 10)))))
(.foo (Temp.))
```

To be clear, I don't think trying to make `#dbg` work around the whole `deftype` is a worthwhile initiative. Users should be informed that `C-u C-M-x` is equivalent to adding a `#dbg` in front of the form and I think most of the users do already  know that. Thus, `C-u C-M-x` will fail on all form where `#dbg` will fail, like deftype or "non-interesting" forms. Instead, users should place `#dbg` in specific locations like in the example above. In multi method deftypes, it's unlikely that someone would be interested in debugging all methods of deftype at once, only some specific method, For this reason the localized debugger fix is a very good tradeoff between simplicity of the code-base and richness of the UI.

cartesian-theatrics:
That's fair, I totally agree. Just the the patch you provided I think would be great.

It might make sense at some point to write a spec for forms that are instrummentable just for the error message.

cartesian-theatrics:
@vspinu , what do you think about merging your patch?

vspinu:
Let me have a look at other debugger issues before the final say. Will try to have a look over the weekend.

PR #546, opened 2018-09-25
Add compatibility with matcher-combinators

plexus:
Currently when running a test through cider, any failures in a
`(is (match? ,,,))` clause will be silently ignores.

This detects mismatch messages and treats them as failure messages. Where there
is a mismatch the user will see a {:expected x :actual y} in the test output.


plexus:
This is a logical follow-up to https://github.com/clojure-emacs/cider-nrepl/pull/510, where it was mentioned that matcher-combinators should use a namespaced keyword for its custom message type, which is does since https://github.com/nubank/matcher-combinators/pull/34.

bbatsov:
@plexus LGTM.

Can you please mention this in the changelog as well? 

Maybe also add that bit of documentation that was discussed in #510 as well for future reference.

plexus:
I've updated the CHANGELOG. Which documentation are you refering to @bbatsov? There's already a [bit about extensibility](https://github.com/clojure-emacs/cider/blob/master/doc/running_tests.md#using-cider-test-with-alternative-test-libraries) in the docs. Or do you mean [this exchange](https://github.com/clojure-emacs/cider-nrepl/pull/510#discussion_r170837027)? I'm not sure what the conclusion from that is...

bbatsov:
@plexus I was referring to:

```
@xiongtx I'm guessing we should add some integration notes in CIDER's manual, otherwise people are going to have a hard time figuring out what they are supposed to do.
```

(meaning documenting how to extend `report`, but that's not a big deal)

PR #551, opened 2018-10-12
[Fix #549] Fix spec error messages for Clojure 1.10

### LINKED ISSUE #549: Clojure 1.10 breaks stacktrace handling reporting
Using Cider with Clojure-1.10 gives:
```
ardoq.api.survey-api> (let [2])
Syntax error macroexpanding clojure.core/let at (*cider-repl ardoq.com/ardoq-api:localhost:53859(clj)*:7015:23).
[2] - failed: even-number-of-forms? at: [:bindings] spec: :clojure.core.specs.alpha/bindings
14715723 [clojure-agent-send-off-pool-28] ERROR []  clojure.tools.nrepl.server - Unhandled REPL handler exception processing message {:op stacktrace, :pprint-fn clojure.pprint/pprint, :print-length 50, :print-level 50, :session 8442d3d0-9a9d-43d6-8693-bdde270bd4e8, :id 237} 
java.lang.NullPointerException: null
	at clojure.string$replace_first.invokeStatic(string.clj:165) ~[clojure-1.10.0-RC1.jar:na]
	at clojure.string$replace_first.invoke(string.clj:138) ~[clojure-1.10.0-RC1.jar:na]
	at cider.nrepl.middleware.stacktrace$relative_path.invokeStatic(stacktrace.clj:204) ~[na:na]
	at cider.nrepl.middleware.stacktrace$relative_path.invoke(stacktrace.clj:198) ~[na:na]
	at cider.nrepl.middleware.stacktrace$extract_location.invokeStatic(stacktrace.clj:219) ~[na:na]
```
According to @puredanger

> Before you would have gotten the spec ExceptionInfo. Now you should get a CompilerException 
> wrapping that ExceptionInfo. It looks like the error is in parsing the location out of the message
> (which is probably missing now as that's built later). The top-level CompilerException now 
> has an ex-data that reports the line number directly so no parsing is needed.

> So if the code first looked for ex-data with source/line/col info then fell back to string parsing, 
> that would work for both:

```
user=> (let [2])
Syntax error macroexpanding clojure.core/let at (1:1). Cause: Call to clojure.core/let did not conform to spec.
[2] - failed: even-number-of-forms? at: [:bindings] spec: :clojure.core.specs.alpha/bindings
user=> (ex-data *e)
#:clojure.error{:line 1, :column 1, :phase :macroexpand, :source "NO_SOURCE_PATH", :symbol clojure.core/let}
```

^^ can just use `(-> *e ex-data :clojure.error/line)` now to grab the line, etc

Which would be a change somewhere around
https://github.com/clojure-emacs/cider-nrepl/blob/master/src/cider/nrepl/middleware/stacktrace.clj#L211

slipset:
This is IMO the simplest fix that can possibly work. Please let me know if there are ways to improve upon this.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -1,4 +1,4 @@
-(def VERSION "0.19.0-SNAPSHOT")
+  (def VERSION "0.19.0-SNAPSHOT")
 
 (defproject cider/cider-nrepl VERSION
   :description "nREPL middlewares for CIDER"
--- a/src/cider/nrepl/middleware/stacktrace.clj
+++ b/src/cider/nrepl/middleware/stacktrace.clj
@@ -205,20 +205,30 @@
 
 (defn extract-location
   "If the cause is a compiler exception, extract the useful location information
-  from its message. Include relative path for simpler reporting."
-  [{:keys [class message] :as cause}]
+  from its message or from `:location` if provided.
+  Include relative path for simpler reporting."
+  [{:keys [class message location] :as cause}]
   (if (= class "clojure.lang.Compiler$CompilerException")
-    (let [[_ msg file line column]
-          (re-find #"(.*?), compiling:\((.*):(\d+):(\d+)\)" message)]
+    (if-not (empty? location)
       (assoc cause
-             :message msg
-             :file file
-             :file-url (some-> file
+             :file (:clojure.error/source location)
+             :file-url (some-> (:clojure.error/source location)
                                path->url
                                u/transform-value)
-             :path (relative-path file)
-             :line (Integer/parseInt line)
-             :column (Integer/parseInt column)))
+             :path (relative-path (:clojure.error/source location))
+             :line (:clojure.error/line location)
+             :column (:clojure.error/column location))
+      (let [[_ msg file line column]
+            (re-find #"(.*?), compiling:\((.*):(\d+):(\d+)\)" message)]
+        (assoc cause
+               :message msg
+               :file file
+               :file-url (some-> file
+                                 path->url
+                                 u/transform-value)
+               :path (relative-path file)
+               :line (Integer/parseInt line)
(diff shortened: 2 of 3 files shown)


bbatsov:
You should also add some unit tests for the new format.

bbatsov:
@slipset Btw, I think one regexp has to be tuned on the Elisp side as well https://github.com/clojure-emacs/cider/issues/2443

slipset:
Thanks for pointing that one out. From glancing quickly over the issue, it seems to be the exact same error, so this patch might just fix the referenced issue as well.

bbatsov:
Maybe you're right, although probably we use this on some raw error message https://github.com/clojure-emacs/cider/blob/6d6aa42ecf5bad9110cc0a781e7786b87812b624/cider-eval.el#L298 Didn't check this at all - I just remember we had this regexp for parsing errors.

project.clj
```diff
@@ -1,4 +1,4 @@
-(def VERSION "0.19.0-SNAPSHOT")
+  (def VERSION "0.18.0")
```
bbatsov: I think you changed this accidentally.

project.clj
```diff
@@ -48,7 +48,7 @@
                                     :password :env/clojars_password
                                     :sign-releases false}]]
 
-  :profiles {:provided {:dependencies [[org.clojure/clojure "1.8.0"]]}
+  :profiles {:provided {:dependencies [[org.clojure/clojure "1.10.0-RC1"]]}
```
bbatsov: There's also a snapshots profile, btw.

bbatsov:
Yeah, that would be ideal.

slipset:
I see I have a formatting thingy to figure out as well.

PR #560, opened 2018-11-07
Fix instrumentation of recursive functions

vspinu:
There is one side-effect of #552 that I have noticed so far  - functions with recur in them cannot be debugged.  For example instrumenting the following form will throw

```cl

#dbg
(defn foo [a]
  (if (pos? a)
    (recur (dec a))
    :done))
```

The reason for this is that instrumentation doesn't instrument forms containing recur unless in a `loop*` form. I am adding `fn*` to the relevant check because `fn*` is also a target of `recur`. This issue went unnoticed because it could occur only for local let bound functions before, but with localized debugger implementation it occurs with top level defn as well. 

@Malabarba, do you still remember why `contains-recur?` doesn't provision for fn* target? Was there a special reason for that? I have been testing this patch for 2 weeks now and haven't noticed any issues.

### DIFF
--- a/src/cider/nrepl/middleware/util/instrument.clj
+++ b/src/cider/nrepl/middleware/util/instrument.clj
@@ -161,12 +161,13 @@
        :else form))))
 
 (defn- contains-recur?
-  "Return true if form is not a `loop` and a `recur` is found in it."
+  "Return true if form is not a `loop` or a `fn` and a `recur` is found in it."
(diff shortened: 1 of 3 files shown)


Malabarba:
> @Malabarba, do you still remember why contains-recur? doesn't provision for fn* target?

I probably just forgot about it back then.

PR #564, opened 2018-11-08
Use deep-diff for diffing expected vs actual in test failures

omartell:
This changes the existing diff library used by the test middleware `clojure.data/diff` for [lambdaisland.deep-diff](https://github.com/lambdaisland/deep-diff) with the purpose of getting nicer diffs to make it easier to spot the differences of expected vs actual in tests. The advantages of deep-diff is that it recursively compares data structures and produces colorised output. 

This is the current output by `clojure.data/diff`:
screenshot 2018-11-08 23 47 22
screenshot 2018-11-08 23 46 42
screenshot 2018-11-08 23 47 45

This is the kind of output that you get with `deep-diff`:
screenshot 2018-11-07 23 34 01
screenshot 2018-11-08 23 10 57
screenshot 2018-11-08 23 28

I am also raising another PR to make `cider` understand this new format, just want to get your thoughts on this first. On the cider side, I understand there's also `cider-test-ediff`, which might not make sense anymore as the diff Is more explicit about the actual differences.

Let me know what you think.

### DIFF
--- a/src/cider/nrepl/middleware/test.clj
+++ b/src/cider/nrepl/middleware/test.clj
@@ -355,11 +355,6 @@
         (t/send transport (response-for msg :status :no-error)))
       (t/send transport (response-for msg :status :done)))))
 
-(defn handle-diff-backend [handler {:keys [diff-backend]
-                                    :or   {diff-backend :deep-diff}
-                                    :as   msg}]
-  (handler (assoc msg :diff-backend diff-backend)))
-
 ;; Before tools.nrepl-0.2.10, `default-executor` was private and
 ;; before 0.2.9 it didn't even exist.
 (def default-executor (delay (if-let [def (resolve 'ie/default-executor)]
--- a/src/cider/nrepl/middleware/test.clj
+++ b/src/cider/nrepl/middleware/test.clj
@@ -4,12 +4,12 @@
   (:require
    [cider.nrepl.middleware.test.extensions :as extensions]
    [cider.nrepl.middleware.util.coerce :as util.coerce]
+   [cider.nrepl.middleware.test.diff :as diff]
    [clojure.pprint :as pp]
    [clojure.test :as test]
    [clojure.walk :as walk]
    [orchard.misc :as u]
    [orchard.query :as query]
-   [lambdaisland.deep-diff :as dd]
    [cider.nrepl.middleware.stacktrace :as st]))
 
 (if (find-ns 'clojure.tools.nrepl)
@@ -79,19 +79,17 @@
          t :type} m
         c (when (seq test/*testing-contexts*) (test/testing-contexts-str))
         i (count (get-in (@current-report :results) [ns (:name (meta v))]))
-        gen-input (:gen-input @current-report)
-        dd-pprint-str  #(with-out-str (dd/pretty-print %))
-        pprint-str #(with-out-str (pp/pprint %))]
+        gen-input (:gen-input @current-report)]
     ;; Errors outside assertions (faults) do not return an :expected value.
     ;; Type :fail returns :actual value. Type :error returns :error and :line.
     (merge (dissoc m :expected :actual)
            {:ns ns, :var (:name (meta v)), :index i, :context c}
            (when (and (#{:fail :error} t) (not fault))
-             {:expected (dd-pprint-str expected)})
+             {:expected (diff/pprint ie/*msg* expected)})
            (when (and (#{:fail} t) gen-input)
-             {:gen-input (dd-pprint-str gen-input)})
+             {:gen-input (diff/pprint ie/*msg* gen-input)})
            (when (#{:fail} t)
-             {:actual (dd-pprint-str actual)})
+             {:actual (diff/pprint ie/*msg* actual)})
            (when diffs
              {:diffs (extensions/diffs-result diffs)})
            (when (#{:error} t)
--- a/src/cider/nrepl/middleware/test.clj
+++ b/src/cider/nrepl/middleware/test.clj
@@ -9,6 +9,7 @@
    [clojure.walk :as walk]
    [orchard.misc :as u]
    [orchard.query :as query]
+   [lambdaisland.deep-diff :as dd]
    [cider.nrepl.middleware.stacktrace :as st]))
 
 (if (find-ns 'clojure.tools.nrepl)
@@ -79,17 +80,18 @@
         c (when (seq test/*testing-contexts*) (test/testing-contexts-str))
         i (count (get-in (@current-report :results) [ns (:name (meta v))]))
         gen-input (:gen-input @current-report)
+        dd-pprint-str  #(with-out-str (dd/pretty-print %))
         pprint-str #(with-out-str (pp/pprint %))]
     ;; Errors outside assertions (faults) do not return an :expected value.
     ;; Type :fail returns :actual value. Type :error returns :error and :line.
     (merge (dissoc m :expected :actual)
            {:ns ns, :var (:name (meta v)), :index i, :context c}
            (when (and (#{:fail :error} t) (not fault))
-             {:expected (pprint-str expected)})
+             {:expected (dd-pprint-str expected)})
            (when (and (#{:fail} t) gen-input)
-             {:gen-input (pprint-str gen-input)})
+             {:gen-input (dd-pprint-str gen-input)})
            (when (#{:fail} t)
-             {:actual (pprint-str actual)})
+             {:actual (dd-pprint-str actual)})
            (when diffs
              {:diffs (extensions/diffs-result diffs)})
            (when (#{:error} t)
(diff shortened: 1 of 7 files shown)


Malabarba:
Code looks good to me.

If I may add. One thing that bothers me about the current test reports is that it will always try to render the entire object being compared. When the object is reasonably large (such as a list with hundreds of entries) the report buffer can take many seconds to render, even though test itself runs very quickly. 

Because of the fact we use special diffing algorithms, the user doesn't have a lot of control over this. (The usual solution of defining a print-method for the large object has no effect here).

Is that something deep-diff might help with? Are there options to limit the print size?

omartell:
Thanks for the comments @Malabarba. Deep diff relies on [fipp](https://github.com/brandonbloom/fipp), which looks at [\*print-length\*](
https://clojuredocs.org/clojure.core/*print-length*), so we can configure that variable to customize how many elements of a collection are printed in the repl. And setting the value of \*print-length\* can be done trough a [cider variable](
https://github.com/clojure-emacs/cider/blob/master/cider-repl.e) or on demand through a dynamic binding in the code. Does that help?

SevereOverfl0w:
> I am also raising another PR to make cider understand this new format

Can you describe the new format? In the tests, you're looking for tagged data structures, but those won't work with bencode.

The current format is just text which is very easy to render in editors like vim. I don't think colour is easily available in the buffer-type that I use.

omartell:
By format I meant that the diff output will now be a collection of two elements instead of three, as you can see [here](
https://github.com/clojure-emacs/cider-nrepl/pull/564/files#diff-b25efee26665f10240ac5d99c8a5e007R46). The diff transport format is still just text but now with escaped ansi color codes as produced by `deep-diff/pretty-print`. 

On the cider side, the only change is [this](
https://github.com/clojure-emacs/cider/blob/0fce64ab246bdd79fb473ecd03736ed3efbe50f6/cider-test.el#L427-L435) for the following:
```elisp
(cl-destructuring-bind (actual diff) d
                       (insert-label "actual")
                       (insert-rect actual)
                       (insert "\n")
                       (insert-label "diff")
                       (insert-rect (ansi-color-apply diff))
                       (insert "\n"))
``` 

Does that help?

SevereOverfl0w:
The ANSI is a problem for me here. I would have to write something to strip it.

Is there an option in deep-diff to not have ANSI? Does that render it too difficult to read?

omartell:
@SevereOverfl0w  In emacs if I do `(insert-rect diff)` instead of  `(insert-rect (ansi-color-apply diff))` I get no colours. And it's only when I do a regular `(insert diff)` that it prints out the escape codes. I am guessing there's must be something similar in VIM?

There's also a way to customise `deep-diff`'s pretty printer and tell it to not output colours, see [link](https://github.com/lambdaisland/deep-diff/blob/master/src/lambdaisland/deep_diff/printer.clj#L96). You should get pretty much the same output you see in the previous screenshots, just without colours.

SevereOverfl0w:
I'm not sure how emacs is behaving here exactly. But if it's automatically hiding ansi codes, then vim has nothing like that.

Having a flag to remove the ANSI color would make this suitable for my use. I don't think it's too vim-specific a requirement, so makes sense to be added.

If you don't have the time to add this, I can look into it at a later date when I have the energy/time.

bbatsov:
On a more general note - ideally the font-locking should be left to the editors IMO. The ansi colors are great for console apps, but they are a bit limiting when you're consuming the library programatically. 

Perhaps we can just use the library for the diffing and apply regular pprint and assume font-locking on the editor side? @plexus I'd also love to hear your thoughts on this, as I'm not sure you've originally intended for the lib to used in this manner.

plexus:
> I'm not sure you've originally intended for the lib to used in this manner.

The library works in two stages, it first performs the diff, which yields a data structure that's similar in shape to the input, but contains `Deletion`, `Insertion`, and `Mismatch` objects to mark the differences. These are `defrecord`s.

The second phase is pretty printing, where it hooks into FIPP to render these tagged objects with a `-` or `+` in front, and with the red/green color applied. Note that for map entries it's the key that gets wrapped, meaning you need to also do custom rendering of maps to make this work.

So you can use the diffing "backend" and hook up your own pretty printing, but pretty printing and coloring should really be done at the same time. Without color the output is ambiguous. Don't try to take the output, strip the ansi color codes, then font-lock it in Emacs. You will not get satisfactory results.

The actual printing is done with Puget, which can output either ANSI colors or HTML, maybe it's possible to hook up another backend there to get the text+colors as some kind of data structure that you can render in Emacs.

SevereOverfl0w:
Data structure would be something I could work with in vim. I might even be able to highlight them. Maybe.

omartell:
If some clients prefer a non-colored diff output, then another option is to support both diff libraries and allow the user to specify which one they’d like to use. So, making the diff backend pluggable in cider itself.

SevereOverfl0w:
Ultimately I think this format is superior. I'd much prefer to have access to it. But I accept that this is probably a lower barrier than configuring the printer.

bbatsov:
> If some clients prefer a non-colored diff output, then another option is to support both diff libraries and allow the user to specify which one they’d like to use. So, making the diff backend pluggable in cider itself.

I guess this would be the best approach for the time being. Let's work in this direction. I'd be fine with CIDER using the new deep diff by default.

bbatsov:
@omartell I plan to cut the new release around Christmas/NY, so if you want to this landing in the next release we'll have to act relatively quickly.

SevereOverfl0w:
I'm going to take a look now into using a custom printer to disable ansi.

SevereOverfl0w:
I have working local changes. However, the way it looks in vim makes me question the usefulness without the color:

maim-2018-12-23-23 29 32

Once the map is large enough for line-breaks though, it is more valuable. I may look into doing basic highlighting based on a simple regex, or in the future perhaps extending this to return a data-fied form of pretty print rather than ANSI.

omartell:
I've added a middleware that allows you to switch the diff backend between `clojure.data/diff` and `lambdaisland.deep-diff/diff`, see the new [commit](https://github.com/clojure-emacs/cider-nrepl/pull/564/commits/bf23ab933c84ec8f0d6c9fb5e6bcf2108a248c72). It defaults to using deep-diff. The middleware accepts `:deep-diff` and `:data-diff` as valid options. Let me know if this is what you had in mind.

bbatsov:
I agree - we can't change the default in `cider-nrepl`. We can make the new diffing default in CIDER, though.

src/cider/nrepl/middleware/test.clj
```diff
@@ -353,6 +355,11 @@
         (t/send transport (response-for msg :status :no-error)))
       (t/send transport (response-for msg :status :done)))))
 
+(defn handle-diff-backend [handler {:keys [diff-backend]
```
bbatsov: I don't think that this is needed. Probably we can just have clients send via the test requests what diff backend to to use or something like this. A separate middleware just to add some key to a message seems a bit excessive to me.

omartell:
Yes, backwards compatibility is a good point. I've removed the middleware as you suggested and changed the default to `:data-diff` in the multimethod definition.

src/cider/nrepl/middleware/test.clj
```diff
@@ -79,17 +80,18 @@
...
         gen-input (:gen-input @current-report)
+        dd-pprint-str  #(with-out-str (dd/pretty-print %))
         pprint-str #(with-out-str (pp/pprint %))]
     ;; Errors outside assertions (faults) do not return an :expected value.
     ;; Type :fail returns :actual value. Type :error returns :error and :line.
     (merge (dissoc m :expected :actual)
            {:ns ns, :var (:name (meta v)), :index i, :context c}
            (when (and (#{:fail :error} t) (not fault))
-             {:expected (pprint-str expected)})
+             {:expected (dd-pprint-str expected)})
```
bbatsov: Why is this changed unconditionally? Shouldn't you be using the methods from the new namespace here?

omartell: Yes, you're right. Updated the PR.

bbatsov:
@omartell Sorry for dropping the ball on this. Can you please rebase this on top of `master` and document the new functionality somewhere. 

Seems to me that the PR is more or less in a mergeable state. //cc @SevereOverfl0w @Malabarba

SevereOverfl0w:
I think the PR is in a satisfactory condition, based on my previous look. I can't see me extending replant to use this because the formatting is too difficult.

Malabarba:
LGTM

vemv:
This would have been an interesting change. Anyway I'd have looked into the following nuances:

* fipp has https://github.com/brandonbloom/fipp/issues/72 atm
* not every failure report can be meaningfully diffed: from exceptions, to `is false`, to `(is (= 1 2))` where the diff will feel redundant.
  * i.e. there should be some kind of heuristic for determining if something is diffable?
* maybe an easy toggle should be provided for turning the feature off in case a bug arises from it
  * system properties are interesting here because they can be toggled from the repl (aside from Lein profiles etc)

PR #565, opened 2018-11-13
Bumping fipp deps for JDK11 support

tetigi:
Just a simple PR to bump the fipp version regarding [this PR](https://github.com/brandonbloom/fipp/commit/1d5b74a3cd384b2673c9f45c204838bab20ac7b3).

The latest fipp version includes updates to rrb-vector, which now works nicely with JDK11 rather than throwing weird errors.

### DIFF
--- a/deps.edn
+++ b/deps.edn
@@ -3,7 +3,7 @@
         cider/orchard {:mvn/version "0.3.1"}
         thunknyc/profile {:mvn/version "0.5.2"}
         mvxcvi/puget {:mvn/version "1.0.2"}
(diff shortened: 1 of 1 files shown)


bbatsov:
@tetigi We don't really use deps.edn in cider-nrepl. The dep is already updated in project.clj. I think @arichiardi uses deps.edn when he hacks on cider-nrepl. All deps should be examined and synced I guess.

arichiardi:
Yeah but many thanks! This is also good for working against shas :smile:

Issue #567, opened 2018-11-18
Provide cider function to find all known impls of a protocol function 


kapilreddy:
The feature-request has been posted on "cider"

https://github.com/clojure-emacs/cider/issues/1969

Creating an issue here to send a pull request and start a discussion about the solution.

PR #568, opened 2018-11-18
Add support to list impls of protocol in `info` middleware

kapilreddy:
Use ns-imports of all namespaces to list all classes that
extend the protocol in question.

HACK - Modify classname to `map->{classname}`. The jump will only work for
defrecords at the moment.

https://github.com/clojure-emacs/cider-nrepl/issues/567

### DIFF
--- a/src/cider/nrepl/middleware/info.clj
+++ b/src/cider/nrepl/middleware/info.clj
@@ -73,6 +73,7 @@
                   f))
                f)))))
 
+
 (defn info
   [{:keys [ns symbol class member] :as msg}]
   (let [[ns symbol class member] (map u/as-sym [ns symbol class member])]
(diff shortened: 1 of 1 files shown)


bbatsov:
@kapilreddy Thanks for looking in this!

I think the right home for this code would be `orchard`, not `cider-nrepl`. Ideally `cider-nrepl` should be a very think wrapper around the orchard library, which does the actually heavy lifting.

kapilreddy:
Makes sense. Will push the changes on orchard.

bbatsov:
@kapilreddy 👍

bbatsov:
@arichiardi I guess it's time to remind you of your promise. ;-)

arichiardi:
Ah ah definitely thanks for the reminder 😅

src/cider/nrepl/middleware/info.clj
```diff
@@ -86,7 +87,28 @@
...
+                  xs (map (fn [x]
+                            (let [xs (str/split (pr-str x)
+                                                #"\.")]
+                              (str (str/replace (str/join  "." (drop-last xs))
+                                                #"\_"
+                                                "-")
+                                   "/map->"
+                                   (last xs))))
+                          x)]
+              {:candidates (into {}
```
arichiardi: is this returning `:candidates`? shouldn't it go in `compilment`?

bbatsov: Those are different candidates. :-)

bbatsov: I think we use the same key for several things here and there (I think it was used somewhere in `info` as well).

bbatsov:
@arichiardi Any progress with this?

kapilreddy:
@arichiardi Hey, Let me know if you need help moving this into Orchard and/or testing it.

arichiardi:
I think I need help yes, my Clojure time has been reduced substantially 😒

viebel:
@bbatsov @arichiardi What's the status of this issue?

bbatsov:
It's waiting for a volunteer to drive it to the finish line.

viebel:
@kapilreddy Would you like to volunteer?

vemv:
Probably it doesn't make much sense to keep the PR open for yet another year, it's bit of a distraction when checking out our projects' status.

Volunteering welcome! (Probably at Orchard level next time)

PR #573, opened 2018-12-07
Fix inspector silently doing nothing if eval errored

alexander-yakushev:
**The problem**
If you inspect an expression that throws an exception (e.g. `(/ 1 0)`), the inspector silently discards the exception as if nothing has happened. This is annoying.

**Why?**
Cider already has machinery for things like that, e.g. if macroexpansion throws an error, it will be displayed in a special cider middleware error buffer. That is implemented by macroexpansion middleware using `with-safe-transport` which, in the case of error, returns a message with status `[:done :macroexpansion-error]`. On the Emacs side, Cider's `cider-nrepl-send-sync-request` listens to `:done` events, and if there's an error status after it the error buffer will be raised.
Why doesn't it work for inspector? Because at some point (#164) the inspector has been changed from having its own middleware command to riding upon `eval` command. `eval` command is asynchronous and can return many results; and if the exception is thrown, it will send the exception as separate message and finish up with `:done` message. And `cider-nrepl-send-sync-request` ignores everything except `:done`.

**What is this fix?**
A hack to make `eval :inspect true` interceptor to short-circuit the failed eval with `[:done :inspect-eval-error]` message. Also, some refactoring to make the inspector middleware more like other middlewares.

**Does it have to be like this?**
I have no idea. I don't know if CLJS implementation will work after this, why it must use `eval` command, and what can be safely refactored. Need someone's help with CLJS side, this is too complicated for me.

BTW, it might be a good idea to hold this until 0.19 release.

### DIFF
--- a/src/cider/nrepl/middleware/inspect.clj
+++ b/src/cider/nrepl/middleware/inspect.clj
@@ -1,7 +1,8 @@
 (ns cider.nrepl.middleware.inspect
   (:require
    [cider.nrepl.middleware.util.cljs :as cljs]
-   [cider.nrepl.middleware.util.error-handling :refer [base-error-response]]
+   [cider.nrepl.middleware.util.error-handling :refer [base-error-response
+                                                       with-safe-transport]]
    [orchard.inspect :as inspect]))
 
 ;; Compatibility with the legacy tools.nrepl and the new nREPL 0.4.x.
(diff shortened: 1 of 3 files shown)


bbatsov:
> I have no idea. I don't know if CLJS implementation will work after this, why it must use eval command, and what can be safely refactored. Need someone's help with CLJS side, this is too complicated for me.

`eval` is the only op we native support for ClojureScript, everything else is done in Clojure. I guess that's the reasoning behind this the breaking change, and I also guess your changes will not break this. It should be easy to test with the basic nashorn REPL or something like this.

src/cider/nrepl/middleware/inspect.clj
```diff
@@ -46,9 +47,16 @@
     (recv [this] (.recv transport))
     (recv [this timeout] (.recv transport timeout))
     (send [this response]
-      (if (contains? response :value)
-        (inspect-reply msg response)
-        (.send transport response))
+      (cond (contains? response :value)
+            (inspect-reply msg response)
+
+            (contains? (:status response) :eval-error)
```
bbatsov: I guess here you can explain the hack a bit, everything else looks good to me.

src/cider/nrepl/middleware/inspect.clj
```diff
@@ -61,48 +69,39 @@
   [handler msg]
   (handler (eval-msg msg)))
 
-(defn- success [{:keys [transport] :as msg} inspector]
-  (transport/send transport (response-for msg :value (:rendered inspector) :status :done)))
-
-(defn- failure [{:keys [transport] :as msg} err err-kw]
-  (transport/send transport (base-error-response msg err err-kw :done)))
+(defn- success [msg inspector]
```
bbatsov: I guess we can give this some more meaningful name now - e.g. `inspector-response`.

bbatsov:
@alexander-yakushev ping :-)

alexander-yakushev:
Yeah, sorry, couldn't find the time to investigate the cljs portion yet.

alexander-yakushev:
Good thing I checked with ClojureScript, this patch didn't work well there:) Now I fixed that and addressed your comments.

PR #576, opened 2018-12-23
[Fix #575] Upgrade to Eastwood 0.3.4

### LINKED ISSUE #575: Bump the Eastwood dep
I tried updating Eastwood to 0.3.4, but I ran into:

```
== Linting cider-nrepl.plugin ==
Exception thrown during phase :analyze+eval of linting namespace cider-nrepl.plugin
CompilerException java.lang.IllegalArgumentException: Cannot open <nil> as a Reader., compiling:(classpath.clj:407:35)
	clojure.lang.Compiler$InvokeExpr.eval (Compiler.java:3657)
	clojure.lang.Compiler$InvokeExpr.eval (Compiler.java:3651)
	clojure.lang.Compiler$DefExpr.eval (Compiler.java:451)
	clojure.lang.Compiler.eval (Compiler.java:6932)
	clojure.lang.Compiler.load (Compiler.java:7379)
	clojure.lang.RT.loadResourceScript (RT.java:372)
```

There were also many issues related to our current use of conditional requires.

I hope @slipset will help us solve those issues.

slipset:
-  add excluded namespaces

This is not done yet, but with three excluded ns's, linting passes without exceptions, but with errors.
One of the exceptions (thrown from `test-sessions` is a bug in eastwood (https://github.com/jonase/eastwood/issues/306)
The two others might be bugs in eastwood, I don't know.

I'll leave this here for now, and work a bit more on it, so please don't merge yet.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -110,4 +110,4 @@
                                          if-class [[:block 1]]}}}
 
              :eastwood {:plugins [[jonase/eastwood "0.3.4"]]
-                        :eastwood {:config-files ["eastwood.clj"] :exclude-namespaces [cider-nrepl.plugin cider.tasks cider.nrepl.test-session]}}})
+                        :eastwood {:config-files ["eastwood.clj"] :exclude-namespaces [cider-nrepl.plugin cider.tasks cider.nrepl.test-session] :exclude-linters [:implicit-dependencies]}}})
--- a/project.clj
+++ b/project.clj
@@ -109,5 +109,5 @@
                                          try-if-let [[:block 1]]
                                          if-class [[:block 1]]}}}
 
-             :eastwood {:plugins [[jonase/eastwood "0.2.5"]]
-                        :eastwood {:config-files ["eastwood.clj"]}}})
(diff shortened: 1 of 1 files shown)


slipset:
Excluded the `implicit-dependencies` linter and everything passes

slipset:
Interestingly, this is caught by the `implicit-dependencies` linter:
https://github.com/clojure-emacs/cider-nrepl/blob/master/src/cider/nrepl.clj#L101
I would assume that since `pprint` is explicitly required here
https://github.com/clojure-emacs/cider-nrepl/blob/master/src/cider/nrepl.clj#L5
it would be ok to change this line?

bbatsov:
So, I guess that's mergeable now, right?

slipset:
Yes, but it’s less than perfect.

bbatsov:
As is life. 😉 I'll keep this open for a while in case you decide to tweak it a bit more. I'm not very worried about the conditional requires as I plan to remove them in the next release, but ideally Eastwood should understand those.

PR #577, opened 2018-12-27
Test repl setup improvements

plexus:
These are all changes I made in the context of #569 that are not directly related to fixing tests (although some are necessary prerequisites for what follows, e.g. the min/max-clj-version fixes).

I'll make a separate PR with the testing fixes once this is merged.

----

### Explicitly add refactor-nrepl.plugin/middleware to Leiningen config

Before this Leiningen middleware was being added implicitly, but in the latest
Leiningen this causes a warning. Configure it explicitly instead, and disable
Leiningen's support for implicit middleware.

----

### Add a ^:max-clj-version test filter, fix min-clj-version

To be able to account for subtle differences between Clojure versions we have
metadata based filters. The min-clojure-version filter however no longer worked
correctly because Clojure 1.10 would sort before 1.9. Fixed this by adding
explicit version parsing, and also added a max-clj-version filter.

----

### Add REPL helpers: repl-session! / close-session!

Many tests make use of `session-fixture`, which sets up an nREPL session, and
tears it down again. When working on tests this makes it tedious to quickly
evaluate test forms.

These helper functions set up the nREPL session similar to `session-fixture` but
allow you to manage the start/stop lifecycle yourself, so you can start the
session once and get to work.

----

### Add a check to the build logic to see if node is present

The ClojureScript tests rely on node being present, if it's not there then they
will fail in non-obvious ways. To provide quick feedback we do a check first
before starting the tests.

----

### Makefile: rename VERSION to CLOJURE_VERSION

For the sake of clarity and unambiguity, add some namespacing to this env var.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,10 +8,10 @@ script:
   - make $TARGET
 env:
   matrix:
-    - VERSION=1.8 TARGET=test-clj
-    - VERSION=1.8 TARGET=test-cljs
-    - VERSION=1.9 TARGET=test-clj
-    - VERSION=1.9 TARGET=test-cljs
+    - CLOJURE_VERSION=1.8 TARGET=test-clj
+    - CLOJURE_VERSION=1.8 TARGET=test-cljs
+    - CLOJURE_VERSION=1.9 TARGET=test-clj
+    - CLOJURE_VERSION=1.9 TARGET=test-cljs
   global:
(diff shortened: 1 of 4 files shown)


plexus:
> I was under the impression that most ClojureScript tests were supposed to use Nashorn.

Seems they've been using Node since 2015: https://github.com/clojure-emacs/cider-nrepl/commit/485615fec10e0327a0db40d610d16fbe32309b91#diff-ca473b121aba839e24575deb13ee5f90R19

plexus:
@bbatsov I've taken out the `:implicits` commit

bbatsov:
Thanks for working on this! I really appreciate it! 🙇

Issue #578, opened 2018-12-27
ClojureScript tests regularly hang/time out


plexus:
This is a spin-off of #569. I've managed to find and fix all "real" test failures, that leaves us with one more issue, being that regularly (one time out of two) a ClojureScript build will stop in the middle of a test and just hang there until after ten minutes the build times out.

Examples: (I'll add more links as I come across them)

- https://circleci.com/gh/plexus/cider-nrepl/145 
- https://travis-ci.org/clojure-emacs/cider-nrepl/jobs/473286780
- https://travis-ci.org/clojure-emacs/cider-nrepl/jobs/473291631

This has been observed on different Java version (8, 11) and ClojureScript versions (1.9, 1.10), so it seems it's a general problem, some kind of race condition. My hunch is that a message is being sent to the ClojureScript environment before it's properly able to receive it, and so a reply is never received.

This is a pain, especially since I'm not sure if this will be reproducible locally (or consistently). Still here are some ideas that could help to pinpoint the issue.

- [ ] Add a flag to generate output similar to `*nrepl-messages*`, so you can see which messages go back and forth during testing
- [ ] Find all places where exceptions are being swallowed, and make that at least during testing they get printed out (I came across one or two of these already)
- [ ] start a background shell in the Makefile that waits five minutes and then calls `jstack`, so we can see where it's blocking
- [ ] print test var names as they run, not just namespaces, so we can see which tests in particular suffer from this
- [ ] Review the code path of these tests, where are we waiting for an async response? What can cause it not to arrive?

plexus:
Relatedly: is there an opportunity to simplify the ClojureScript test setup as @arichiardi hinted at in #555?

bbatsov:
> Relatedly: is there an opportunity to simplify the ClojureScript test setup as @arichiardi hinted at in #555?

Isn't he asking for ClojureScript support in `cider-test`?

bbatsov:
> This has been observed on different Java version (8, 11) and ClojureScript versions (1.9, 1.10), so it seems it's a general problem, some kind of race condition. My hunch is that a message is being sent to the ClojureScript environment before it's properly able to receive it, and so a reply is never received.

It's interesting to see if those race conditions will still happen with nREPL 0.6-snapshot, as @cgrand simplified recently a lot the code evaluation logic (see https://github.com/nrepl/nrepl/pull/98).

 >  Add a flag to generate output similar to *nrepl-messages*, so you can see which messages go back and forth during testing

[...]

PR #580, opened 2018-12-28
CI test fixes

plexus:
This branch contains the "real" fixes that came out of #569. It also adds Clojure 1.10 to the build matrix.

See individual commits for exact changes. Unless #578 gets in the way this should go green now :crossed_fingers:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,8 @@ env:
     - CLOJURE_VERSION=1.8 TARGET=test-cljs
     - CLOJURE_VERSION=1.9 TARGET=test-clj
     - CLOJURE_VERSION=1.9 TARGET=test-cljs
+    - CLOJURE_VERSION=1.10 TARGET=test-clj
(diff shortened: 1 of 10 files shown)


bbatsov:
> See individual commits for exact changes. Unless #578 gets in the way this should go green now 🤞

It seems indentation problems got in the way. :-)

bbatsov:
All the changes look good to me.

plexus:
Ah yes, let's try that again. Bozhidar I also added back the `:implicits`, seems it's actually mranderson which is triggering the warning.

plexus:
So close! But still one failure I hadn't seen before in the debug test.

https://travis-ci.org/clojure-emacs/cider-nrepl/jobs/473008395

bbatsov:
> So close! But still one failure I hadn't seen before in the debug test.

I think there's often been flakiness in this test - I've definitely seen this failure before. I've long suspected it's some race condition.

I'm merging the PR regardless, as it gets us extremely close to the coveted destination and to the release of 0.19.

PR #582, opened 2018-12-30
Increase timeout in debug-integration-test, fail loudly upon timeout

plexus:
These tests use special `--->` and `<---` macros to sketch out an interaction
with the debugger, but the `<---` macro (nrepl-recv) did not account for the
possibility of a timeout. A timeout would result in receiving a `nil` message,
which was then treated as an empty map as far as assertions are concerned.

When this happened the actual and expected messages would get out of sync,
causing all future assertions in the same test var to also fail.

This commit increases the timeout from 200 to 500 ms, and throws an exception
when a timeout does occur.


bbatsov:
Nicely done! 🙇

PR #583, opened 2018-12-30
When running test-cljs, after 5 minutes print stacktraces and exit

plexus:
A successful ClojureScript build takes 2~3 minutes, however sometimes it will
block until CI times out, and we're not sure yet why.

To further diagnose this issue we now start a background shell which after five
minutes will print the stacktraces of all threads of all active JVMs using
jcmd/jstack. Hopefully this provides some insight in what/where the tests get
stuck.

Additionally this kills all Java processes after five minutes so CI finishes a
little quicker.

- [ ] All tests are passing

Thanks!


bbatsov:
I think it'd be nice if you added the description of the PR somewhere in the beginning of the script and probably a note in the travis config (e.g. why are we doing this only for ClojureScript builds and why it needs to be run in the background). 

Apart from this small remark - everything looks great to me.

plexus:
Added comments as requested.

The good news is that we already got some diagnostic stack traces \o/ 
https://travis-ci.org/clojure-emacs/cider-nrepl/jobs/473545815

PR #585, opened 2019-01-10
Use Nashorn for CLJS tests

### LINKED ISSUE #578: ClojureScript tests regularly hang/time out

cichli:
Fixes #578.

When I added the CLJS tests, Nashorn was unavailable on JDK 7 – but we no longer support that. Nashorn is now deprecated as of JDK 11, the suggested alternative being Graal, but that is unavailable on JDK 8 which we do still support. Additionally `cljs.repl.graaljs` was only added in ClojureScript 1.10.439.

For now: just use Nashorn; if it's removed in some future JDK version then what we do will depend on which JDKs/CLJS versions we still support at that time.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,9 +4,10 @@
 /checkouts
 pom.xml
 pom.xml.asc
+nashorn_code_cache
 *.jar
 *.class
-.cljs_node_repl
+.cljs_nashorn_repl
 .lein-deps-sum
 .lein-failures
 .lein-plugins
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: test-clj test-cljs eastwood cljfmt cloverage install smoketest release deploy clean check_node detect_timeout
+.PHONY: test-clj test-cljs eastwood cljfmt cloverage install smoketest release deploy clean detect_timeout
 
 CLOJURE_VERSION ?= 1.9
 export CLOVERAGE_VERSION = 1.0.13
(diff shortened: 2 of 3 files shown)


bbatsov:
Fine by me! Thanks for tackling this!

PR #586, opened 2019-01-15
Faster tests and build improvements

cichli:
Historically the ClojureScript tests have always been somewhat slow but I noticed they're taking even longer since switching to Nashorn. On my machine:

```
Ran 30 tests in 81.063 seconds
162 assertions, 0 failures, 0 errors.
```

If we instead call `piggieback-fixture` once per test namespace rather than once per test (N.B. this required a small change to the inspector so we could clear its state between test runs, but apart from that the tests are stateless enough that this works fine):

```
Ran 30 tests in 27.430 seconds
162 assertions, 0 failures, 0 errors.
```

The equivalent change for `session-fixture` is not necessary for the Clojure tests – it simply doesn't have any real impact on the time taken.

Additionally, we can cache the result of `cljs.repl.nashorn/repl-env` and reuse it across test namespaces:

```
Ran 30 tests in 15.642 seconds
162 assertions, 0 failures, 0 errors.
```

I'd say they're now fast enough that there's not much point having separate `:test-clj` and `:test-cljs` profiles – I've now merged them and made some further simplifications to the build process, most importantly:

* Consolidate all lint/coverage/etc. config that was in `Makefile` into `project.clj`
* Use composite profiles where possible in `project.clj` – generally this means we only need `with-profile` for mranderson and the lint tools; `lein test` works, and the tests are on the classpath and can be run when using `lein repl`
* Switch to Travis Xenial infrastructure
* Use `#{openjdk8 openjdk11 openjdk-ea}` in the Travis matrix – the Oracle JDKs are becoming non-free this/next year anyway (and currently [have issues](https://github.com/travis-ci/travis-ci/issues/10290) on Xenial)
* Remove the `smoketest` target from the `test` target – it's a bit surprising that this installs an artifact to the local repository, so let's call it explicitly in the Travis config

Here's an [example build](https://travis-ci.org/cichli/cider-nrepl/builds/479951764) finishing in 8 min 11 sec, compared to a [master build](https://travis-ci.org/clojure-emacs/cider-nrepl/builds/479411377) in 31 min 43 sec (N.B. this is the time for all _required_ jobs to finish; it doesn't include the `allow_failures` jobs).


bbatsov:
Fantastic work! 🙇 🙇 🙇 

If you have some time, you can do some of that magic to Orchard and the nREPl projects - it'd be nice if they all had consistent build setups.

PR #587, opened 2019-01-16
Fix NPE crash in "slurp" middleware

tatut:
Fallback to "application/octet-stream" if a content-type cannot be determined for the file/url. 

Don't show the content of unrecognized files, but return a description "#binary[location=<path or url>,size=<size in bytes>".

Added test in slurp_test.clj to test unrecognized file.


cichli:
Thanks for the contribution! Please will you rebase to master? It's just a small conflict. Note that this also conflicts with #589.

The changes look fine to me – @arrdem do you mind double-checking?

cichli:
Cheers @tatut @arrdem!

PR #590, opened 2019-01-22
Use new session exec in test and refresh middleware

### LINKED ISSUE #581: Update the test middleware for nREPL 0.6
nREPL 0.6 features a simplified and more robust evaluation model (see https://github.com/nrepl/nrepl/pull/98), but this break the current implementation of the test middleware, which was making some assumptions about the internals of the eval middleware. 

This ticket is a reminder to address this before nREPL 0.6 is released.

@jeffvalk would you like to tackle this? The updates should be pretty simple - seems to me that only the `with-interruptible-eval` macro should be updated.

cichli:
Pending 0.6.0 release of nREPL – piggieback will also need updating to fix the build. I will take a look at that.

* Use session exec in test middleware
* Use session exec in refresh middleware
  * This makes the relevant ops interruptible and also serialised with evaluations / other ops that use exec (in the same session)
  * We still serialise all refresh ops globally but now using an explicit lock rather than an agent

Fixes #581.


bbatsov:
Nicely done! Btw, probably we should document the usage of interruptible eval, so people would know it’s there mostly so that the ops can be interrupted. (I know recently there was some confusion about why it was used in the test middleware)

As I’m without a computer for the next couple of months I’ve given you permissions to push nREPL updates to Clojars as well (I don’t recall if we published builds automatically or not).

CHANGELOG.md
```diff
@@ -2,6 +2,16 @@
 
 ## master (unreleased)
 
+### New features
+
+* The `refresh` and `test` ops are now interruptible.
```
bbatsov: I think `test` has been interruptible for quite a while now.

Issue #588, opened 2019-01-16
complete causing 100% CPU emacs/JVM unresponsiveness


jappy:
## Expected behavior

Whenever I type a keyword in Emacs I would like to just keep typing without Emacs or my JVM becoming unresponsive.

## Actual behavior

Whenever I type a keyword in Emacs while jacked in my JVM with millions of keyword using cider/cider-repl my CPU blows up to 100% for several minutes. Sometimes it goes back to normal, sometimes it doesn't.

## Steps to reproduce the problem

My JVM has millions of keywords. This issue only happens when you have lots of them (I have 14+ million of them). Then when you are jacked in from emacs and type a keyword the `handle-complete` function is called and all of the threads on my system are filled with calls through this function--thread dump example below.

My guess is that this code is not written with the possibility that there might be many millions of keywords in a system... and that it is trying to find all of the matches and sort them, etc. My best guess anyway. So to reproduce make a ton (like I said I have 14+ million) of pretty lengthy keywords and put them in the same namespace. Then jack into the JVM and then start typing one of the keywords in this namespace.

Here is a stack trace (many thread in the VM are doing the same):

```
"clojure-agent-send-off-pool-6" #27 prio=5 os_prio=31 cpu=26244.47ms elapsed=33.95s tid=0x00007f8ca13e6800 nid=0x14e03 runnable  [0x000070000e982000]
   java.lang.Thread.State: RUNNABLE
	at compliment.core$reify__15238.compare(core.clj:35)
	at clojure.core$sort_by$fn__5963.invoke(core.clj:3131)
	at clojure.lang.AFunction.compare(AFunction.java:51)
	at java.util.TimSort.binarySort(java.base@11.0.1/TimSort.java:296)
	at java.util.TimSort.sort(java.base@11.0.1/TimSort.java:239)
	at java.util.Arrays.sort(java.base@11.0.1/Arrays.java:1441)
	at clojure.core$sort.invokeStatic(core.clj:3115)
	at clojure.core$sort_by.invokeStatic(core.clj:3119)
	at clojure.core$sort_by.invoke(core.clj:3119)
	at compliment.core$completions.invokeStatic(core.clj:90)
	at compliment.core$completions.invoke(core.clj:54)
	at cider.nrepl.middleware.complete$complete.invokeStatic(complete.clj:20)
	at cider.nrepl.middleware.complete$complete.invoke(complete.clj:10)
[...]

PR #599, opened 2019-02-24
[complete] Limit the maximum number of candidates

alexander-yakushev:
Addresses #588.

### DIFF
--- a/src/cider/nrepl/middleware/complete.clj
+++ b/src/cider/nrepl/middleware/complete.clj
@@ -11,13 +11,15 @@
   [{:keys [ns symbol context extra-metadata] :as msg}]
   (let [ns (u/as-sym ns)
         prefix (str symbol)
-        extra-metadata (set (map keyword extra-metadata))]
-    (if-let [cljs-env (cljs/grab-cljs-env msg)]
-      (cljs-complete/completions cljs-env prefix {:context-ns ns
-                                                  :extra-metadata extra-metadata})
-      (jvm-complete/completions prefix {:ns ns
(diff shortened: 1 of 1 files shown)


cichli:
Thanks! I think it's sensible but maybe 100 is an order of magnitude too conservative, considering some frontends like Company and Ivy make it convenient to search candidates. For example, `clojure.core` has 659 publics in 1.10, and I'd expect to have them all available when completing `clojure.core/`. I think we should optimise this limit so that the JVM-side of things (including encoding the response) still performs acceptably. If there are problems on the client above a certain number of candidates, we can also do some limiting there.

One other thing though: the stacktrace in the linked issue points to [where the candidates are sorted](https://github.com/alexander-yakushev/compliment/blob/master/src/compliment/core.clj#L86-L92) in Compliment. Should we limit the number of candidates there (and likewise in cljs-tooling) before sorting them?

alexander-yakushev:
Damn, you are absolutely right on both points. I'll make the necessary changes to Compliment instead.

cichli:
I wonder if Compliment could also be a bit smarter about where the limit is applied. Maybe it's only needed for the keyword completions, at least until someone complains about performance in other cases :-).

bbatsov:
Yeah, that's a good point. Alternatively we can expose the ability to set the limit in the query, so users can do whatever they want.

bbatsov:
@alexander-yakushev So what are we doing with this PR?

bbatsov:
@alexander-yakushev ping 2 :-)

PR #601, opened 2019-02-27
Remove :provided profile mapping in project.clj

arichiardi:
This patch removes the :provided key as it seems not necessary for correctly
packaging and testing cider-nrepl. It actually gets kind of in the way when
testing where the :1.8 was always loaded regardless of the env var version.

It is more here for myself to know more but the doc says:

> The :provided profile is used to specify dependencies that should be
> available during jar creation, but not propagated to other code that depends
> on your project. These are dependencies that the project assumes will be
> provided by whatever environment the jar is used in, but are needed during
> the development of the project. This is often used for frameworks like Hadoop
> that provide their own copies of certain libraries.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -54,9 +54,7 @@
                                     :password :env/clojars_password
                                     :sign-releases false}]]
 
-  :profiles {:provided [:1.8]
-
-             :dev {:dependencies [[boot/base "2.8.2"]
+  :profiles {:dev {:dependencies [[boot/base "2.8.2"]
                                   [boot/core "2.8.2"]
                                   [leiningen-core "2.8.3"]]}
 


cichli:
Could you be more specific about the problems it's causing? If you set `CLOJURE_VERSION` when you run some Makefile target it will use `with-profile +$(CLOJURE_VERSION)` and include those dependencies in the default scope, overriding those in the provided scope. If it doesn't, that's an issue with the Makefile target in question IMO.

We do expect those versions (at a minimum) of CLJ/CLJS to be provided at runtime for things to work. Including them in the provided profile means they will be used by default for any Leiningen task (e.g. `lein repl`) so the CLJS tests will work. Otherwise, Leiningen will include some default version of Clojure (1.7 before Leiningen 2.9.0 and 1.10 after that).

arichiardi:
@cichli Thanks for the detailed explanation I will try to actually understand more where the failure is. I empirically observed that when doing `make test` it was failing and outputting:

```
$ make test
lein with-profile +1.9,+test,+plugin.mranderson/config test
...

Error encountered performing task 'test' with profile(s): 'base,system,user,1.8,dev,1.9,test,config'
Tests failed.
```

So maybe I was confused by that but maybe it is actually innocuous and just a lein output "issue"..

PR #603, opened 2019-03-11
Allow for multiple Clojures when checking Clojure version

dpsutton:
## Problem:

Check out [nippy](https://github.com/ptaoussanis/nippy) and jack in. You are greeted with:

```shell
;; Clojure 1.10.0, Java 1.8.0_201                ;; <----------------- note 1.10.0
;;     Docs: (doc function-name)
;;           (find-doc part-of-name)
;;   Source: (source function-name)
;;  Javadoc: (javadoc java-object-or-class)
;;     Exit: <C-c C-q>
;;  Results: Stored in vars *1, *2, *3, an exception in *e;
WARNING: CIDER requires cider-nrepl to be fully functional. Many things will not work without it!
         More information.
user> 
```

If you run the start up command on the shell you are greeted with:

```bash
[dan@fedora nippy]$ /home/dan/bin/lein update-in :dependencies conj \[nrepl\ \"0.6.0\"\] -- update-in :plugins conj \[cider/cider-nrepl\ \.21.2-SNAPSHOT\"\] -- repl :headless :host localhost
Warning: cider-nrepl requires Clojure 1.8 or greater.    < ----------- startup thinks 1.5.1
Warning: cider-nrepl will not be included in your project.
```

## The reason

A common pattern is an old clojure in the deps and a modern clojure in
a profile

```clj
(defproject project
  :dependencies [[org.clojure/clojure "1.5.1"]]
  :profiles {:1.10 {:dependencies [[org.clojure/clojure "1.10.0"]]})
```
When checking for the clojure version allow for the presence of multiple clojure versions rather than just taking the first one.

- [x] All tests are passing

Thanks!

### DIFF
--- a/src/cider_nrepl/plugin.clj
+++ b/src/cider_nrepl/plugin.clj
@@ -6,17 +6,22 @@
    [clojure.java.io :as io]
    [leiningen.core.main :as lein]))
 
-(def min-lein-version "2.8.2")
+(def minimum-versions {:lein    "2.8.2"
+                       :clojure "1.8.0"})
+
+(defn valid-version? [kind version] (lein/version-satisfies? version (minimum-versions kind)))
+(def valid-lein-version? (partial valid-version? :lein))
+(def valid-clojure-version? (partial valid-version? :clojure))
 
 ;; Exists for the sole purpose of modifying the current project's metadata.
 ;; See https://github.com/technomancy/leiningen/blob/master/doc/PLUGINS.md#project-middleware
 (defn middleware
   [{:keys [dependencies exclusions] :as project}]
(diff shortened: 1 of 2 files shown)


bbatsov:
Hmm, I wonder if there's some way to figure out which is the current profile, as now we might end up in the opposite situation if someone tries to run a profile using say Clojure 1.6.

dpsutton:
I thought i could use something from lein to introspect the actual computed deps but I'm not sure anything like that exists. The thing I was using would throw errors if deps weren't already downloaded and it kept barfing because clojure 1.6.0 wasn't in the travis container.

This seems like better than current although i take your point @bbatsov about overriding a 1.10 with a profile of 1.5 or similar.

PR #604, opened 2019-03-25
Upgrade mranderson to 0.5.0

benedekfazekas:
mranderson 0.5.x specific config added to the project file as
well. **Resolved tree** mode is used. This means that
mranderson works with a resolved dependency tree and does not create a
deeply nested directory structure, only adds prefixes to all
dependencies. Handles all depedendencies as first level ones and
processes them in a topological order.

Add project prefix in the project file for mranderson, also some
config for **unresolved tree** if that would get used in the near
future.

Breaking changes in mranderson are followed up in Makefile and README

- [x] All tests are passing

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -9,18 +9,19 @@
                  ^:source-dep [cider/orchard "0.4.0"]
                  ^:source-dep [thunknyc/profile "0.5.2"]
                  ^:source-dep [mvxcvi/puget "1.1.0"]
-                 ^:source-dep [fipp "0.6.15"]
+                 ^:source-dep [fipp "0.6.15"]; can be removed in unresolved-tree mode
                  ^:source-dep [compliment "0.3.8"]
                  ^:source-dep [cljs-tooling "0.3.1"]
                  ^:source-dep [cljfmt "0.6.4" :exclusions [org.clojure/clojurescript]]
-                 ;; Not used directly in cider-nrepl, but needed because of tools.namespace
-                 ;; and the way MrAnderson processes dependencies
-                 ;; See https://github.com/clojure-emacs/cider/issues/2176 for details
-                 ^:source-dep [org.clojure/java.classpath "0.3.0"]
                  ^:source-dep [org.clojure/tools.namespace "0.3.0-alpha4"]
                  ^:source-dep [org.clojure/tools.trace "0.7.10"]
                  ^:source-dep [org.clojure/tools.reader "1.2.2"]]
-  :plugins [[thomasa/mranderson "0.4.9"]]
+  :plugins [[thomasa/mranderson "0.5.0"]]
+  :mranderson {:project-prefix "cider.nrepl.inlined-deps"
+               :overrides       {[mvxcvi/puget fipp] [fipp "0.6.15"]}; only takes effect in unresolved-tree mode
+               :expositions     [[mvxcvi/puget fipp]]; only takes effect unresolved-tree mode
+               :unresolved-tree false}
+
   :exclusions [org.clojure/clojure]
 
   :filespecs [{:type :bytes :path "cider/cider-nrepl/project.clj" :bytes ~(slurp "project.clj")}]
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 dist: xenial
 language: clojure
-lein: 2.8.3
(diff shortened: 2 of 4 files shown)


project.clj
```diff
@@ -9,18 +9,19 @@
...
-                 ;; See https://github.com/clojure-emacs/cider/issues/2176 for details
-                 ^:source-dep [org.clojure/java.classpath "0.3.0"]
                  ^:source-dep [org.clojure/tools.namespace "0.3.0-alpha4"]
                  ^:source-dep [org.clojure/tools.trace "0.7.10"]
                  ^:source-dep [org.clojure/tools.reader "1.2.2"]]
-  :plugins [[thomasa/mranderson "0.4.9"]]
+  :plugins [[thomasa/mranderson "0.5.0"]]
+  :mranderson {:project-prefix "cider.inlined-deps"
+               :overrides       {[mvxcvi/puget fipp] [fipp "0.6.15"]}; only takes effect in unresolved-tree mode
+               :expositions     [[mvxcvi/puget fipp]]; only takes effect unresolved-tree mode
```
bbatsov: Btw, shouldn't this be named `exposes` to be consistent with `overrides`. I guess it's consistent in a way even now, but I've never heard expositions used in this sense. :-)

project.clj
```diff
@@ -9,18 +9,19 @@
...
-                 ;; Not used directly in cider-nrepl, but needed because of tools.namespace
-                 ;; and the way MrAnderson processes dependencies
-                 ;; See https://github.com/clojure-emacs/cider/issues/2176 for details
-                 ^:source-dep [org.clojure/java.classpath "0.3.0"]
                  ^:source-dep [org.clojure/tools.namespace "0.3.0-alpha4"]
                  ^:source-dep [org.clojure/tools.trace "0.7.10"]
                  ^:source-dep [org.clojure/tools.reader "1.2.2"]]
-  :plugins [[thomasa/mranderson "0.4.9"]]
+  :plugins [[thomasa/mranderson "0.5.0"]]
+  :mranderson {:project-prefix "cider.inlined-deps"
```
bbatsov: I think this should be `cider.nrepl.inlined-deps`.

project.clj
benedekfazekas: as much as I googled it seemed that the noun form is `exposition`. eg something being exposed is an exposition (or an exposure but that has negative connotations).

bbatsov: Well, I can live with the name. :-) My actually confusion was from the fact that `overrides` can be interpreted both as a verb and a noun and first I thought it was a verb. No a big deal.

project.clj
```diff
@@ -9,18 +9,19 @@
...
-                 ^:source-dep [org.clojure/java.classpath "0.3.0"]
                  ^:source-dep [org.clojure/tools.namespace "0.3.0-alpha4"]
                  ^:source-dep [org.clojure/tools.trace "0.7.10"]
                  ^:source-dep [org.clojure/tools.reader "1.2.2"]]
-  :plugins [[thomasa/mranderson "0.4.9"]]
+  :plugins [[thomasa/mranderson "0.5.0"]]
+  :mranderson {:project-prefix "cider.nrepl.inlined-deps"
+               :overrides       {[mvxcvi/puget fipp] [fipp "0.6.15"]}; only takes effect in unresolved-tree mode
+               :expositions     [[mvxcvi/puget fipp]]; only takes effect unresolved-tree mode
+               :unresolved-tree false}
```
bbatsov: Unrelated comment - why is this needed? Wasn't `:resolved-tree` the default anyways?

bbatsov:
Thanks for all the work you've put in the new release!

project.clj
benedekfazekas: it is not really needed. but for now i left it like that with the above options that are not needed in resolved tree mode either to make it easier to switch to unresolved tree mode in the near future if the need arises. if not everything apart from the project prefix can be removed.

benedekfazekas:
cheers :)

project.clj
bbatsov: My point was that I thought the default would be `:resolved-deps true` and its opposite would be `:resolved-tree false`. Might have been a bit better if this was `:mode :(un)resolved-tree`, but as long as it gets the job done anything goes.

PR #605, opened 2019-03-27
Add a option for filtering vars to the ns-vars middleware

### LINKED ISSUE #566: Should ns-vars-with-meta-clj return public vars?
## Question
*  `ns-vars-with-meta-clj` uses `ns-interns` and returns all vars including private vars
   * Should `ns-vars-with-meta-clj` use `ns-publics` ?

## Actual behavior
* `ns-vars-clj` uses `ns-publics`
  * https://github.com/clojure-emacs/cider-nrepl/blob/b13881e18eb332619f2c87c5a6d6637f28442c77/src/cider/nrepl/middleware/ns.clj#L20
* `ns-vars-cljs` uses `cljs-analysis/public-vars`
  * https://github.com/clojure-emacs/cider-nrepl/blob/b13881e18eb332619f2c87c5a6d6637f28442c77/src/cider/nrepl/middleware/ns.clj#L40
* `ns-vars-with-meta-cljs` uses `cljs-analysis/public-var`
  * https://github.com/clojure-emacs/cider-nrepl/blob/b13881e18eb332619f2c87c5a6d6637f28442c77/src/cider/nrepl/middleware/ns.clj#L47
* but `ns-vars-with-meta-clj` uses `ns-interns`
  * https://github.com/clojure-emacs/cider-nrepl/blob/b13881e18eb332619f2c87c5a6d6637f28442c77/src/cider/nrepl/middleware/ns.clj#L27

### cider-nrepl version

0.18.0

liquidz:
Fix #566.

### DIFF
--- a/src/cider/nrepl/middleware/ns.clj
+++ b/src/cider/nrepl/middleware/ns.clj
@@ -2,12 +2,14 @@
   (:refer-clojure :exclude [ns-aliases])
   (:require
    [cider.nrepl.middleware.util.cljs :as cljs]
+   [cider.nrepl.middleware.util.coerce :as util.coerce]
    [cider.nrepl.middleware.util.error-handling :refer [with-safe-transport]]
    [cider.nrepl.middleware.util.meta :as um]
    [cljs-tooling.info :as cljs-info]
    [cljs-tooling.util.analysis :as cljs-analysis]
    [orchard.misc :as u]
-   [orchard.namespace :as ns]))
+   [orchard.namespace :as ns]
+   [orchard.query :as query]))
 
 (defn ns-list-vars-by-name
   "Return a list of vars named `name` amongst all namespaces.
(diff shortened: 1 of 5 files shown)


src/cider/nrepl/middleware/ns.clj
```diff
@@ -17,39 +17,51 @@
...
-       sort))
-
-(defn ns-vars-with-meta-clj [ns]
-  (->> (symbol ns)
-       ns-interns
-       (u/update-vals (comp um/relevant-meta meta))
-       (u/update-keys name)
-       (into (sorted-map))))
+(defn ns-vars-clj [ns & [include-privates?]]
+  (let [fetch-vars (if include-privates?
```
bbatsov: We can also use `var-query` here directly. That would simplify a bit the code.

bbatsov:
The PR looks good overall. I mostly wonder if we should try to use here the var-query API, as this would give more flexibility to the ns filtering in general.

src/cider/nrepl/middleware/ns.clj
liquidz: Thank you for reviewing!
I missed `var-query` API. I'll fix.

bbatsov: We don't have it yet (something that @arichiardi can probably tackle), but I think we can still adopt the same API and just "fake" it on the cljs side for now (by ignoring everything but the private filter).

liquidz: @bbatsov OK! I fixed to take `var-query` option for filtering vars.
On the cljs side, it only support `private?` query currently.

bbatsov:
Fantastic work! Thanks! 🙇

arichiardi:
A newbie question on this, what is the difference between this `ns-vars` middleware and the information returned by `orchard/info`?

The kind of seem redundant but maybe `ns-vars` does something different?

bbatsov:
I was referring to https://github.com/clojure-emacs/orchard/blob/master/src/orchard/query.clj How is `info` related to `ns-vars`?

arichiardi:
Oh ok it is a different use case...I was just wondering if `info` can provide the same kind of information there. I did not even know we had such a feature :smile: Sorry for the noise here :smile:

PR #611, opened 2019-05-24
plugin: ensure there is a version to check to begin with

pyr:
When using dependency supplying modules such as `lein-modules`,
by the time cider-nrepl does its version check it has no indication
of version.

In this case, the code throws leaving no chance to the REPL to
be spun up.

When versions are unavailable, assume the project knows what it
is doing instead and discard it from the version check altogether.

### DIFF
--- a/src/cider_nrepl/plugin.clj
+++ b/src/cider_nrepl/plugin.clj
@@ -22,7 +22,8 @@
         clojure-versions    (when-not clojure-excluded?
                               (->> dependencies
                                    (keep (fn [[id version & _]]
-                                           (when (= id 'org.clojure/clojure)
+                                           (when (and (= id 'org.clojure/clojure)
+                                                      (string? version))
                                              version)))))
         clojure-version-ok? (cond clojure-excluded?
                                   ;; In this case the onus is on the user. A warning will be emitted
--- a/src/cider_nrepl/plugin.clj
+++ b/src/cider_nrepl/plugin.clj
@@ -23,6 +23,10 @@
                               (->> dependencies
                                    (keep (fn [[id version & _]]
                                            (when (and (= id 'org.clojure/clojure)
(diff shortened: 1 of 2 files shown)


src/cider_nrepl/plugin.clj
```diff
@@ -22,7 +22,8 @@
         clojure-versions    (when-not clojure-excluded?
                               (->> dependencies
                                    (keep (fn [[id version & _]]
-                                           (when (= id 'org.clojure/clojure)
+                                           (when (and (= id 'org.clojure/clojure)
+                                                      (string? version))
```
bbatsov: I'd add some comment here explaining the need for the extra check.

test/clj/cider/nrepl/plugin_test.clj
```diff
@@ -0,0 +1,42 @@
...
+      cider.nrepl/wrap-spec
+      cider.nrepl/wrap-stacktrace
+      cider.nrepl/wrap-test
+      cider.nrepl/wrap-trace
+      cider.nrepl/wrap-tracker
+      cider.nrepl/wrap-undef
+      cider.nrepl/wrap-version
+      cider.nrepl/wrap-xref]}})
+
+(deftest version-checks
```
bbatsov: I guess you can add some test(s) about the normal scenarios as well.

src/cider_nrepl/plugin.clj
pyr: Good point, I added one.

test/clj/cider/nrepl/plugin_test.clj
pyr: Done as well

pyr:
Thanks for the review!

bbatsov:
Thanks!

PR #614, opened 2019-06-02
Fix cider.tasks/nrepl-server

arichiardi:
It had been broken after the removal of `cider-nrepl.main`. The functionality is
now restored by delegating to `nrepl.cmdline/dispatch-commands`. Probably some
renaming and/or refactoring is necessary in nrepl for programmatic use.

- [x] All tests are passing

### DIFF
--- a/src/cider/tasks.clj
+++ b/src/cider/tasks.clj
@@ -30,7 +30,7 @@
   (let [default-mws @@(resolve 'boot.repl/*default-middleware*)]
     (util/dbug* "nREPL middleware: %s\n" (vec default-mws))
     (boot.core/with-pass-thru [_]
-      (require 'cider-nrepl.main)
-      ((resolve 'cider-nrepl.main/start-nrepl) {:middleware default-mws
-                                                :port port
-                                                :bind bind}))))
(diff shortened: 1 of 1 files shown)


bbatsov:
Hmm, I hadn't noticed that the old namespace was used in the boot task. My bad! 

Do we even need this task anymore? Boot has been shipping the new nREPL for a while now and I recall we added it as a workaround when boot was still bundling tools.nrepl. I guess one advantage of using it is that it honours nREPL's config file, but I don't recall if it doesn't anything extra.

arichiardi:
I have to say I quickly fixed it because I am using `master` everywhere and I had to do some work on a project today :smile: 

The task reads the middlewares (which are globally modifiable by any task) and calls the right function from `nrepl` now. If there is a `boot` built-in task that does that we do not need our own. Sorry stating the obvious :smile: ...mainly because the I am not sure and I would need to investigate.

arichiardi:
Relevant link: https://github.com/boot-clj/boot/blob/master/boot/core/src/boot/task/built_in.clj#L448

arichiardi:
Superseded by #615

PR #617, opened 2019-06-16
Move resource implementation to orchard

arichiardi:
This patch moves the code for the resource op down into the orchard.

- [X] All tests are passing

### DIFF
--- a/src/cider/nrepl/middleware/stacktrace.clj
+++ b/src/cider/nrepl/middleware/stacktrace.clj
@@ -10,10 +10,10 @@
    [nrepl.misc :refer [response-for]]
    [nrepl.transport :as t]
    [orchard.info :as info]
-   [orchard.resource :as resource]
    [orchard.java :as java]
    [orchard.misc :as u]
-   [orchard.namespace :as namespace])
+   [orchard.namespace :as namespace]
+   [orchard.resource :as resource])
   (:import
    (clojure.lang Compiler$CompilerException)
    (java.io StringWriter)))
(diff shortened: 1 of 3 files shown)


bbatsov:
The changes look ok to me, but I'll wait for us to finish the Orchard changes first.

src/cider/nrepl/middleware/stacktrace.clj
```diff
@@ -10,10 +10,10 @@
    [nrepl.misc :refer [response-for]]
    [nrepl.transport :as t]
    [orchard.info :as info]
-   [orchard.resource :as resource]
    [orchard.java :as java]
    [orchard.misc :as u]
-   [orchard.namespace :as namespace])
+   [orchard.namespace :as namespace]
+   [orchard.resource :as resource])
```
arichiardi: this now is sorted

test/clj/cider/nrepl/middleware/resource_test.clj
```diff
@@ -19,17 +18,3 @@
                        (:resources-list response))))
       (is (seq (filter #(re-matches #".*test/resources/test\.txt" (:file %))
                        (:resources-list response)))))))
-
-(deftest resource-op-error-handling-test
```
bbatsov: Probably we should keep the error handling tests.

arichiardi: Uhm, I have checked that and it is `redef`-ing an api that is in `orchard` so I am not sure it still makes sense to test it?

bbatsov: The point of those tests was not to test something in the API, but that unexpected errors are properly handled. Basically it tests that the middleware is using `with-safe-transport`. As if you've inlined the reply functions that's a bit hard to do today, though. Anyways, not a big deal.

arichiardi: Oh I see, good to know, it would be good to test these cross cutting things in some more generic way but yeah .. not easy with middlewares maybe? Dunno in any case thanks for merging!

PR #618, opened 2019-06-23
Ensure ns-query map is built correctly

dpsutton:
The required structure for the 'query/var' function is `{:var-query
{:ns-query {...}}`. The older implementation did not grab those
keywords from the cider message and put them at the correct
level. they were left top level and then missed by the
`query/namespace` function.

In addition to this, there was a bug in orchard which had the wrong
prefix for inlined deps so these were not omitted in _2_ places: both
in the client-side specification of which namespaces to ignore and
orchard's own mechanism to elide cider's internal namespaces.

See related:
- orchard https://github.com/clojure-emacs/orchard/pull/59/
- CIDER

- [ ] All tests are passing

Thanks!

### DIFF
--- a/test/clj/cider/nrepl/middleware/apropos_test.clj
+++ b/test/clj/cider/nrepl/middleware/apropos_test.clj
@@ -1,6 +1,6 @@
 (ns cider.nrepl.middleware.apropos-test
   (:require
-   [cider.nrepl.middleware.apropos :refer [apropos]]
+   [cider.nrepl.middleware.apropos :refer [apropos] :as apropos]
    [cider.nrepl.test-session :as session]
    [clojure.string :as str]
    [clojure.test :refer :all]))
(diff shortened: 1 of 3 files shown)


test/clj/cider/nrepl/middleware/apropos_test.clj
```diff
@@ -9,6 +9,14 @@
 
 (use-fixtures :each session/session-fixture)
 
+(deftest msg->var-query-map-test
+  (testing "Constructs the ns-query map correctly"
+    (let [msg {:exclude-regexps ["^cider.nrepl" "^refactor-nrepl" "^nrepl"]
+               :query "spelling"}
+          query-map (#'apropos/msg->var-query-map msg)]
+      (is (contains? (:var-query query-map) :ns-query))
```
bbatsov: Won't it be better to just compare the result with the expected map?

dpsutton: i thought about that but i think the returned map will have regex's rather than strings for the exclude-regexps and I wasn't sure of the semantics of comparing regexes

dpsutton: and it appears they compare by reference. `(= #"bob" #"bob")` is false. so it would require some wrangling that i wasn't into at the time :) I can take another crack at it later if you like.

bbatsov: Got it. It's fine like this.

PR #619, opened 2019-06-23
Fix ci

dpsutton:

- [x] All tests are passing

Fixing up the tests. One test had an error and two simple cljfmt problems. All tests pass for me locally but from some other runs it looks like two more might fail on CI. Waiting to see.

### DIFF
--- a/src/cider/nrepl/middleware/xref.clj
+++ b/src/cider/nrepl/middleware/xref.clj
@@ -19,7 +19,7 @@
 
 (defn fn-refs-reply [{:keys [ns symbol]}]
   (let [var (ns-resolve (misc/as-sym ns) (misc/as-sym symbol))]
-   {:fn-refs (map xref-data (xref/fn-refs var))}))
+    {:fn-refs (map xref-data (xref/fn-refs var))}))
 
 (defn fn-deps-reply [{:keys [ns symbol]}]
   (let [var (ns-resolve (misc/as-sym ns) (misc/as-sym symbol))]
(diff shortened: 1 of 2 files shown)


dpsutton:
quite confusing. Circle is failing everything but if I run
`docker run -it --ulimit nofile=122880:122880 -m 3G -v $(pwd):/src -u node circleci/clojure:openjdk-8-lein-2.9.1-node` it passes just fine.

dpsutton:
Ok. I have figured out how to fail the test locally.

The config we have is 2.1 which the circleci cli cannot transform. You have to grab the configuration off of circle (which has been transformed to 2.0) from [the configuration tab](https://circleci.com/gh/clojure-emacs/cider-nrepl/413#config/containers/0).

Paste this over the local copy in .circleci/config.yaml and then you can run `~/bin/circleci local execute --job "Java 8, Clojure 1.8"` locally to see the failure.

(note ~/bin/circleci is my install location because the default install failed :( )

dpsutton:
```

**Comment by dpsutton** [comment]
I just solved the last issue. Had to do with a var from compliment

`(def prefix-placeholder '__prefix__)`

that was choking up the `as-val` lookup:

```clj
orchard.xref> (var-get (find-var '__prefix__))
Execution error (IllegalArgumentException) at orchard.xref/eval8444 (form-init3787013978346022571.clj:769).
Symbol must be namespace-qualified
```

This is incredibly weird but you can investigate by copying the circleCI config from the config tab and then  running `~/bin/circleci local execute --job "Java 8, Clojure 1.8"`

Long and short is that once the PR in orchard is accepted we are all green when we bump our dep.

bbatsov:
Thanks!

PR #622, opened 2019-06-26
Fix dynamic font locking for fns defined with def in cljs

oskarkv:
I have tried to fix an issue with some functions not being properly dynamically font-locked in ClojureScript.

As you can see in the diff, I have used the tags generated by, as far as I understand, `cljs.analyzer`. The following screen shot shows the result, as well as the tags given to the various definitions. You can see in the bottom vector the result, where of course functions are colored with green, and other values with orange.

Results

Previously, only `fn1` and `fn5` would have been recognized as functions (and of course `comp`). As you can see, `fn7`, `fn8` and `fn8` are not recognized, but that is not something we can fix in CIDER, I guess.

I wrote made this pull request to get feedback---what do you think about this? Also, is there anything else I should do, such as adding a line in `CHANGELOG.md`? If so, should I add to cider's chanelog, cider-nrepl's changelog, or both? I was a bit unsure about that, which is why I haven't done it.

### DIFF
--- a/src/cider/nrepl/middleware/track_state.clj
+++ b/src/cider/nrepl/middleware/track_state.clj
@@ -40,10 +40,14 @@
                               [sym (um/relevant-meta the-meta)]))))))))
 
 (defn- cljs-meta-with-fn
-  "Like (:meta m) but adds {:fn true} if (:fn-var m) is true."
+  "Like (:meta m) but adds {:fn true} if (:fn-var m) is true, (:tag m) is
+  'function, or if m does not have a :tag key."
   [m]
   (cond-> (:meta m)
-    (:fn-var m) (assoc :fn true)))
+    (or (:fn-var m)
+        (= (:tag m) 'function)
+        (not (contains? m :tag)))
+    (assoc :fn true)))
 
 ;;; Namespaces
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ### Bugs fixed
 
+* Fix some functions defined with `def` not being properly font locked when using dynamic font locking for ClojureScript.
 * [#618](https://github.com/clojure-emacs/cider-nrepl/pull/618): Fix apropos to honor exclude-regexps to filter out namespaces by regex
 * [#605](https://github.com/clojure-emacs/cider-nrepl/pull/605): Fix `ns-vars-with-meta` to return public vars.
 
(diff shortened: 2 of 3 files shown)


bbatsov:
Looks like a great improvement to me. There should be more explanations about those tags, though, as it's not really obvious what they are. Maybe we can also ping @mfikes to advise us if this covers everything we need to consider. 

As for the changelog - this should be only in `cider-nrepl`'s changelog.

bbatsov:
You should also look into those failing tests on Java 11.

oskarkv:
I have no idea about those failing tests. I have only touched `track-state-test`/`track-state`, which is passing. And on my machine that `info-test` fails if I checkout master, without my changes.

oskarkv:
Oh, I'm a bit unsure, is this a "new feature", "bug fix" or "change"?

dpsutton:
Rebase onto master. Those tests are not your fault and have already been fixed. It's the silly javadoc prefix test.

oskarkv:
Should I do anything else?

src/cider/nrepl/middleware/track_state.clj
```diff
@@ -40,10 +40,14 @@
...
 (defn- cljs-meta-with-fn
-  "Like (:meta m) but adds {:fn true} if (:fn-var m) is true."
+  "Like (:meta m) but adds {:fn true} if (:fn-var m) is true, (:tag m) is
+  'function, or if m does not have a :tag key."
   [m]
   (cond-> (:meta m)
-    (:fn-var m) (assoc :fn true)))
+    (or (:fn-var m)
+        (= (:tag m) 'function)
+        (= (get m :tag :no-tag) :no-tag))
```
Malabarba: Normally this would be `(not (contains? m :tag))`.
If that doesn't work for some reason (if `m` is not a collection) the ideally correct way to do this would be with `gensym` (so there's zero risk of mistaking `{}` with `{:tag :no-tag}`).

``` clojure
(let [no-tag (gensym)]
  (cond-> (meta m)
    (or (:fn-var m)
        (= (:tag m) 'function)
        (= (get m :tag no-tag) no-tag))
    (assoc :fn true)))
```

Malabarba:
> Oh, I'm a bit unsure, is this a "new feature", "bug fix" or "change"?


Looks like a bugfix to me. :-)

src/cider/nrepl/middleware/track_state.clj
oskarkv: Ah, of course. I forgot `contains?` exists. Thanks!

bbatsov:
Thanks!

PR #625, opened 2019-07-12
Dynamically font lock multifns as fns in CljS

oskarkv:
I discovered that multifns have a tag I didn't account for earlier, in #622


bbatsov:
Thanks!

PR #626, opened 2019-07-15
Fix content-types.edn shadowing resource with the same name

lukaszkorecki:
Hi! 

All credit goes to @AfonsoTsukamoto - he run into this issue and spent time debugging it.

One of our applications ships with `content-types.edn` file, and that clashes with the one provided with Cider causing hard to debug errors. Moving `content-types.edn` provided by cider-nrepl to a different location fixes the issue.


bbatsov:
Nice catch! We probably have a similar problem with the `see-also.edn` in Orchard. By now I had totally forgotten that Java resources were supposed to be namespaced. :D

PR #628, opened 2019-07-23
Add a middleware to find docs from ClojureDocs

liquidz:
Introduce a middleware to find docs from not Grimoire but ClojureDocs.
See also https://github.com/clojure-emacs/cider/issues/2663 and https://github.com/clojure-emacs/orchard/pull/64

- [x] All tests are passing


src/cider/nrepl.clj
```diff
@@ -478,12 +478,32 @@
               :returns {"fn-deps" "A list of function deps."
                         "status" "done"}}}})
 
+(def-wrapper wrap-clojuredocs cider.nrepl.middleware.clojuredocs/handle-clojuredocs
+  {:doc "Middleware to find a documents from ClojureDocs."
+   :handles {"load-clojuredocs-cache!"
```
bbatsov: I don't think clients will need to call this explicitly, probably they'll just need the ability to reload the cache.

src/cider/nrepl.clj
```diff
@@ -478,12 +478,32 @@
...
+             {:doc "Load exported documents file from ClojureDocs, and store it as a cache."
+              :requires {}
+              :optional {"export-edn-url" "EDN file URL exported from ClojureDocs. Defaults to \"https://clojuredocs-edn.netlify.com/export.edn\"."}
+              :returns {"status" "\"ok\" if loading was successful"}}
+             "clean-clojuredocs-cache!"
```
bbatsov: Same with this one - probably we can expose a single "clojuredocs-refresh-cache"  op instead of two.

src/cider/nrepl.clj
```diff
@@ -478,12 +478,32 @@
...
+             {:doc "Clean a cached file and documents"
+              :requires {}
+              :returns {"status" "\"ok\" if cleaning was successful"}}
+             "clojuredocs"
```
bbatsov: I think `clojuredocs-lookup` is a more informative name for this one.

Makefile
```diff
@@ -6,13 +6,16 @@ export CLOVERAGE_VERSION = 1.0.13
...
+	curl -o $@ https://clojuredocs-edn.netlify.com/export.edn
+
 .inline-deps:
 	lein inline-deps
 	touch .inline-deps
 
 inline-deps: .inline-deps
 
-test: .inline-deps
+test: .inline-deps test/resources/clojuredocs/export.edn
```
bbatsov: It's best to put resources under the `cider/nrepl` package for consistency with the existing resources.

src/cider/nrepl.clj
liquidz: Apart from refreshing cache, I thought it could be used for cache loading in the background (e.g. when connecting nREPL).
But once the cache file is generated, loading of the cache is fast enough, so it seems that just refreshing is good.

liquidz:
@bbatsov Thank you for your reviewing!
I had forgotten to update README, so I fixed them including README!

bbatsov:
Excellent! I'll cut a new beta now.

PR #629, opened 2019-07-26
Move transform-value from the orchard to cider-nrepl

arichiardi:
The transform-value function is moved here to cider-nrepl, where it belongs, as
it seems to be mainly related to presentation only.

PR friend of https://github.com/clojure-emacs/orchard/pull/66 

- [X] All tests are passing


bbatsov:
I think an `output` ns is a bit misleading as it generally makes you think about console/file output. This used to be in the main util ns in the past and probably we should add it back there.

arichiardi:
Ok I will put it there then, thank you!

arichiardi:
Ok addressed the comment.

bbatsov:
Thanks!

PR #633, opened 2019-08-15
dynamic cljs completions with suitable

rksm:

- [x] All tests are passing

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -389,6 +389,7 @@ Let's also acknowledge some of the projects leveraged by cider-nrepl:
 * [orchard][] - extracted from `cider-nrepl`, so that non-nREPL clients can leverage the generic tooling functionality (like `inspect`, `apropos`, `var-info`, etc
 * [compliment][] - for Clojure code completion
 * [cljs-tooling][] - for ClojureScript code completion
+* [suitable][] - for ClojureScript code completion using runtime inspection
 * [tools.trace][] - for tracing
 * [tools.namespace][] - for namespace reloading
 * [cljfmt][] - for code formatting
@@ -402,6 +403,7 @@ Distributed under the Eclipse Public License, the same as Clojure.
 [orchard]: https://github.com/clojure-emacs/orchard
 [compliment]: https://github.com/alexander-yakushev/compliment
 [cljs-tooling]: https://github.com/clojure-emacs/cljs-tooling
+[suitable]: https://github.com/rksm/clj-suitable
 [tools.trace]: https://github.com/clojure/tools.trace
 [tools.namespace]: https://github.com/clojure/tools.namespace
 [cljfmt]: https://github.com/weavejester/cljfmt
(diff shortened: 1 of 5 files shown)


src/cider/nrepl/middleware/complete.clj
```diff
@@ -5,16 +5,22 @@
...
    [compliment.utils :as jvm-complete-utils]
-   [orchard.misc :as u]))
+   [orchard.misc :as u]
+   [suitable.complete-for-nrepl :as suitable]))
+
+(defn- cljs-complete
+  [msg cljs-env ns prefix extra-metadata]
+  (concat (cljs-complete/completions cljs-env prefix {:context-ns ns
+                                                      :extra-metadata extra-metadata})
+          (suitable/complete-for-nrepl msg)))
```
bbatsov: Why does `suitable` need the entire `msg`? I'm asking mostly because it seems to me that all the other params should be enough to get candidates.

bbatsov:
We should also mention this in the changelog, as it's certainly a noteworthy change.

README.md
```diff
@@ -402,6 +403,7 @@ Distributed under the Eclipse Public License, the same as Clojure.
 [orchard]: https://github.com/clojure-emacs/orchard
 [compliment]: https://github.com/alexander-yakushev/compliment
 [cljs-tooling]: https://github.com/clojure-emacs/cljs-tooling
+[suitable]: https://github.com/rksm/suitable
```
Olical:
```suggestion
[suitable]: https://github.com/rksm/clj-suitable
```

src/cider/nrepl/middleware/complete.clj
rksm: It needs access to the cljs options + repl env + compiler env. We can pull those out and just pass them along but it seemed simpler to hand in the msg.

rksm:
> I would use another name though for the namespace. `cljs-complete` is/will be already included in `compliment` and `suitable` will be providing `js-objects` completions only.

What exactly do you mean? `suitable/complete-for-nrepl`? Or the `cljs-complete` function name?

README.md
rksm: Thanks!

src/cider/nrepl/middleware/complete.clj
bbatsov: Got it. As this is something minor I guess there's no need to fret over it too much.

bbatsov:
I guess he meant the `cljs-complete`, as we also have this as an ns alias right now. Maybe there's also such a function in `compliment` now, but I don't think we need to change the name as the prefix will be enough to tell them apart. Plus, we can always change this down the road.

arichiardi:
Well yes you are right, I can actually change it with my integration PR or something

arichiardi:
I meant the function name but no probs, it can be tuned later on.

Issue #648, opened 2019-09-20
Make it possible to disable suitable with a flag passed to the complete op


bbatsov:
As `suitable` is still under active development and there are some problems discovered by users it'd be great if they could turn it off if they experience problems. 

I was thinking we should add some flag like `enhanced-cljs-completion` or something along those lines that clients will have to set to true to enable suitable completions. It'd be true in CIDER and I'll leave it to others to decide how to handle this. Generally there hasn't been that many complaints, so it seems that suitable works reasonably well, but it'd be nice to give a way out to the people who run into problems. 

Other ideas are welcome as well, of course. This just seems like the simplest solution to me. 

//cc @rksm @dpsutton sutton @pez @arichiardi

PEZ:
A flag to send with the completion request works for me. I'll surface it as an opt-out setting in Calva.

FYI. No complaints from Calva users so far, about problems with suitable.

bbatsov:
> So one way to implement this is to attach an additional field to completion requests such. This can then simply be tested for in cider.nrepl.middleware.complete/cljs-complete (which just does a concatenation of completion results anyway`.

That's exactly what I had in mind.

> The downside is that it adds more "noise" to the protocol and one more detail for the client to implement.

I'm not very worried about this, as probably most clients are not going to implement it anyways. That's going to be a temporary workaround that will remove at some point.

> If this is OK, I can create a PR to implement that.

That'd be great. Thanks!

bbatsov:
Fixed in #649.

PR #649, opened 2019-09-21
add enhanced-cljs-completion flag

rksm:
to make suitable cljs completions optional

https://github.com/clojure-emacs/cider-nrepl/issues/648

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,16 @@
 
 ## master (unreleased)
 
+### New features
+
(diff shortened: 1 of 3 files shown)


bbatsov:
I see this broke the completion tests. :-) Might be a good idea to add one unit test for the change as well.

rksm:
Of course, sorry for that.

bbatsov:
No worries! Thanks for tackling this! 🙇

PR #650, opened 2019-09-23
Fix execution order for (is (= a b c))

jsab:
The current implementation of `=-body` evaluates the rest of the arguments to `=` before evaluating the first argument. This causes problems in the presence of side-effects.


bbatsov:
Thanks!

PR #652, opened 2019-10-01
Respect default *print-length* and *print-level* in debugger output

yuhan0:
Fixes https://github.com/clojure-emacs/cider/issues/2621

I'm not familiar at all with the internals of `cider-nrepl`, but having been annoyed by the above issue one too many times I decided to grep for `*print-length*` in this repo and found that simply changing these lines solved the issue. 

Perhaps there's a better way to tackle it, hopefully this PR gives a starting point :)

- [x] All tests are passing

### DIFF
--- a/src/cider/nrepl/middleware/debug.clj
+++ b/src/cider/nrepl/middleware/debug.clj
@@ -182,8 +182,10 @@ this map (identified by a key), and will `dissoc` it afterwards."}
 (defn pr-short
   "Like `pr-str` but limited in length and depth."
   [x]
-  (binding [*print-length* (:length @print-options)
-            *print-level*  (:level @print-options)]
+  (binding [*print-length* (or (:length @print-options)
+                               *print-length*)
+            *print-level*  (or (:level @print-options)
+                               *print-level*)]
     ;; TODO: Make it possible to use a random print function here
     (pr-str x)))
 


bbatsov:
We should track down why the `print-options` are `nil` in the debugger, as I'm pretty sure something's set by default in CIDER (and they were supposed to fallback to `*print-length*` and friends anyways). Mostly likely something is not passed/set properly somewhere. Btw, the print options come straight from nREPL - they are not something specific to `cider-nrepl`.

src/cider/nrepl/middleware/debug.clj
```diff
@@ -182,8 +182,10 @@ this map (identified by a key), and will `dissoc` it afterwards."}
 (defn pr-short
   "Like `pr-str` but limited in length and depth."
   [x]
-  (binding [*print-length* (:length @print-options)
-            *print-level*  (:level @print-options)]
+  (binding [*print-length* (or (:length @print-options)
```
bbatsov: I'm wondering why `print-options` is some var, as if I recall correctly it's passed as part of each nREPL message. Most likely it's not initialized properly. Perhaps we broke something when we introduced support for nREPL 0.6.

bbatsov:
Just as I figured the problem is that what we current sent to init the debugger is not what the debugger expects:

```
(-->
  id                             "7"
  op                             "init-debugger"
  session                        "da8bc456-b029-4d88-b295-3fe4189f2a3c"
  time-stamp                     "2019-10-03 09:16:08.460679000"
  nrepl.middleware.print/options (dict
                                   right-margin 80)
  nrepl.middleware.print/print   "cider.nrepl.pprint/pprint"
  nrepl.middleware.print/quota   1048576
  nrepl.middleware.print/stream? "1"
)
```

On the other hand the debugger does the initialization like this:

```
(defn- initialize
  "Initialize the channel used for debug-input requests."
  [{:keys [print-options] :as msg}]
  (when (map? @debugger-message)
    (debugger-send :status :done))
  ;; The above is just bureaucracy. The below is important.
  (reset! @#'print-options print-options)
  (reset! debugger-message msg))
```

But there's no key named `print-options` and that's the real problem. That will be a simple fix, but probably your suggestions also makes sense in the absence of some explicit print options.

PR #653, opened 2019-10-12
Add inspect-def-current-value to inspect middleware

jpmonettas:
This PR adds inspect-def-current-value that will enable cider-inspector to define a var in your current namespace with a given name and current inspector value.

- [x] All tests are passing


bbatsov:
Thanks!

PR #654, opened 2019-10-16
Change representation of debug-commands to a set of command names

yuhan0:
Note: This is a breaking change that sends debug commands to the client as an unsorted
list of command names without key mapping information, as a means of separating
concerns.

Quick draft of the changes suggested in https://github.com/clojure-emacs/cider/pull/2731
If everything looks alright I'll go ahead with adding changelog notes and tests

- [x] All tests are passing

### DIFF
--- a/src/cider/nrepl/middleware/debug.clj
+++ b/src/cider/nrepl/middleware/debug.clj
@@ -297,19 +297,20 @@ this map (identified by a key), and will `dissoc` it afterwards."}
                               last :stacktrace)}]}))
 
 (def debug-commands
-  {"c" :continue
-   "C" :Continue
-   "e" :eval
-   "h" :here
-   "i" :in
-   "j" :inject
-   "l" :locals
-   "n" :next
-   "o" :out
-   "p" :inspect
-   "q" :quit
-   "s" :stacktrace
-   "t" :trace})
+  "An unsorted set of commands supported by the debugger."
+  #{:continue
+    :Continue
+    :eval
+    :here
+    :in
+    :inject
+    :locals
+    :next
+    :out
+    :inspect
+    :quit
+    :stacktrace
+    :trace})
 
 (defn read-debug-command
   "Read and take action on a debugging command.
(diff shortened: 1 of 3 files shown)


src/cider/nrepl/middleware/debug.clj
```diff
@@ -297,19 +297,21 @@ this map (identified by a key), and will `dissoc` it afterwards."}
...
-   "l" :locals
-   "n" :next
-   "o" :out
-   "p" :inspect
-   "q" :quit
-   "s" :stacktrace
-   "t" :trace})
+  "An unsorted set of commands supported by the debugger."
+  #{:continue
+    :Continue
```
bbatsov: We should use a proper name here - the current one was just to convey the relationship with `C`.

src/cider/nrepl/middleware/debug.clj
```diff
@@ -297,19 +297,21 @@ this map (identified by a key), and will `dissoc` it afterwards."}
...
+    :eval
+    :here
+    :in
+    :inject
+    :locals
+    :next
+    :out
+    :inspect
+    :insPect
```
bbatsov: Same here.

src/cider/nrepl/middleware/debug.clj
yuhan0: `:continue-all` and `:inspect-prompt`? Or maybe `:inspect-current` for `p` 
Would that also call for a command name -> display name mapping on the client side? Right now it's just using the plain command names as overlay titles.

bbatsov: You read my mind. I was just writing about some display name mapping on the other ticket. :-)

> :continue-all

👍 

For inspect - let's have the default command inspect the current value and the other to be `inspect-prompt`.

bbatsov:
Btw, I guess the second inspect should go in a different PR, as we don't really have that command available currently.

yuhan0:
Ok, I rebased the conflicts, removed the references to :insPect and renamed :Continue to :continue-all. 

Somehow I can't seem to get the test suite working locally - `make test` reports an unrelated error in the info-test and passes the cider.nrepl.debug tests.

```
FAIL in (integration-test) (info_test.clj:161)
info op get info of a java method
expected: (.startsWith (:javadoc response) "https://docs.oracle.com/")
  actual: false
```

yuhan0:
Ok it was a fairly simple change, now all existing tests are passing :)

bbatsov:
@yuhan0 Just squash the last two commits and we're good to go.

bbatsov:
Great work! Thanks for tackling this! 🙇

PR #658, opened 2019-11-29
Move maint out of src

dpsutton:
Two issues:
1. Our CI is failing our linting process with

```
Directory: /home/circleci/repo/src
    File                   : maint/cider/nrepl/impl/docs.clj
    has namespace          : cider.nrepl.impl.docs
    should have namespace  : maint.cider.nrepl.impl.docs
    or should be in file(s): cider/nrepl/impl/docs.cljc
                             cider/nrepl/impl/docs.clj

```
this arises from maint being under src. Src is on the classpath so
maint is included as well. Meaning that the file/ns didn't work out
correctly.

2. This maintenance script has been included in jars for a
while. Since its under src/ it gets put in the jar and there's no need
for this.

There were two ways to solve this: move maint out of src or nest all
the current files under some common folder so that the lib would be
src/lib and the docs would be src/maint. Seemed a bit heavy so this
was chosen.


bbatsov:
Sounds good to me. I just copied the approach we had taken in nREPL. Probably we should update this there as well.

dpsutton:
Alternative PR at https://github.com/clojure-emacs/cider-nrepl/pull/659 which mimics the change here. Noteworthy is that it moves 41 files rather than 2. Up to you which you prefer: consistency but with a larger commit or inconsistency but a much smaller footprint.

bbatsov:
I'd probably go with the smaller change. In nREPL's case we actually had the subdirs under `src` since the beginning, because of some legacy Java code. I guess I just didn't pay attention to this while porting the docs generator to cider-nrepl.

PR #659, opened 2019-12-01
Move src/maint out of normal app classpath

dpsutton:
mimics the approach taken in nREPL


dpsutton:
closing in favor of https://github.com/clojure-emacs/cider-nrepl/pull/658

PR #661, opened 2019-12-28
Fix info op not to return nil as a document for protocl

liquidz:
- [x] All tests are passing

Currently, the meta data for protocol method without docstring has a `nil` value for `:doc` key.
```clj
(defn foo [])
(defn bar "..." [])
(defprotocol Dummy
  (hello [this])
  (world [this] "..."))

(meta #'foo)   ; not containing `:doc`
(meta #'bar)   ; containing `:doc`
(meta #'hello) ; containing `:doc`, but nil
(meta #'world) ; containing `:doc`
```

The `nil` value is treated as a list by nrepl/bencode, so `info` op returns a empty list as a docstring for protocol method.
https://github.com/nrepl/bencode/blob/v1.0.0/src/bencode/core.clj#L324

E.g. `info` response for `hello`
```sh
d12:arglists-str6:[this]3:docle4:file47:file:/.../core.clj4:linei7e4:name5:hello2:ns8:foo.core8:protocol16:#'foo.core/Dummy7:session36:3b4401ac-a962-43b1-8fbc-1c17d
b0477346:statusl4:doneee

# 3:docle <= Here!
```

Is this an expected behavior?

This PR will dissociate `:doc` key when the value is nil not to return empty list as a bencode response.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@
   Change the format of debugger command messages to a set of command names, leaving key mappings up to client implementations.
   * [#653](https://github.com/clojure-emacs/cider-nrepl/pull/653): Add `inspect-def-current-value` to inspect middleware.
 
+### Changes
+
+* [#661](https://github.com/clojure-emacs/cider-nrepl/pull/661): Fix `info` op not to return nil as a value of key.
+
 ## 0.22.4 (2019-10-08)
 
 ### Changes
--- a/src/cider/nrepl/middleware/info.clj
+++ b/src/cider/nrepl/middleware/info.clj
@@ -11,6 +11,16 @@
 
 (declare format-response)
 
+(defn dissoc-nil-keys
+  "Dissociate keys which has nil as a value to avoid returning empty list as a nil.
+  nrepl/bencode converts nil to empty list."
+  [info]
+  (reduce-kv
+   (fn [res k v]
+     (cond-> res
+       (some? v) (assoc k v)))
+   {} info))
+
 (defn format-nested
   "Apply response formatting to nested `:candidates` info for Java members."
   [info]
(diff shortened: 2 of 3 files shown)


bbatsov:
> Is this an expected behavior?

It is, but it's definitely not ideal as clients can't tell apart an empty list from a `nil`. Generally it's best to avoid returning empty keys as you suggested.

> This PR will dissociate :doc key when the value is nil not to return empty list as a bencode response.

Probably we can have a more generic solution and drop any metadata that's nil from the response. Then the clients would know that if something's missing it's nil.

liquidz:
@bbatsov Thanks! I fixed a commit.

CI for `Java 11, Clojure 1.10` seems failure, but I didn't change anything about `cljs-inspect-test` and the CI is successful in my local environment.
```
$ java -version
openjdk version "11.0.2" 2019-01-15
OpenJDK Runtime Environment 18.9 (build 11.0.2+9)
OpenJDK 64-Bit Server VM 18.9 (build 11.0.2+9, mixed mode)

$ lein with-profile +1.10,+test test cider.nrepl.middleware.cljs-inspect-test

lein test cider.nrepl.middleware.cljs-inspect-test
Warning: Nashorn engine is planned to be removed from a future JDK release

Ran 19 tests containing 50 assertions.
0 failures, 0 errors.
```

bbatsov:
Looks good to me. Just mention this fix in the changelog and we're good to go.

liquidz:
@bbatsov I just fixed changelog!
https://github.com/clojure-emacs/cider-nrepl/compare/525c0871900b18e3703f9f6e44d4d9f50bf0ff26..e4b0a131d1b3ea29714814701783876fcdc118fa

bbatsov:
Thanks!

PR #663, opened 2020-02-03
Fix non-recognized `recur` inside `case` in the debugger.  

FiV0:
Case expands the non-default branches into a map. This can be seen
when evaluating the following code:
```clj 
(walk/macroexpand-all '(case i
                         10 (recur (inc i))
                         :halt))
```
The `contains-recur?` function did so far not take that into account.

- [x] All tests are passing (kind of)
I was able to run tests for 1.9 and 1.10. 1.9 had 1 failure non related to my code and 1.10 none.
I was unable to run the tests for 1.8 because of some leiningen issue.

This should be fixing [cider#2772](https://github.com/clojure-emacs/cider/issues/2772)

### DIFF
--- a/src/cider/nrepl/middleware/util/instrument.clj
+++ b/src/cider/nrepl/middleware/util/instrument.clj
@@ -164,16 +164,19 @@
 (defn- contains-recur?
   "Return true if form is not a `loop` or a `fn` and a `recur` is found in it."
   [form]
-  (cond (seq? form) (case (first form)
-                      recur true
-                      loop* false
-                      fn*   false
-                      (some contains-recur? (rest form)))
-        ;; case* expands the non-default branches into a map
-        ;; this depends on internal logic
-        (map? form) (some contains-recur? (vals form))
-        (vector? form) (some contains-recur? (seq form))
-        :else false))
+  (cond
+    (seq? form) (case (first form)
+                  recur true
+                  loop* false
+                  fn*   false
+                  (some contains-recur? (rest form)))
+    ;; `case` expands the non-default branches into a map.
+    ;; We therefore expect a `recur` to appear in a map only
+    ;; as the result of a `case` expansion.
+    ;; This depends on Clojure implementation details.
+    (map? form) (some contains-recur? (vals form))
+    (vector? form) (some contains-recur? (seq form))
+    :else false))
 
 (defn- dont-break?
   "Return true if it's NOT ok to wrap form in a breakpoint.
--- a/src/cider/nrepl/middleware/util/instrument.clj
+++ b/src/cider/nrepl/middleware/util/instrument.clj
@@ -164,12 +164,15 @@
 (defn- contains-recur?
   "Return true if form is not a `loop` or a `fn` and a `recur` is found in it."
   [form]
-  (if (seq? form)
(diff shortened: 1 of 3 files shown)


src/cider/nrepl/middleware/util/instrument.clj
```diff
@@ -164,12 +164,16 @@
...
+  (cond (seq? form) (case (first form)
+                      recur true
+                      loop* false
+                      fn*   false
+                      (some contains-recur? (rest form)))
+        ;; case* expands the non-default branches into a map
+        ;; this depends on internal logic
+        (map? form) (some contains-recur? (vals form))
+        (vector? form) (some contains-recur? (seq form))
+        :else false))
```
FiV0: I don't know if this is the best way to handle the issue. I am open for other suggestions. I mean normally a `recur` should only appear in a map when used inside a `case`.

src/cider/nrepl/middleware/util/instrument.clj
```diff
@@ -164,12 +164,16 @@
 (defn- contains-recur?
   "Return true if form is not a `loop` or a `fn` and a `recur` is found in it."
   [form]
-  (if (seq? form)
-    (case (first form)
-      recur true
-      loop* false
-      fn*   false
-      (some contains-recur? (rest form)))))
+  (cond (seq? form) (case (first form)
```
bbatsov: I'd put a linebreak after `cond`, so that the branches are easier to spot.

src/cider/nrepl/middleware/util/instrument.clj
```diff
@@ -164,12 +164,16 @@
...
+                      recur true
+                      loop* false
+                      fn*   false
+                      (some contains-recur? (rest form)))
+        ;; case* expands the non-default branches into a map
```
bbatsov: I think this comment should probably be expanded a bit for posterity's sake. Perhaps you can also add here your remark that we expect `recur` to appear in a map/vector only as the result of a `case` expansion.

src/cider/nrepl/middleware/util/instrument.clj
```diff
@@ -164,12 +164,16 @@
...
+        ;; this depends on internal logic
```
bbatsov: internal logic -> Clojure implementation details

src/cider/nrepl/middleware/util/instrument.clj
bbatsov: I think your approach is pretty reasonable, but perhaps @Malabarba or @vspinu  have some better ideas.

bbatsov:
The code looks good to me overall, we just need a changelog entry. 

P.S. I really appreciate that you took the time to get familiar with the debugger and tackle some issues there! 🙇

FiV0:
@bbatsov I don't know if you saw it, but I also left a longer comment in https://github.com/clojure-emacs/cider/issues/2772 outlining some more general problem in the debugger. No pressure to reply soon. Just wanted to make sure you had seen it. I can also open an issue here for discussion as it is more related to cider-nrepl than to the cider emacs side.

bbatsov:
Just replied there as well.

PR #664, opened 2020-02-04
Fix continue-all in conditional break.

### LINKED ISSUE #656: :continue-all does not skip form with :break/when metadata
## Expected behavior
Sending a :continue-all command ("C") in the debugger should skip all instrumented forms, regardless of whether they have the `^{:break/when ...}` metadata applied.

## Actual behavior
The debugger behaves like regular :continue ("c") and only skips the current breakpoint

```
(for [n (range 10)]
  #dbg ^{:break/when (even? n)}
  (str (inc n)))
```
Evaluate in Cider and press "C", observe that it breaks on every even number instead of skipping to the end of the evaluation.

### cider-nrepl version

0.22.4

### Java version
;; Clojure 1.10.1, Java 12.0.1

### Operating system
macOS 10.14

FiV0:
The :continue-all command was previously not
ignoring conditional breakpoints as in the
following code.

(for [i (range 10)]
  #dbg ^{:break/when (even? i)}
  (inc i))

Fixes #656.

- [x] All tests are passing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## master (unreleased)
 
+### New Features
+
+### Changes
+
+* [#664](https://github.com/clojure-emacs/cider-nrepl/pull/664): Fix continue-all in conditional break.
+
 ## 0.23.0 (2019-01-18)
 
 ### New Features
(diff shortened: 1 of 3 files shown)


bbatsov:
Thanks for tackling this! The code looks good. Just mention the bugfix in the changelog.

PR #665, opened 2020-02-05
Fix instrumentation of small map literals. Minor doc updates.

FiV0:
Map literal instrumentation was relying on some heuristic before.
We use Clojures underlying array-map implementation for small map
literals and fall back to a sorted order of keys if that is not the
case. If this fails as well we simply instrument only the values of
the map in the order they appear in a sequential pass.

This fixes https://github.com/clojure-emacs/cider/issues/2773 for small maps at least.

- [x] All tests are passing

### DIFF
--- a/src/cider/nrepl/middleware/util/instrument.clj
+++ b/src/cider/nrepl/middleware/util/instrument.clj
@@ -26,7 +26,7 @@
 
 ;;;; ## Instrumentation
 ;;;
-;;; The top-level instrumenting function is `read-and-instrument`. See
+;;; The top-level instrumenting function is `instrument-tagged-code`. See
 ;;; its doc for more information.
 ;;;
 ;;; Each of the other `instrument-*` functions is responsible for
@@ -208,7 +208,7 @@
   Only forms with a ::breakfunction metadata will be
   instrumented, and even then only if it makes sense (e.g., the
   binding names in a let form are never instrumented).
-  See `read-and-instrument` for more information."
+  See `instrument-tagged-code` for more information."
   [form]
   (condp #(%1 %2) form
     ;; Function call, macro call, or special form.
@@ -225,7 +225,7 @@
 ;;;
 ;;; The following functions are used to populate with metadata and
 ;;; macroexpand code before instrumenting it. This is where we
-;;; calculate all the ::coor vectors. See `read-and-instrument`.
+;;; calculate all the ::coor vectors. See `instrument-tagged-code`.
 (defn- walk-indexed
   "Walk through form calling (f coor element).
   The value of coor is a vector of indices representing element's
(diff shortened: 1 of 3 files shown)


FiV0:
I haven't yet added any tests as I want to get your feedback first.

src/cider/nrepl/middleware/util/instrument.clj
```diff
@@ -236,20 +236,38 @@
...
-                      :else form)]
+         ;; Clojure uses array-maps up to some threshold (8 currently).
+         ;; So for small maps we take advantage of that, otherwise fall
+         ;; back to the heuristic below.
+         ;; Maps are unordered, but we can try to use the keys as order
+         ;; hoping they can be compared one by one and that the user
+         ;; has specified them in that order. If that fails we fall
+         ;; back to simply instrumenting the values in the order they
+         ;; they appear in the map.
+         ;; This depends on Clojure implementation details.
```
FiV0: More than happy to consider other heuristics or if you think this is a bad idea.

src/cider/nrepl/middleware/util/instrument.clj
```diff
@@ -236,20 +236,38 @@
...
+                                  (into {} (map-indexed (fn [i [k v]]
+                                                          [(walk-indexed (conj coor (* 2 i)) f k)
+                                                           (walk-indexed (conj coor (inc (* 2 i))) f v)])
+                                                        (into (sorted-map) form)))
+                                  (catch Exception e
+                                    (into {} (map-indexed (fn [i [k v]]
+                                                            [k (walk-indexed (conj coor (inc (* 2 i))) f v)])
+                                                          form)))))
+                  ;; Order of sets is unpredictable, unfortunately.
+                  (set? form)  form
```
FiV0: Also happy to take suggestions for heuristics for set literals.

src/cider/nrepl/middleware/util/instrument.clj
```diff
@@ -236,20 +236,38 @@
...
-                                                 form))
-                      ;; Order of sets is unpredictable, unfortunately.
-                      (set? form)  form
-                      ;; Borrowed from clojure.walk/walk
-                      (list? form) (apply list (map-inner form))
-                      (instance? clojure.lang.IMapEntry form) (vec (map-inner form))
-                      (seq? form)  (doall (map-inner form))
-                      (coll? form) (into (empty form) (map-inner form))
-                      :else form)]
+         ;; Clojure uses array-maps up to some threshold (8 currently).
```
bbatsov: to some map size

src/cider/nrepl/middleware/util/instrument.clj
bbatsov: The idea is reasonable, although this is something we should probably mention in some "Caveats" section of the debugger, so users are not surprised by the behaviour in different scenarios.

src/cider/nrepl/middleware/util/instrument.clj
bbatsov: I'm afraid nothing comes to mind.

src/cider/nrepl/middleware/util/instrument.clj
FiV0: I changed the policy once more. I think the best thing is to do nothing if the ordered keys approach fails. I will also start a caveats section in the cider docs.

bbatsov: I like the updated approach.

bbatsov:
Thanks!

Issue #530, opened 2018-05-09
Apropos creates useless regex when not passed a `search`


SevereOverfl0w:
https://github.com/clojure-emacs/cider-nrepl/blob/d18a6e89fc0057a9c78531b5ddf3c35229020993/src/cider/nrepl/middleware/apropos.clj#L21-L22

If there is no `search` this will generate `(?i:null)` which doesn't match much. It should do a `when` around the `search` first.

bbatsov:
But isn't `search` mandatory? What exactly is apropos going to do without it?

SevereOverfl0w:
It's fully capable of working without it now, the var-query will return everything if no `search` is provided (excluding other filters such as metadata and test). For example, you should be able to use apropos to filter for all tests by using:

```
{:op "apropos" :var-query {"test?" 1}}
```

Not providing search is the same as having search set to `#".*"`, but without the effort of actually running regex.

MalloZup:
autogenerated with https://github.com/MalloZup/doghub:  issue inactive since 450 days. Please update the issue or close it

bbatsov:
Fixed by #666. The PR of the Beast! 😆

PR #666, opened 2020-02-17
Add a check in apropos var-query-map

Cartmanishere:
This is a change as described in https://github.com/clojure-emacs/cider-nrepl/issues/530.

In `apropos` middleware, if no `query` is passed in the message then a redundant regex is formed.
This is because there is no condition that checks whether the value exists before constructing the regex.

I have added such a check. I have also added a test for this specific case.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### Bugs fixed
+
+* [#666](https://github.com/clojure-emacs/cider-nrepl/pull/666): Add a check in apropos var-query-map.
+
 ### New Features
 
 ## 0.24.0 (2020-02-14)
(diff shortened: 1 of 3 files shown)


bbatsov:
The change looks good to me. We should also mention it in the changelog.

Cartmanishere:
Added the pull request in the changelog

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
 * [#663](https://github.com/clojure-emacs/cider-nrepl/pull/663): Fix non-recognized `recur` inside `case` in the debugger.
 * [#664](https://github.com/clojure-emacs/cider-nrepl/pull/664): Fix continue-all in conditional break.
 * [#665](https://github.com/clojure-emacs/cider-nrepl/pull/665): Fix form location in map literals.
+* [#666](https://github.com/clojure-emacs/cider-nrepl/pull/666): Add a check in apropos var-query-map.
```
bbatsov: This should be under `master`. Add a new "Bugs fixed" section there, as 0.24 was release a few days ago.

bbatsov:
Thanks!

PR #667, opened 2020-02-24
Filter non file urls from classpath response

fp7:
Not all urls on a classpath are necessarily files. Eg. spring boot
adds "jar:file:..." urls

This can only be merged after https://github.com/clojure-emacs/orchard/commit/9fe7a70e7ea0e7984317aa80634367ff5ca327e9 was released.

This is at least related to https://github.com/clojure-emacs/orchard/pull/83

### DIFF
--- a/src/cider/nrepl/middleware/classpath.clj
+++ b/src/cider/nrepl/middleware/classpath.clj
@@ -2,10 +2,17 @@
   (:require
    [cider.nrepl.middleware.util.error-handling :refer [with-safe-transport]]
    [clojure.java.io :as io]
-   [orchard.java.classpath :as cp]))
+   [orchard.java.classpath :as cp]
+   [orchard.misc :as misc]))
+
+(defn file-url?
+  [u]
+  (and (misc/url? u)
+       (= (.getProtocol ^java.net.URL u) "file")))
 
 (defn classpath-reply [msg]
   {:classpath (->> (cp/classpath)
+                   (filter file-url?)
                    (map io/as-file)
                    (map str))})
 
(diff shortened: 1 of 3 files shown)


bbatsov:
> This can only be merged after clojure-emacs/orchard@9fe7a70 was released.

👍 I'll probably cut a new release in a week or so.

fp7:
Here we go

src/cider/nrepl/middleware/classpath.clj
```diff
@@ -2,10 +2,17 @@
   (:require
    [cider.nrepl.middleware.util.error-handling :refer [with-safe-transport]]
    [clojure.java.io :as io]
-   [orchard.java.classpath :as cp]))
+   [orchard.java.classpath :as cp]
+   [orchard.misc :as misc]))
+
+(defn file-url?
```
bbatsov: Btw, doesn't it make more sense for this to be in `orchard` as well? Or at least a private function here?

src/cider/nrepl/middleware/classpath.clj
```diff
@@ -2,10 +2,17 @@
...
+  [u]
+  (and (misc/url? u)
+       (= (.getProtocol ^java.net.URL u) "file")))
 
 (defn classpath-reply [msg]
   {:classpath (->> (cp/classpath)
```
bbatsov: I've also been thinking we can just add a `classpath-files` function to orchard.

fp7: I wonder if that would filter out too much for other clients of orchard. Maybe some clients want to fix this problem instead of just removing the classpath entries which causes this function to fail. Would it not take that possibility from them if we filter on the orchard side?

fp7:
Sorry for the late reply. Was busy the last two weeks

src/cider/nrepl/middleware/classpath.clj
bbatsov: Well, adding the function there wouldn't mean that people are forced to use it. It simply seems to me like something that's generally useful outside `cider-nrepl`.

Issue #534, opened 2018-06-15
clojure.java.io/file throws `Unhandled REPL handler exception processing message`


taylskid:
## Expected behavior

```
user> (clojure.java.io/file "/Users/taylorskidmore/foo.json")
user> (slurp (clojure.java.io/file "/Users/taylorskidmore/foo.json"))
""
```

## Actual behavior

```
ERROR: Unhandled REPL handler exception processing message {:op slurp, :url file:/Users/taylorskidmore/foo.json, :session 6dfeb733-5821-43e2-9f7e-fa952127297d, :id 11}
java.lang.NullPointerException
	at java.util.regex.Matcher.getTextLength(Matcher.java:1283)
	at java.util.regex.Matcher.reset(Matcher.java:309)
	at java.util.regex.Matcher.<init>(Matcher.java:229)
	at java.util.regex.Pattern.matcher(Pattern.java:1093)
	at clojure.core$re_matcher.invokeStatic(core.clj:4674)
	at clojure.core$re_find.invokeStatic(core.clj:4716)
	at clojure.core$re_find.invoke(core.clj:4716)
	at cider.nrepl.middleware.slurp$normalize_content_type.invokeStatic(slurp.clj:55)
	at cider.nrepl.middleware.slurp$normalize_content_type.invoke(slurp.clj:50)
	at cider.nrepl.middleware.slurp$slurp_url_to_content_PLUS_body.invokeStatic(slurp.clj:92)
	at cider.nrepl.middleware.slurp$slurp_url_to_content_PLUS_body.invoke(slurp.clj:85)
	at cider.nrepl.middleware.slurp$handle_slurp.invokeStatic(slurp.clj:124)
	at cider.nrepl.middleware.slurp$handle_slurp.invoke(slurp.clj:114)
	at clojure.lang.Var.invoke(Var.java:383)
	at cider.nrepl$wrap_slurp$fn__1441.invoke(nrepl.clj:139)
	at clojure.tools.nrepl.middleware$wrap_conj_descriptor$fn__424.invoke(middleware.clj:22)
	at cider.nrepl$wrap_stacktrace$fn__1573.invoke(nrepl.clj:425)
	at clojure.tools.nrepl.middleware$wrap_conj_descriptor$fn__424.invoke(middleware.clj:22)
	at cider.nrepl$wrap_classpath$fn__1457.invoke(nrepl.clj:155)
[...]

PR #669, opened 2020-02-29
Fix NPE and add isDirectory check in slurp

Cartmanishere:
Fix some issues in `slurp` middleware as seen in https://github.com/clojure-emacs/cider-nrepl/issues/534

### DIFF
--- a/src/cider/nrepl/middleware/slurp.clj
+++ b/src/cider/nrepl/middleware/slurp.clj
@@ -87,13 +87,14 @@
     (if (= (.getProtocol url) "file") ;; expected common case
       (let [^Path p (Paths/get (.toURI url))
             content-type (normalize-content-type (get-file-content-type p))
-            buff (Files/readAllBytes p)]
+            buff (when-not (.isDirectory (io/as-file url)) (Files/readAllBytes p))]
         (slurp-reply p content-type buff))
 
       ;; It's not a file, so just try to open it on up
       (let [^URLConnection conn (.openConnection url)
             content-type (normalize-content-type
-                          (.getContentType conn))
+                          (or (.getContentType conn)
+                              "application/octet-stream"))
             ;; FIXME (arrdem 2018-04-03):
             ;;   There's gotta be a better way here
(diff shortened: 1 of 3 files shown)


test/clj/cider/nrepl/middleware/slurp_test.clj
```diff
@@ -26,3 +26,10 @@
     (t/is (= ["application/octet-stream" {}] (:content-type resp)))
     (t/is (str/starts-with? (:body resp) "#binary[location="))
     (t/is (str/ends-with? (:body resp) ",size=681]"))))
+
+(t/deftest test-directory
```
bbatsov: Don't we also need for whatever was causing the NPE?

src/cider/nrepl/middleware/slurp.clj
```diff
@@ -43,7 +43,8 @@
   `[type {:as attrs}]`. This method normalizes RFC
   compliant content-types to this form."
   [^String content-type]
-  (if-let [match (re-find content-type-pattern content-type)]
+  (if-let [match (and content-type
```
bbatsov: Shouldn't this `nil` check be handled in the caller of `normalize-content-type`? I think it's weird to expect to receive a `nil` argument.

bbatsov:
Thanks for tackling this! I've left some small remarks. Don't forget to add a changelog entry as well.

src/cider/nrepl/middleware/slurp.clj
Cartmanishere: Yeah. Makes sense.

test/clj/cider/nrepl/middleware/slurp_test.clj
Cartmanishere: The problem was some URLs return nil when doing `.getContentType`. I couldn't really figure out why the URLs did not return a content type. 
I thought it was because of redirection but not all redirected URLs return nil. Tried various HTTP return codes. No idea why this happens.

bbatsov:
Thanks for tackling this! 🙇

PR #670, opened 2020-03-03
Extend undef op to work on fully qualified symbols

yuhan0:

- [ ] All tests are passing

See https://github.com/clojure-emacs/cider/issues/2814.
The current cider.el frontend sends `undef` op with `ns` as the current ns and `symbol` as a possibly ns-qualified var, which this PR interprets as unmapping the symbol from the corresponding other namespace.

For some reason running tests locally with `make test` turns up many failures in `cider.nrepl.middleware.debug-integration-test` which seem to be unrelated to the changes in this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### New Features
 
+* [#670](https://github.com/clojure-emacs/cider-nrepl/pull/670): Extend undef op to work on fully qualified symbols.
+
 ## 0.24.0 (2020-02-14)
 
 ### Bugs fixed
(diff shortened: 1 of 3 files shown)


bbatsov:
> The current cider.el frontend sends undef op with ns as the current ns and symbol as a possibly ns-qualified var, which this PR interprets as unmapping the symbol from the corresponding other namespace.

This looks reasonable to me. 
I'd suggest documenting the update behaviour in the ns docstring (high-level) and the `undef` function's docstring. 

This should also be mentioned in the changelog.

bbatsov:
Thanks for tackling this! 🙇

Issue #671, opened 2020-03-07
Debugger doesn't instrument inside dot special forms


Cartmanishere:
## Expected behavior
Debugging should work with all possible versions of dot special form.

## Actual behavior
When using the `(. Classname-symbol (method-symbol args*))` dot special form as given in the [Java Interop](https://clojure.org/reference/java_interop) documentation, the debugger does not stop at the `args*` that are provided.

## Steps to reproduce the problem

```
(defn test
  [a]
  (. clojure.lang.Numbers (add 1 (let [x 1] (+ 2 x)))))
```
If you instrument and debug the above function, the debugger does not step inside the `let` expression. Ideally there should be a breakpoint at the `x` and `(+ 2 x)`. 

But all the code inside the `*args` is not instrumented.

In contrast, the dot special form of this type `(. Classname-symbol method-symbol args*)` works properly.

```
(defn test
  [a]
  (. clojure.lang.Numbers add 1 (let [x 1] (+ 2 x))))
```
When debugging the above function, we hit the breakpoints at `x` and `(+ 2 x)`.

I have isolated the problem in the `instrument.clj` where the `.` special form is handled in the code for instrumentation.

I am working on a patch for this.

### cider-nrepl version

`0.25.0-SNAPSHOT`

### Java version

Java 1.8

### Operating system

MacOS

PR #672, opened 2020-03-09
Fix instrument for a dot special form case

Cartmanishere:

- [x] All tests are passing
--- 

Fix for https://github.com/clojure-emacs/cider-nrepl/issues/671

Instrumentation for dot special forms of type `(. clojure.lang.Numers (add 1 x))` fails. There is no breakpoint set for `x`.

This PR fixes this so that the breakpoints are set inside the dot special form.

### DIFF
--- a/src/cider/nrepl/middleware/util/instrument.clj
+++ b/src/cider/nrepl/middleware/util/instrument.clj
@@ -73,7 +73,15 @@
             '#{new} (cons (first args) (instrument-coll (rest args)))
             '#{quote & var clojure.core/import*} args
             '#{.} (list* (first args)
-                         (second args)
+                         ;; To handle the case when second argument to dot call
+                         ;; is a list e.g (. class-name (method-name args*))
+                         ;; The values present as args* should be instrumented.
+                         (let [s (second args)]
+                           (if (coll? s)
+                             (->> (rest s)
+                                  instrument-coll
+                                  (concat (cons (first s) '())))
+                             s))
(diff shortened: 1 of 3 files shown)


src/cider/nrepl/middleware/util/instrument.clj
```diff
@@ -73,7 +73,12 @@
             '#{new} (cons (first args) (instrument-coll (rest args)))
             '#{quote & var clojure.core/import*} args
             '#{.} (list* (first args)
-                         (second args)
+                         (let [s (second args)]
```
bbatsov: Might be a good idea to add some comments explaining this block of code.

test/clj/cider/nrepl/middleware/util/instrument_test.clj
```diff
@@ -138,6 +138,14 @@
          '#{[(def foo "foo doc" (bp (bar) {:coor [3]} (bar))) []]
             [(bar) [3]]})))
 
+(deftest instrument-dot-test
+  (is (= (t/breakpoint-tester '(. clojure.lang.Numbers add 1 x))
```
bbatsov: Might be a good idea to wrap the tests in `testing` blocks with some explanations of why those cases are different.

bbatsov:
Thanks for tackling this! I've left a couple of tiny remarks, but overall your changes look great!

Cartmanishere:
Added some explanation in code and tests.

bbatsov:
Thanks!

PR #674, opened 2020-04-05
fix double-reference to clj-suitable in the README

blak3mill3r:
Took me a moment to realize these are the same project, clj-suitable


bbatsov:
Thanks!

PR #683, opened 2020-11-21
Support custom print-methods in test reports

FelipeCortez:
This PR fixes test reporting for libraries that rely on custom print logic that can be added by `(defmethod print-method ,,,`, such as `matcher-combinators`: https://github.com/clojure-emacs/cider/issues/2901#issuecomment-723312712

### DIFF
--- a/src/cider/nrepl/middleware/test.clj
+++ b/src/cider/nrepl/middleware/test.clj
@@ -63,30 +63,34 @@
        (filter #(= (:class %) (.getName (class f))))
        (first)))
 
-(defn- custom-print-method? [o]
-  (= (get-method print-method (:type (meta o)))
-     (get-method print-method :default)))
+(defn- print-object [object]
+  (let [print-fn (if (= (get-method print-method (:type (meta object)))
+                        (get-method print-method :default))
+                   pp/pprint
+                   println)]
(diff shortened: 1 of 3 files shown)


src/cider/nrepl/middleware/test.clj
```diff
@@ -72,7 +76,7 @@
         c (when (seq test/*testing-contexts*) (test/testing-contexts-str))
         i (count (get-in (@current-report :results) [ns (:name (meta v))]))
         gen-input (:gen-input @current-report)
-        pprint-str #(with-out-str (pp/pprint %))]
+        pprint-str #(with-out-str (if (custom-print-method? %) (pp/pprint %) (println %)))]
```
bbatsov, line 79: Might also be a good idea to add some comment explaining the need for this, and ideally a unit test.

bbatsov:
Don't forget to add a changelog entry for this bug-fix.

src/cider/nrepl/middleware/test.clj
FelipeCortez: @bbatsov done! can you check again?

bbatsov:
Looks good. Thanks!

Issue #447, opened 2017-10-17
NPE on connect with embedded nREPL server


DerGuteMoritz:
## Expected behavior

Embedding an nREPL server as described in https://github.com/clojure-emacs/cider-nrepl#via-embedding-nrepl-in-your-app works.

## Actual behavior

With version 0.15.x this still worked but with 0.16.0-SNAPSHOT (specifically version `bd7b2bc4c780549594bd7f03c772465216c1e9ca`), the server runs into a `NullPointerException` when a client connects, like this:

```
18:26:38.661 ERROR [clojure.tools.nrepl.server] (clojure-agent-send-off-pool-3) Unhandled REPL handler exception processing message {:op describe, :session 2a63e04c-29af-4e9b-acbb-5247dcc140f2, :id 61}
java.lang.NullPointerException: null
	at clojure.core$deref_future.invokeStatic(core.clj:2297) ~[foo-0.1.0-SNAPSHOT-standalone.jar:na]
	at clojure.core$deref.invokeStatic(core.clj:2317) ~[foo-0.1.0-SNAPSHOT-standalone.jar:na]
	at clojure.core$deref.invoke(core.clj:2303) ~[foo-0.1.0-SNAPSHOT-standalone.jar:na]
	at cider.nrepl$wrap_pprint_fn$fn__24216.invoke(nrepl.clj:85) ~[foo-0.1.0-SNAPSHOT-standalone.jar:na]
	at clojure.tools.nrepl.middleware$wrap_conj_descriptor$fn__23554.invoke(middleware.clj:22) ~[foo-0.1.0-SNAPSHOT-standalone.jar:na]
	at cider.nrepl$wrap_tracker$fn__24336.invoke(nrepl.clj:410) ~[foo-0.1.0-SNAPSHOT-standalone.jar:na]
	at clojure.tools.nrepl.middleware$wrap_conj_descriptor$fn__23554.invoke(middleware.clj:22) ~[foo-0.1.0-SNAPSHOT-standalone.jar:na]
	at clojure.tools.nrepl.middleware.session$session$fn__23873.invoke(session.clj:192) ~[foo-0.1.0-SNAPSHOT-standalone.jar:na]
	at clojure.tools.nrepl.middleware$wrap_conj_descriptor$fn__23554.invoke(middleware.clj:22) ~[foo-0.1.0-SNAPSHOT-standalone.jar:na]
[...]

Issue #657, opened 2019-11-05
cider.nrepl/delayed-handlers empty with AOT


deas:
Cider-Nrepl is part of [our app](https://github.com/pink-gorilla/gorilla-notebook) which we ship AOT compiled. At this time, we use version 0.22.4. 

PS: Current workaround is to use:

```
:uberjar-exclusions [#"cider/nrepl.*\.class$"]
```
In `project.clj`.

## Expected behavior
Fully AOT compiled cider working.

## Actual behavior
`cider.nrepl/delayed-handlers` is empty. Calls to cider middleware throw NPE

```
java.lang.NullPointerException
        at clojure.core$deref_future.invokeStatic(core.clj:2300)
        at clojure.core$deref.invokeStatic(core.clj:2320)
        at clojure.core$deref.invoke(core.clj:2306)
        at cider.nrepl$wrap_stacktrace$fn__7449.invoke(nrepl.clj:391)
        at nrepl.middleware$wrap_conj_descriptor$fn__5254.invoke(middleware.clj:16)
```

## Steps to reproduce the problem
Make sure to use Java 8.

`lein uberjar` at commit `2ca797e94a19ab0feaaa7780e9fa9b311389e3fe` of gorilla-notebook.

Run the application:

```
java -jar target/gorilla-notebook-standalone.jar
```

and go [here](http://localhost:9000/worksheet.html).

Raise exception in a code cell:

```
(.toString nil)
```

### cider-nrepl version

0.22.4

### Java version

1.8

### Operating system
Any

Issue #533, opened 2018-05-29
Side Effects in top level code => breaks use in compiled scenarios


jsa-aerial:
See how and when delayed-handlers is updated, and also that cider-nrepl-handler is built as a side-effecting top level form.  This is the cause of (for example) #447. The key thing to note in the hack to work around this as presented in that issue is that cider.nrepl _cannot_ be required at _compile_ time (as in a ns form). So, until this is fixed, that should be explicitly mentioned (in bold!) in the documented workaround.

## Expected behavior
Embedding an nREPL server in compiled situations (for example, uberjar application) by explicitly using / constructing middleware  using cider.nrepl/cider-nrepl-handle supports connections with no problem

## Actual behavior
Embedding an nREPL server in compiled situations (for example, uberjar application) by explicitly using / constructing middleware  using cider.nrepl/cider-nrepl-handler results in NPE on any connection

## Steps to reproduce the problem
1. Embed an nREPL server by requiring cider.nrepl in an ns form
2. Connect to the server using the Emacs CIDER client

### cider-nrepl version
Any version of cider.nrepl since Sep 2017. The one I used was 0.17.0  The problem is how the deferred middleware was implemened

### Java version

1.8

### Operating system

Linux

bbatsov:
Yeah, I guess we overlooked this initially. Hopefully @vspinu can take a look at this once he's done with the connection rework.

vspinu:
Unless I miss something this is an exact duplicate of #447 and #464.

jsa-aerial:
Sort of - Bozhidar explicitly asked me to make this issue (on Slack) as it explicitly notes that the problem is due to side effecting top level forms and where and how those effects are happening. Neither of the other two (I actually reference #447 in the description...) note the base problem or where it occurs.

If you want to get rid of this issue - go for it. Doesn't matter to me as long as it gets corrected

vspinu:
>  explicitly notes that the problem is due to side effecting top level forms and where and how those effects are happening.

I am afraid this is not that helpful (to me at least). By "side-effecting" you presumably mean updating the delays atom. Ok, and? What problems does that cause and through which mechanism more concretely? Do you have a proposal for an alternative implementation?

I will have a close look. It bit enough people that an intervention or an alternative middleware loader seems necessary.

jsa-aerial:
[...]

Issue #678, opened 2020-06-16
Weird test failures


bbatsov:
After https://github.com/clojure-emacs/cider-nrepl/commit/ed793166f8d33a2e18939cafcc2e1951c1cc890c the build started failing in a really weird way (like ClojureScript/Piggieback's missing). I can't see any connection between disabling suitable and the failure's, though.

bbatsov:
@arichiardi Any idea what might be causing this?

arichiardi:
I knew suitable has some shadow-specific parts, not really in the meta/info part...it smells like the possible problem but I will definitely take a look when I have some time.

bbatsov:
I'm just not sure those failures are even related to suitable. Anyways, any help would be welcome!

bbatsov:
@clojure-emacs/cider Any ideas from anyone? Sorry about the ping, but I just came back to this ticket and I still can't figure out what broke everything cljs-related. It's like in the tests Piggieback doesn't run at all.

bbatsov:
Fixed in #691. Turns out it was suitable indeed.

vemv:
Issue good to close?

bbatsov:
Indeed.

PR #684, opened 2020-11-22
Fix delayed middleware loading issue

Cartmanishere:
Hi,

This is an attempt to fix issue around delayed middleware loading.
The relevant discussion is present at https://github.com/clojure-emacs/cider-nrepl/issues/533 

Instead of adding the handler delay to the `delayed-handlers` atom in the macro, this takes care of adding
it in the atom through a function. 

Fixes:
https://github.com/clojure-emacs/cider-nrepl/issues/447
https://github.com/clojure-emacs/cider-nrepl/issues/533
https://github.com/clojure-emacs/cider-nrepl/issues/657

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### Bugs Fixed
+
+* [#684](https://github.com/clojure-emacs/cider-nrepl/pull/684): Fix delayed middleware loading issue.
+
 ## 0.25.4 (2020-10-08)
 
 ### Changes
--- a/src/cider/nrepl.clj
+++ b/src/cider/nrepl.clj
@@ -40,19 +40,27 @@
   (or (resolve sym)
       (throw (IllegalArgumentException. (format "Cannot resolve %s" sym)))))
 
+(defn- handler-future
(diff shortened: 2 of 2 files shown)


Cartmanishere:
Test for this are failing. The failures seem related to `cljs` and are not caused by the changes in this PR.

bbatsov:
Did you manage to confirm that the suggested fix solves the race conditions? //cc @vspinu 

Also - this needs a changelog entry. 

> Test for this are failing. The failures seem related to cljs and are not caused by the changes in this PR.

Yeah, there's a ticket for this #678 but I haven't been able to find time to investigate things further.

Cartmanishere:
This is not a race condition per se. The problem is that we are relying on compile time side effects to be present at runtime. This works in normal scenario but breaks when the middlewares are compiled from within an application. 

So to fix this, instead of relying on compile time side effects, we handle it at runtime. This fixes the issue. https://github.com/clojure-emacs/cider-nrepl/issues/447 has a reproducible example against which you can check.

bbatsov:
Got it. Thanks for tackling this! 🙇‍♂️ I'll cut a new release right away, as that is an important fix.

timvisher:
_This_ is seriously exciting. 😂

PR #691, opened 2021-04-05
Fix the build

vemv:
Fixes the build by:

* "soft-reverting" the `suitable` removal
  * it's now only enabled in test environments
  * that way, the tests will pass again, while not affecting end users
* Disabling a specific clj-kondo pass for now
  * https://github.com/clj-kondo/clj-kondo/issues/1247
* Fixing a broken test related to `slurp`
  * I can't make a super precise analyis but a few things seemed off:
    * `(t/is (= ["application/octet-stream" {}]))` **always** passes (there's just one hand to the `=`)
    * The result of hitting `https://clojure.org/no-such-page` is subject to change as it's a page outside CIDER's control
    * Anyway, the broken test revealed (?) that hitting a 404 page would throw an exception
      * It seems a good thing to handle that exception
      * Unknown: is the caught result usable for end users? (haven't tried)

### DIFF
--- a/.clj-kondo/config.edn
+++ b/.clj-kondo/config.edn
@@ -4,6 +4,7 @@
                                          (cider.nrepl.middleware.util.instrument/instrument-special-form)
                                          (cider.nrepl.middleware.util.instrument/instrument-coll)
                                          (cider.nrepl.print-method/def-print-method)]}
+           :unused-import {:level :off}
            :unresolved-namespace {:exclude [clojure.main]}}
  :output {:progress true
           :exclude-files ["data_readers" "tasks"]}}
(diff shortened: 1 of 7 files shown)


vemv:
Verified green build over my personal Circle account:

image

bbatsov:
I wonder why the CI status is not visible on this PR, but I'll take your word for it. :-)

Big thanks for looking into this! 🙇‍♂️ I guess I'll have to do some thinking about the fate of the `suitable` dep - now that it's not maintained and it's broken on shadow-cljs perhaps we should just remove it altogether and go back to the old completion logic.

vemv:
🍻!

> I wonder why the CI status is not visible on this PR, but I'll take your word for it. :-)

I think that Circle will just build in one of the two possible upstreams for a fork, if the fork itself is configured to build on Circle (which is what I did)

PR #693, opened 2021-04-09
Run clj-kondo by a more stable means

vemv:
Running `curl | bash` didn't seem exactly ideal, and also it seems a good idea to run a stable, fixed version.

Finally it's also good to stick to vanilla JVM clojure, removing the possibility of graalvm doing something funny.

Green build in my personal account

image

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -59,10 +59,6 @@ commands:
           command: |
             sudo apt-get install make
       - run:
-          name: Install clj-kondo
-          command: |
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -177,10 +173,6 @@ workflows:
       - util_job:
           name: Code Linting
           steps:
-            - run:
```
vemv, line 180: Probably it's best to let Eastwood run last (as it's the slowest linter of the bunch), for a faster feedback loop

bbatsov:
I like the suggest change. Thanks!

PR #695, opened 2021-04-10
Fix debugger quit for http handler functions

### LINKED ISSUE #689: Debugger breaks on quit when debugging function called via http
## Expected behavior
Pressing **q** (for quit) when debugging a function called indirectly via http should quit the debugger without causing future issues. 

## Actual behavior
Pressing **q** (for quit) when debugging a function called indirectly via http causes an error:
**Execution error (NullPointerException) at cider.nrepl.middleware.debug/abort! (debug.clj:143)**

After this error occurs, all future attempts to invoke the debugger fail silently. 

1. Set up a simple ring server e.g. `git clone https://github.com/danielneal/cider-debugger-issue`
2. Connect to the problem e.g. using `cider-jack-in`
3. Run `cider-debug-defun-at-point` with the point in the body of the `handler` function. 
4. Start the server by running the `start-server` command in the `comment` block. 
5. Call the server through http, using the `http/get` command in the `comment` block. 
6. The cider debugger will appear, press **q**uit. This will cause an error. 
7. Repeat step 4, the cider debugger will no longer appear. 

### cider-nrepl version
0.25.5

### Java version
1.8

### Operating system
Mac OS X 10.15.7 Catalina

Cartmanishere:
Fixes #689 

In the `break` function, we reset the `:skip` in `STATE` when the debuggable function is called again. 
This is checked using the `coor` as follows -- 
```clojure
(if-let [first-coor @(:session-id STATE__)]
    (when (= first-coor coor)
      (reset! (:skip STATE__) false)
    (reset! (:session-id STATE__) coor))
```

But in case of functions called via HTTP -- 
- We skip breakpoints not using the `STATE__` but by setting `*skip-breaks*` atom as `{:mode :all}`. 
- Hence along with resetting `STATE__` on function being called again, we also need to reset the `*skip-breaks*` atom. 

This fixes the behavior of `:continue-all` and `:quit` commands for functions being called via HTTP.

### DIFF
--- a/src/cider/nrepl/middleware/debug.clj
+++ b/src/cider/nrepl/middleware/debug.clj
@@ -501,8 +501,12 @@ this map (identified by a key), and will `dissoc` it afterwards."}
   response with `read-debug-command`'."
   [coor val locals STATE__]
   (if-let [first-coor @(:session-id STATE__)]
+    ;; Check if the instrumented function is being evaluated
+    ;; from the root again.
     (when (= first-coor coor)
-      (reset! (:skip STATE__) false))
+      ;; Clear any previously set skip state.
+      (reset! (:skip STATE__) false)
+      (skip-breaks! false))
     (reset! (:session-id STATE__) coor))
   (cond
     (skip-breaks? coor STATE__) val
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## master (unreleased)
 
+### Bugs Fixed
+* [#695](https://github.com/clojure-emacs/cider-nrepl/pull/695): Fix debugger quit for http handler functions
+
 ## 0.25.10 (2021-04-08)
 
 ### Bugs Fixed


src/cider/nrepl/middleware/debug.clj
```diff
@@ -502,7 +502,8 @@ this map (identified by a key), and will `dissoc` it afterwards."}
   [coor val locals STATE__]
   (if-let [first-coor @(:session-id STATE__)]
     (when (= first-coor coor)
-      (reset! (:skip STATE__) false))
+      (reset! (:skip STATE__) false)
+      (skip-breaks! false))
```
bbatsov, line 506: Can you also add some explanatory comments here, so this part of the code is easier to understand for future readers?

bbatsov:
Great catch! I'm sure that a lot of people will be thrilled by this fix.

src/cider/nrepl/middleware/debug.clj
Cartmanishere: Sure, I'll add comment.

bbatsov:
Thanks!

PR #698, opened 2021-04-23
Add undef-all op

yuhan0:
The purpose of this op is to support a complete "reload" of a file or `ns` form in Cider (C-u M-x cider-load-file) and other clients.
Which is sometimes necessary when you have conflicting aliases or old/renamed vars sticking around in the REPL state.

----

- [x] All tests are passing

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### New features
+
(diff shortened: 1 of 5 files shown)


bbatsov:
Looks like a good change to me. Don't forget to run `lein docs` to also regen the op docs.

bbatsov:
Thanks! I'll push a new snapshot right away.

PR #699, opened 2021-07-18
Address most reflection warnings

vemv:


project.clj
```diff
@@ -140,8 +140,14 @@
                          {:dependencies [[clj-kondo "2021.03.31"]]}]
 
              :eastwood [:test
-                        {:plugins [[jonase/eastwood "0.4.0"]]
+                        {:plugins [[jonase/eastwood "0.9.1"]]
```
vemv, line 143: New linter https://github.com/jonase/eastwood#reflection

project.clj
```diff
@@ -140,8 +140,14 @@
...
                          :eastwood {:config-files ["eastwood.clj"]
                                     :exclude-namespaces [cider.nrepl.middleware.test-filter-tests]
                                     :ignored-faults {:unused-ret-vals-in-try {cider.nrepl.middleware.profile-test [{:line 25}]}
+                                                     ;; This usage of `proxy` can't avoid reflection warnings given that the `proxy` construct dispatches based on name only:
+                                                     :reflection {cider.nrepl.middleware.out [{:line "55"}
```
vemv, line 148: These strings should be numbers - it's a minor bug I just found out now

(will fix later today)

project.clj
bbatsov: Nice!

project.clj
vemv: > (will fix later today)

done && amended!

project.clj
```diff
@@ -140,8 +140,14 @@
                          {:dependencies [[clj-kondo "2021.03.31"]]}]
 
              :eastwood [:test
-                        {:plugins [[jonase/eastwood "0.4.0"]]
+                        {:plugins [[jonase/eastwood "0.9.2"]]
                          :eastwood {:config-files ["eastwood.clj"]
                                     :exclude-namespaces [cider.nrepl.middleware.test-filter-tests]
                                     :ignored-faults {:unused-ret-vals-in-try {cider.nrepl.middleware.profile-test [{:line 25}]}
+                                                     ;; This usage of `proxy` can't avoid reflection warnings given that the `proxy` construct dispatches based on name only:
+                                                     :reflection {cider.nrepl.middleware.out [{:line 55}
```
bbatsov, line 148: An idea for the future - inline comments (or something like this) to supress some linters. I think it's going to be more robust than relying on the line numbers, which are bound to change at some point and will have to be updated.

bbatsov:
Thanks!

project.clj
vemv: Cheers. I like that in a way the current approach is extremely simple - no magic comment syntax (which might be cleaned up by people or formatters), no metadata that might alter program semantics.

We leave Clojure just like it is, so that anyone maintaining the code doesn't have to scratch his head wondering what a given magic comment means, its scope (LOC, block, defn), etc.

With this said, I can think of a meta-linter that would fail if `:reflection {cider.nrepl.middleware.out [{:line 55}` silenced nothing, which would mean either `:line 55` drifted, or the underlying thing was fixed. This way one would guaranteedly keep these correct at all times.

bbatsov: Yeah, that sounds reasonable. I have something similar in RuboCop for inline directives that don't do anything.

PR #701, opened 2021-08-09
Make `stacktrace/ns-common-prefix` work in presence of user.clj files

vemv:
Prior to this fix, the cider-stacktrace "project-only" filter could be easily broken in presence of user.clj/dev.clj, because those single-segment namespaces would break the commonality between all other namespaces.

Also, performance is improved. The old `(map vector common ns)` created two vectors of chars, an operation to be repeated once per ns in the project. This is memory-hungry, even more so with the related `take-while` which is lazy.

An approach based on `reduce`/`reduce-kv`/`reduced` is introduced instead.

Parallelism is introduced as well.

Finally, `def directory-namespaces` is changed for `defn directory-namespaces`. The set of directory namespaces is always subject to change, so one has to ensure up-to-date accuracy on every invocation.

Accordingly, I had to perform adaptations throughout `middleware/stacktrace.clj` and `middleware/test.clj`, making sure that `(directory-namespaces)` is computed once per middleware op (instead of once _per frame_ - woud be quite expensive!).

### DIFF
--- a/src/cider/nrepl/middleware/stacktrace.clj
+++ b/src/cider/nrepl/middleware/stacktrace.clj
@@ -183,9 +183,9 @@
   "Flag the frame if it is from the users project. From a users
   project means that the namespace is one we have identified or it
   begins with the identified common prefix."
-  [{:keys [ns] :as frame}]
+  [namespaces {:keys [ns] :as frame}]
   (if (and ns
-           (or (contains? (directory-namespaces) (symbol ns))
+           (or (contains? namespaces (symbol ns))
                (when (:valid @ns-common-prefix)
                  (.startsWith ^String ns (:common @ns-common-prefix)))))
     (flag-frame frame :project)
@@ -206,15 +206,18 @@
 
 (defn analyze-frame
   "Return the stacktrace as a sequence of maps, each describing a stack frame."
-  [frame]
-  ((comp flag-repl flag-project analyze-fn analyze-file stack-frame) frame))
+  [namespaces frame]
+  (let [f (comp flag-repl (partial flag-project namespaces) analyze-fn analyze-file stack-frame)]
+    (f frame)))
 
 (defn analyze-stacktrace
   "Return the stacktrace as a sequence of maps, each describing a stack frame."
   [^Exception e]
-  (-> (map analyze-frame (.getStackTrace e))
-      (flag-duplicates)
-      (flag-tooling)))
+  (let [namespaces (directory-namespaces)]
+    (-> (pmap (partial analyze-frame namespaces)
+              (.getStackTrace e))
+        (flag-duplicates)
+        (flag-tooling))))
 
 ;;; ## Causes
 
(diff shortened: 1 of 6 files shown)


src/cider/nrepl/middleware/stacktrace.clj
```diff
@@ -128,14 +128,17 @@
...
-                (let [ns (str/lower-case ns)
-                      matched (map vector common ns)
-                      coincident (take-while (fn [[a b]] (= a b)) matched)]
-                  (apply str (map first coincident))))
-              (str/lower-case (first directory-namespaces))
-              directory-namespaces)
+        (try (->> directory-namespaces
+                  ;; remove usually present namespaces
+                  ;; that would break the commonality:
+                  (remove #{'dev 'user})
```
vemv, line 134: Maybe this can be generalized to "single-segment namespace", for covering unforeseen patterns as well.

I think I see other single-segment namespaces from time to time e.g. `test-runner` or `test-setup`

src/cider/nrepl/middleware/stacktrace.clj
```diff
@@ -128,14 +128,17 @@
...
+                  (reduce
+                   (fn [common ns]
+                     (let [ns (str/lower-case ns)
+                           matched (map vector common ns)
```
vemv, line 138: Maybe while I'm here I could change `map vector` (which converts a string into a vec of chars) into something else.

Else the algo doesn't seem very efficient, especially for projects with many namespaces.

Can add test coverage.

src/cider/nrepl/middleware/stacktrace.clj
bbatsov: I agree. It's better to make this more robust.

src/cider/nrepl/middleware/stacktrace.clj
bbatsov: > Maybe while I'm here I could change map vector (which converts a string into a vec of chars) into something else.

Fine by me.

bbatsov:
Don't forget to mention the fix in the changelog.

vemv:
🍻

Ready, changeloged, improved with a touch of https://github.com/clojure-emacs/cider-nrepl/issues/702 and updated linters (notably cljfmt, which wasn't able to run locally. Some cljs thing)

src/cider/nrepl/middleware/stacktrace.clj
```diff
@@ -119,43 +119,75 @@
...
   "Flag the frame if it is from the users project. From a users
   project means that the namespace is one we have identified or it
   begins with the identified common prefix."
   [{:keys [ns] :as frame}]
-  (if (and directory-namespaces ns
-           (or (contains? directory-namespaces (symbol ns))
-               (when (:valid ns-common-prefix)
-                 (.startsWith ^String ns (:common ns-common-prefix)))))
+  (if (and ns
+           (or (contains? (directory-namespaces) (symbol ns))
```
vemv, line 188: Hmm `flag-project` is invoked N times from other ops, which would result in `(directory-namespaces)` being invokes N times too. Not good.

The fix would be compute `(directory-namespaces)` once, and pass it as an argument around.

vemv:
(Ready again)

bbatsov:
Looks good! Thanks!

PR #703, opened 2021-08-15
Introduce `middleware.test/*test-error-handler*`

vemv:
This allows users to attach arbitrary functionality when tests fail due to an exception.

For example, I used this to add an Expound-backed `println` for exceptions that were caused by Spec failures.

...While CIDER already prints Spec failures nicely via an Emacs UI, not all cider-nrepl clients necessarily have such an UI.

One might also appreciate a println with Expound's format, which is familiar for many, and will be consistent with printlns placed elsewhere in a given codebase.


bbatsov:
Looks useful, but we'll also have to document it somewhere.

bbatsov:
Yeah, that'd be fine.

PR #705, opened 2021-08-31
Use session classloader when loading deferred middleware

plexus:
When loading CIDER middleware via the sideloader we need to make sure that extra
namespaces that are loaded later are also requested via the sideloader. The
original sideloader already provided a macro for
this (with-session-classloader), we just need to use it. This does mean that now
all our middlewares need to depend on the session middleware.

(see an earlier comment hinting at the fact this would still need to happen: https://github.com/nrepl/nrepl/pull/185#issuecomment-624317912)

Part of: https://github.com/clojure-emacs/cider/issues/3037

- [x] All tests are passing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 ### Changes
 
 * Parallelize `cider.nrepl.middleware.stacktrace/analyze-stacktrace`.
+* [#705](https://github.com/clojure-emacs/cider-nrepl/pull/705): Use the session classloader when loading deferred middleware
 
 ## 0.26.0 (2021-04-22)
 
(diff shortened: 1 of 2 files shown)


plexus:
I didn't add tests or docs yet for this one. wdyt @bbatsov, maybe not worth the trouble?

plexus:
Not sure what's going on here
```
Using SSH Config Dir '/home/circleci/.ssh'
git version 2.11.0
Cloning git repository
Cloning into '.'...
Warning: Permanently added the RSA host key for IP address '140.82.114.4' to the list of known hosts.
Permission denied (publickey).
fatal: Could not read from remote repository.
```

bbatsov:
No idea as well. I wonder if it's some temporary issue with Circle or our build isn't using some legacy Circle functionality.

bbatsov:
> I didn't add tests or docs yet for this one. wdyt @bbatsov, maybe not worth the trouble?

Probably we should update the docs, but we can do this in bulk at the end of the "project". Tests for this are probably an overkill. If the sideloading is working we can potentially dropped the deferred loading at some point, as we can load the middleware on demand via the sideloader, which would be nicer IMO.

PR #713, opened 2021-09-17
Fix a regression in test results formatting.

temochka:
The changes introduced in #683 resulted in test results always being
printed via `println` instead of `clojure.pprint/pprint`.

BetterThanTomorrow/calva#1280 describes the degraded user experience
caused by this issue.


test/clj/cider/nrepl/middleware/test_test.clj
```diff
@@ -152,7 +152,11 @@
     (is (= "custom-output\n"
            (#'test/print-object (with-meta {:not :printed} {:type ::custom}))))
     (is (= "{:a :b, :c :d}\n"
```
temochka, line 154: This is a somewhat deceitful example, because a map of symbols printed by `println` and `pprint` looks identical:

```
user=> (println {:a :b :c :d})
{:a :b, :c :d}
nil
user=> (clojure.pprint/pprint {:a :b :c :d})
{:a :b, :c :d}
nil
```

the differences become apparent once you add strings, symbols, and other data:

```
user=> (println {:a "string" :c "42" :d 'symbol})
{:a string, :c 42, :d symbol}
nil
user=> (clojure.pprint/pprint {:a "string" :c "42" :d 'symbol})
{:a "string", :c "42", :d symbol}
nil
```

src/cider/nrepl/middleware/test.clj
```diff
@@ -67,10 +67,12 @@
 (defn- print-object
   "Print `object` using pprint or a custom print-method, if available."
   [object]
-  (let [print-fn (if (= (get-method print-method (:type (meta object)))
-                        (get-method print-method :default))
-                   pp/pprint
-                   println)]
+  (let [custom-type (:type (meta object))
```
temochka, line 70: `(:type (meta object))` is clearly `nil` for any standard Clojure data type and `(get-method print-method nil)` is never `(get-method print-method :default)` (it is `(get-method print-method nil)`).

vemv: Can you add a test that explicitly exercises this dispatch?

src/cider/nrepl/middleware/test.clj
```diff
@@ -67,10 +67,12 @@
...
+        print-fn (if (and custom-type
```
vemv, line 71: I don't understand much what's going on here. 

* What's a `custom-type`?
  * i.e. worth adding a comment
* why can custom types result in a vanilla `println`?
  * I'm not sure of how 'customness' relates to the choice of `println`.

test/clj/cider/nrepl/middleware/test_test.clj
vemv: Nice observation! These tests are good to keep, however please write them in the form `(is (= ...) "pprint is chosen, as it can be seen in the quotation marks")`

test/clj/cider/nrepl/middleware/test_test.clj
```diff
@@ -152,7 +152,11 @@
...
-           (#'test/print-object {:a :b :c :d})))))
+           (#'test/print-object {:a :b :c :d})))
+    (is (= "{:a \"b\", :c \"d\"}\n"
+           (#'test/print-object {:a "b" :c "d"})))
+    (is (= "\"string\"\n"
```
vemv, line 158: I'd add another case in which a `\n` can be seen. That makes it more even more evident that `pprint` is being used.

test/clj/cider/nrepl/middleware/test_test.clj
```diff
@@ -152,7 +152,11 @@
...
+           (#'test/print-object "string")))))
```
vemv, line 159: As hinted in `Can you add a test that explicitly exercises this dispatch?` I'd suggest that one also tests that one can get a vanilla `print`ln by passing the right option.

IOW, I should see a test case for `{:a string, :c 42, :d symbol}` _and_ another for `{:a "string", :c "42", :d symbol}` per https://github.com/clojure-emacs/cider-nrepl/pull/713/files#r710635100

vemv (commented):
Thanks for this one! Looking good.

A few suggestions for shipping this one with full confidence.

src/cider/nrepl/middleware/test.clj
temochka: See #683 for more details on that. I’m not sure myself, to be honest. I don’t understand how exactly the `:type` meta attribute gets assigned to a value, but I don’t want to repeat #683’s mistakes and break the formatting for that case. Basically, my goal here is to keep the existing tests passing.

temochka:
@vemv I made the changes that you requested, attempting to preserve the logic added in #683 unchanged while fixing the regression. That said, if you want my honest opinion: #683 should be completely reverted. I don’t understand that change, and I have a suspicion that it doesn’t work at all. That `:type` meta attribute seems made up: `print-method` dispatches based on value type, not metadata. E.g., this works as I expect:

``` clojure
(= (get-method print-method "foo")
   (get-method print-method :default))
; => true

(= (get-method print-method {})
   (get-method print-method :default))
; => true
```

maybe the below would work instead, but if it’s in fact broken, I’d rather revert than try to fix something that has been broken and nobody complained 🤷‍♂️ 

```clojure
(defn- print-object
  "Print `object` using pprint or a custom print-method, if available."
  [object]
  (let [print-fn (if (not= (get-method print-method object)
                           (get-method print-method :default))
                   println
                   pp/pprint)]
    (with-out-str (print-fn object))))
```

vemv:
Thanks for the iteration! Looking great.

Taking a quick look at the https://github.com/clojure-emacs/cider-nrepl/pull/683 rationale, it seems a good idea to try exercising `matchers-combinators` or at least try getting an idea of how `matchers-combinators` prints things, how https://github.com/clojure-emacs/cider-nrepl/pull/683 intended to handle it, etc.

i.e. if we can ensure the matchers-combinators integration will keep working (or identify that it wasn't working properly), so much the bettter.

We can always ping the author in question @FelipeCortez (done!)

temochka:
@vemv I spent a little more time looking into this and turned out I was wrong about this:

> `print-method` dispatches based on value type, not metadata

Example:

```clojure
(defmethod clojure.core/print-method ::custom [_ out]
  (binding [*out* out]
    (print "custom-output")))

(println (with-meta {:foo "bar"} {:type ::custom}))

; custom output
; nil
```

Overall, I don’t feel like I have enough context (or proper motivation) to propose changes to existing matchers-combinators support. This PR fixes an issue affecting the majority of cider-nrepl users while satisfying the tests (and preserving the logic) provided for matchers-combinators support. To me, it sounds like a win-win. Can we merge it and track any potential matchers-combinators issues separately?

vemv:
Hi,

> Can we merge it and track any potential matchers-combinators issues separately?

I don't think so, we can't be casually wrong about things and still go ahead with our original plans. A great PR will fix things for everyone.

If you'd like to give the complete task a shot let us know!

Cheers - V

vemv (commented):
This PR is on the right track, following https://github.com/BetterThanTomorrow/calva/issues/794 I can relate that issue to weird output formats I've seen in certain CIs when tests fail.

Suggesting a simple actionable thing for the test coverage, after that it would be probably ready.

Maybe we'd need an integration test that can't be fooled so easily, but that can be tackled separately.

temochka:
> A great PR will fix things for everyone.

@vemv fair enough. I’m going on vacation until later next week so don‘t expect any updates in the meantime. I’ll be thinking about the following things (feedback welcome):

* is the reference check against `(get-method print-method :default)` sufficient for detecting a user-defined `print-method`?
* does #683 serve exclusively matcher-combinators users or there are indeed other libs that rely on this?
* related to ☝️, can we target specifically matcher-combinators metadata?
* does it make sense to use `println` (discards string quotes, generally intended for user-facing values) and not `prn`, which would at least print values in a way that can be fed back to REPL?

vemv:
Cheers. I wonder if we could ask the matchers-combinators authors to use something ns-qualified in addition / instead of :type if that helps?

vemv:
Released in `[cider/cider-nrepl "0.27.3"]`!

PR #715, opened 2021-09-30
Upgrade some dependencies

vemv:
Mainly updates `suitable`, for shipping https://github.com/clojure-emacs/clj-suitable/issues/22

@bbatsov : please let's cut a minor release with this one 🙏

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -10,11 +10,11 @@
                  ^:inline-dep [mvxcvi/puget "1.3.1"]
                  ^:inline-dep [fipp "0.6.24"] ; can be removed in unresolved-tree mode
                  ^:inline-dep [compliment "0.3.11"]
-                 ^:inline-dep [org.rksm/suitable "0.4.0" :exclusions [org.clojure/clojurescript]]
+                 ^:inline-dep [org.rksm/suitable "0.4.1-alpha1" :exclusions [org.clojure/clojurescript]]
                  ^:inline-dep [cljfmt "0.8.0" :exclusions [org.clojure/clojurescript]]
                  ^:inline-dep [org.clojure/tools.namespace "1.0.0"]
                  ^:inline-dep [org.clojure/tools.trace "0.7.10"]
-                 ^:inline-dep [org.clojure/tools.reader "1.3.2"]]
+                 ^:inline-dep [org.clojure/tools.reader "1.3.6"]]
   :exclusions [org.clojure/clojure] ; see Clojure version matrix in profiles below
 
   :plugins [[thomasa/mranderson "0.5.3"]]
--- a/project.clj
+++ b/project.clj
@@ -5,12 +5,12 @@
             :url "http://www.eclipse.org/legal/epl-v10.html"}
   :scm {:name "git" :url "https://github.com/clojure-emacs/cider-nrepl"}
   :dependencies [[nrepl "0.8.3"]
-                 ^:inline-dep [cider/orchard "0.7.2" :exclusions [com.google.code.findbugs/jsr305 com.google.errorprone/error_prone_annotations]]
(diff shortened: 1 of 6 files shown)


vemv:
Mmm, the tests will need some adapting

vemv:
The culprit of the failing build is a string that is not being mrandersonized. This is what I see after running `make test`:

```
srcdeps/cider/nrepl/inlined_deps/suitable/v0v4v1_alpha1/suitable/js_completions.clj
21:     ;;   suitable.js-introspection/property-names-and-types
22:     (let [template "(suitable.js-introspection/property-names-and-types ~A ~S)"
24:       (cljs-eval-fn ns "(require 'suitable.js-introspection)")
```

@bbatsov @benedekfazekas is there a way to tell mranderson to prefix `"(require 'suitable.js-introspection)"` as seen in L24?

I can imagine why a string might not be updated, out of caution. Or it might be a vanilla bug, IDK

vemv:
Nevermind, I debugged it (`suitable` had some existing setup for mranderson, but a couple places where it was missing)

So all `clj-suitable` problems are fixed in CI apparently.

I'm seeing these however which seem unrelated:

```
FAIL in (missing-info-test) (info_test.clj:318)
ensure info returns a no-info packet if ns not found
expected: (= (:status response) #{"no-info" "done"})
  actual: (not (= #{"done"} #{"no-info" "done"}))

lein test :only cider.nrepl.middleware.info-test/missing-eldoc-test

FAIL in (missing-eldoc-test) (info_test.clj:335)
ensure eldoc returns a no-eldoc packet if ns not found
expected: (= (:status response) #{"no-eldoc" "done"})
  actual: (not (= #{"done"} #{"no-eldoc" "done"}))
```

Do these ring a bell? @bbatsov

vemv:
The failing tests were already failing as of `master`, they're an instance of https://github.com/clojure-emacs/orchard/issues/133

They represent a pretty minor bug so if possible let's publish a minor ,release with this one so that we can have `suitable` working asap.

bbatsov:
> Mainly updates suitable, for shipping clojure-emacs/clj-suitable#22

Oh, shit! I totally forgot we hadn't updated this yet. I see however that a lot more tests are failing than were failing on `master`. There the failures seemed like some build flakiness.

project.clj
```diff
@@ -58,8 +63,14 @@
                                     :password :env/clojars_password
                                     :sign-releases false}]]
 
-  :profiles {:provided {:dependencies [[org.clojure/clojure "1.10.1"]
-                                       [org.clojure/clojurescript "1.10.520" :scope "provided"]]
+  :profiles {:provided {:dependencies [[org.clojure/clojure "1.10.3"]
+                                       [org.clojure/clojurescript "1.10.866" :scope "provided"]
+                                       [com.cognitect/transit-clj "1.0.324"]
```
bbatsov, line 68: What are all those extra provided deps for?

vemv: These are the latest transitive deps from clojurescript, satisfying  `:pedantic?`

vemv (commented):
> I see however that a lot more tests are failing than were failing on master.

Just in case, finally they fail no more than in master.

The test failures essentially tell us that querying eldoc for `non.existing/+` will show the doc for `clojure.core/+`. Seems like we can live with that for a few days

vemv:
Build is green again. Will be able to add a couple extra test cases after https://github.com/clojure-emacs/orchard/pull/134 is merged/released but that's not related to this PR

project.clj
bbatsov: Got it. Might be a good idea to add some note about this, as I'm likely to forget otherwise. I never used `:pedantic?` myself.

bbatsov:
Sounds good. I guess we should also cut `suitable` 0.4.1 first, as it'd be a bit weird to use an alpha version for the bugfix release.

vemv:
All green, upgraded, with increased coverage and released ✌️

Issue #716, opened 2021-10-01
StackOverflowError in cider.nrepl.middleware.out


lassemaatta:
## Expected behavior

Logging with `clojure.tools.logging` should work when running application over emacs/cider repl.

## Actual behavior

A StackOverflowException is thrown when logging is attempted:

```
...
[java.io.PrintWriter write "PrintWriter.java" 542]
[jdk.internal.reflect.GeneratedMethodAccessor3 invoke nil -1]
[jdk.internal.reflect.DelegatingMethodAccessorImpl invoke "DelegatingMethodAccessorImpl.java" 43]
[java.lang.reflect.Method invoke "Method.java" 566]
[clojure.lang.Reflector invokeMatchingMethod "Reflector.java" 167]
[clojure.lang.Reflector invokeInstanceMethod "Reflector.java" 102]
[cider.nrepl.middleware.out$forking_printer$fn__55622 invoke "out.clj" 59]
[cider.nrepl.middleware.out.proxy$java.io.Writer$ff19274a write nil -1]
[java.io.PrintWriter write "PrintWriter.java" 542]
...
```

## Steps to reproduce the problem

1. Using `org.clojure/tools.logging` `1.1.0`
2. Invoke `(clojure.tools.logging/debug "hello")` within a `try` block
3. catch any Errors and see the StackOverflowError

Version `0.26.0` does not exhibit this issue (tested by overriding version in `profiles.clj`).

Also, running the application through `lein repl` does not exhibit this issue.

### cider-nrepl version

`0.27.0`

### Java version

```
openjdk version "11.0.12" 2021-07-20
OpenJDK Runtime Environment (build 11.0.12+7)
OpenJDK 64-Bit Server VM (build 11.0.12+7, mixed mode)
```

### Operating system

`Linux archlinux 5.14.7-arch1-1 #1 SMP PREEMPT Wed, 22 Sep 2021 21:35:11 +0000 x86_64 GNU/Linux`

vemv:
Thanks! I assume the section of the stracktrace is repeating? Do you see any different lines?

(Especially at its beginning)

lassemaatta:
Yeah, that section of the stack trace seemed to be the loop. I managed to grab the StackOverflowError object and `tap>` it into reveal, but the stack trace was just filled with that repeating section without any hints on where it all started.

vemv:
Thanks. With a bit of luck we might be able to repro this by exercising t.logging within our test suite.

[...]

PR #717, opened 2021-10-03
Fix stack overflow error in cider.nrepl.middleware.out

vmfhrmfoaj:
> Fixes https://github.com/clojure-emacs/cider-nrepl/issues/716

I saw following error messages when starting REPL.
```
;; on `*nrepl-server ... *` buffer
Exception: java.lang.StackOverflowError thrown from the UncaughtExceptionHandler in thread "Timer-0"

Exception: java.lang.StackOverflowError thrown from the UncaughtExceptionHandler in thread "Timer-1"
```

It was introduced at 1007408 commit.
```clojure
;; new
(defn original-output
  "Store the values of the original output streams so we can refer to them."
  ^PrintWriter
  [k]
  ({:out *out*
    :err *err*} k))

;; old
(def original-output
  "Store the values of the original output streams so we can refer to them."
  {:out *out*
   :err *err*})
```

`original-output` in `cider.nrepl.middleware.out` was changed to the function from the variable. 
If you change `*out*`/`*err*` via `alter-var-root` or another ways, `original-output` will use the changed value.

In `write`/`flush` method of proxy of `forking-printer`, call `write`/`flush` method with a value that return from `original-output`
`tracked-sessions-map-watch` replace `*out*` and `*err*` to return value of `forking-printer`.

I think this circle (i.e. infinite recursion) cuase the stack overflow.

(Sorry for the lack of explanation. Explaining in English difficult for me.)

### DIFF
--- a/src/cider/nrepl/middleware/out.clj
+++ b/src/cider/nrepl/middleware/out.clj
@@ -18,12 +18,11 @@
 
 (declare unsubscribe-session)
 
-(defn original-output
-  "Store the values of the original output streams so we can refer to them."
-  ^PrintWriter
-  [k]
-  ({:out *out*
-    :err *err*} k))
+(defonce original-output
+  ^{:doc "Store the values of the original output streams so we can refer to them.
+Please do not inline; they must not be recomputed at runtime."}
+  {:out *out*
+   :err *err*})
 
 (defmacro with-out-binding
   "Run body with v bound to the output stream of each msg in msg-seq.
--- a/project.clj
+++ b/project.clj
@@ -158,9 +158,10 @@
                                     :exclude-namespaces [cider.nrepl.middleware.test-filter-tests]
                                     :ignored-faults {:unused-ret-vals-in-try {cider.nrepl.middleware.profile-test [{:line 25}]}
                                                      ;; This usage of `proxy` can't avoid reflection warnings given that the `proxy` construct dispatches based on name only:
-                                                     :reflection {cider.nrepl.middleware.out [{:line 55}
-                                                                                              {:line 57}
-                                                                                              {:line 59}
-                                                                                              {:line 61}
-                                                                                              {:line 65}]}
+                                                     :reflection {cider.nrepl.middleware.out [{:line 54}
+                                                                                              {:line 56}
+                                                                                              {:line 58}
+                                                                                              {:line 60}
+                                                                                              {:line 62}
+                                                                                              {:line 64}]}
                                                      :suspicious-test {cider.nrepl.middleware.profile-test [{:line 25}]}}}}]})


vemv:
Makes sense, `original-output` shouldn't be a defn since it intends to keep the "original" output.

src/cider/nrepl/middleware/out.clj
```diff
@@ -18,12 +18,14 @@
 
 (declare unsubscribe-session)
 
+(defonce org-out {:out *out*
+                  :err *err*})
+
 (defn original-output
   "Store the values of the original output streams so we can refer to them."
-  ^PrintWriter
+  ^Writer
```
vemv, line 26: Let's remove the type hint entirely to make it just like the previous version.

vemv (commented):
LGTM, please update project.clj adding all the necessary silencings ([example](https://github.com/clojure-emacs/cider-nrepl/blob/05be239f2d66b2f035af0afe1b40e6af3086eb2f/project.clj#L161-L165))

You can run `make eastwood` to progress locally fast.

Thanks!

src/cider/nrepl/middleware/out.clj
```diff
@@ -18,12 +18,14 @@
 
 (declare unsubscribe-session)
 
+(defonce org-out {:out *out*
+                  :err *err*})
```
vemv, line 22:
```suggestion
(defonce
  ^{:doc "The original streams. Please do not inline;
they must not be recomputed at runtime."}
  org-out
  {:out *out*
   :err *err*})
```

vmfhrmfoaj:
I reverted to previous version and udpated `:eastwood` in project.clj.

vemv:
Thanks! Still, let's insert the `defonce` as suggested above.

(that might need updating project.clj with silencings again, `make eastwood` can help as mentioned)

src/cider/nrepl/middleware/out.clj
vmfhrmfoaj: I think if don't use type hint, no reason to wrap a map with the function.

vemv: Ok, let's prefer `defonce` to `def` and use the following docstring:

```
Store the values of the original output streams so we can refer to them.
Please make sure they are not recomputed at runtime.
```

vemv: (The idea is that we make it as hard as possible to make the same mistake again)

vemv (approved):
Looks like you'll have to make `make cljfmt` happy and we'll be good to go👍

vmfhrmfoaj:
I updated. Thank you.

vemv:
Thanks to you 🍻

vemv:
Released cider-nrepl 0.27.2. You can set it in cider.el [via](https://github.com/clojure-emacs/cider/blob/d61fff06b4c012de790646c003065e8aaeec016e/cider.el#L420).

PR #720, opened 2021-10-05
`middleware.test`: gracefully handle exceptions thrown within fixtures

vemv:
Fixes https://github.com/clojure-emacs/cider-nrepl/issues/719 (moved from the cider.el issue tracker)


bbatsov (approved):
Looks good to me, apart that remark about the public fallback name var.

PR #729, opened 2021-11-12
fix: report test assertion line on exception

hugoduncan:
Currently, when a clojure test assertion fails with an exception, the
:line reported is for the start of the test function.

This changes the reported line to the first line of any function within
the test function.

### DIFF
--- a/src/cider/nrepl/middleware/test.clj
+++ b/src/cider/nrepl/middleware/test.clj
@@ -7,6 +7,7 @@
    [cider.nrepl.middleware.util :as util]
    [cider.nrepl.middleware.util.coerce :as util.coerce]
    [clojure.pprint :as pp]
+   [clojure.string :as string]
    [clojure.test :as test]
    [clojure.walk :as walk]
    [nrepl.middleware.interruptible-eval :as ie]
@@ -65,7 +66,8 @@
           .getStackTrace
           (map (partial st/analyze-frame namespaces))
           (filter
-           #(clojure.string/starts-with? (:class %) (.getName (class f))))
+           #(some->  (:class %)
+                     (clojure.string/starts-with? class-name)))
           first))))
 
 (defn- print-object
--- a/src/cider/nrepl/middleware/test.clj
+++ b/src/cider/nrepl/middleware/test.clj
@@ -64,7 +64,8 @@
      (->> e
           .getStackTrace
           (map (partial st/analyze-frame namespaces))
-          (filter #(= (:class %) class-name))
+          (filter
+           #(clojure.string/starts-with? (:class %) (.getName (class f))))
           first))))
 
 (defn- print-object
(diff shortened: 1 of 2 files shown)


src/cider/nrepl/middleware/test.clj
```diff
@@ -64,7 +64,8 @@
      (->> e
           .getStackTrace
           (map (partial st/analyze-frame namespaces))
-          (filter #(= (:class %) class-name))
+          (filter
+           #(clojure.string/starts-with? (:class %) (.getName (class f))))
```
vemv, line 68: `(.getName (class f))` doesn't appear to be different from `class-name`, see surrounding when-let

vemv, line 68: I'd use some-> out of caution

(`=` doesn't fail on nils, `starts-with?` does)

vemv (commented):
Some coverage would seem desirable

hugoduncan:
@vemv thanks for the quick feedback :)

src/cider/nrepl/middleware/test.clj
```diff
@@ -64,7 +65,9 @@
      (->> e
           .getStackTrace
           (map (partial st/analyze-frame namespaces))
-          (filter #(= (:class %) class-name))
+          (filter
+           #(some->  (:class %)
+                     (clojure.string/starts-with? class-name)))
```
vemv, line 70:
```suggestion
           #(some-> (:class %)
                    (string/starts-with? class-name)))
```

vemv (commented):
@hugoduncan  I was thinking, what if `string/starts-with?` can't find anything but the old impl could?

For that case, having the old impl as a fallback would be useful. Admittedly not perfect (as the line wouldn't be as accurate), but better than finding nothing.

hugoduncan:
@vemv how would that be possible? Isn't `starts-with?` going to match a superset of things matched by `=`?

I suppose I'm still not checking if `(:class %)` is a string.

vemv:
>  Isn't starts-with? going to match a superset of things matched by =?

Sounds right. However the 'superset' nuance brings another thing to mind:

what if we're matching an unrelated class? For example `FooBar` starts with `Foo`.

And, unrelatedly, reading the PR desc again:

> This changes the reported line to the first line of any function within
the test function.

this means we can return a randomly selected candidate, right? a deftest can have many defns within it.

Probably clients would appreciate a stable result?

hugoduncan:
> what if we're matching an unrelated class? For example `FooBar` starts with `Foo`.

The test function is a nested class (has a `$` in the name), so we can only match classes that nest within the function

> this means we can return a randomly selected candidate, right? a deftest can have many fns within it.

No, first in stack order, ie. the lowest down the call stack.

vemv (approved):

vemv:
Alright! Happy we made sure 🍻

Issue #721, opened 2021-08-27
Cider-macroexpand-all in deps.edn project of definterface crashes


jellelicht:
## Expected behavior
Being able to macro expand forms that happen to define interfaces, in either deps.edn or leiningen projects

## Actual behavior
A crash pop-up when macro expanding forms that happen to define interfaces, but only in deps.edn projects.

## Steps to reproduce the problem

I made a tiny example that seems to be reproducible for at least one other person:
 https://github.com/jellelicht/cider-macro-repro

## Steps to reproduce the problem (alt)

`(macroexpand-1 '(definterface IFoo))` with cider-nrepl's machinery

### CIDER version information

```
; Connected to nREPL server - nrepl://localhost:44805                                                                                 
;; CIDER 1.1.1 (Plovdiv), nREPL 0.8.3                                                                                                  
;; Clojure 1.10.3, Java 12
```

### Emacs version
27.2

### Operating system

GNU Guix (current master)

## Stacktrace

<details>

```
  Show: Project-Only All 
  Hide: Clojure Java REPL Tooling Duplicates  (52 frames hidden)

  This is an unexpected CIDER middleware error.
  Please submit a bug report via `M-x cider-report-bug`.

  If these stacktraces are occurring frequently, consider using the
  button(s) below to suppress these types of errors for the duration of
  your current CIDER session. The stacktrace buffer will still be
  generated, but it will "pop under" your current buffer instead of
  "popping over". The button toggles this behavior.

 Suppress macroexpand-error 

2. Unhandled java.lang.ClassCastException
   class clojure.lang.Var$Unbound cannot be cast to class
   clojure.lang.DynamicClassLoader (clojure.lang.Var$Unbound and
   clojure.lang.DynamicClassLoader are in unnamed module of loader
   'app')

              genclass.clj:  722  clojure.core/gen-interface
              genclass.clj:  688  clojure.core/gen-interface
               RestFn.java:  142  clojure.lang.RestFn/applyTo
                  Var.java:  705  clojure.lang.Var/applyTo
             Compiler.java: 6997  clojure.lang.Compiler/macroexpand1
                  core.clj: 4012  clojure.core/macroexpand-1
                  core.clj: 4014  clojure.core/macroexpand
                  walk.clj:  130  clojure.walk/macroexpand-all/fn
                  walk.clj:   65  clojure.walk/prewalk
                  walk.clj:   61  clojure.walk/prewalk
[...]

PR #731, opened 2021-12-01
cider-macroexpand: Bind `LOADER` if unbound

vemv:
Fixes https://github.com/clojure-emacs/cider-nrepl/issues/721

This issue is best summarized at https://github.com/clojure-emacs/cider-nrepl/issues/721#issuecomment-982690676

I was only able to reproduce it with tools.deps (just like OP), which is why a tools.deps-based integration test is proposed.

I verified that the test would fail the build if the fix is reverted: https://app.circleci.com/pipelines/github/clojure-emacs/cider-nrepl/500/workflows/bf930f3b-3deb-4c75-9f03-aa8c30ebca49

### DIFF
--- a/src/cider/nrepl/middleware/macroexpand.clj
+++ b/src/cider/nrepl/middleware/macroexpand.clj
@@ -113,10 +113,16 @@
   context of the given :ns, using the provided :expander and :display-namespaces
   options."
   [{:keys [code expander ns] :as msg}]
-  (->> (let [expander-fn (resolve-expander-clj expander)]
-         (binding [*ns* (find-ns ns)]
-           (expander-fn (read-string code))))
-       (walk/prewalk (post-expansion-walker-clj msg))))
+  ;; Bind LOADER, which can be unbound under certain code paths, particularly when using tools.deps:
+  (with-bindings {clojure.lang.Compiler/LOADER
+                  (if (instance? clojure.lang.Var$Unbound
+                                 @clojure.lang.Compiler/LOADER)
+                    (clojure.lang.DynamicClassLoader. (clojure.lang.RT/baseLoader))
+                    @clojure.lang.Compiler/LOADER)}
+    (->> (let [expander-fn (resolve-expander-clj expander)]
+           (binding [*ns* (find-ns ns)]
+             (expander-fn (read-string code))))
+         (walk/prewalk (post-expansion-walker-clj msg)))))
 
 ;; ClojureScript impl
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -85,6 +85,8 @@ commands:
       - save_cache:
           paths:
             - ~/.m2
+            - linux-install-1.10.3.1040.sh
+            - clojure-tools-1.10.3.1040.tar.gz
             - .cpcache
           key: clojure-<< parameters.cache_version >>-{{ checksum "/tmp/clojure_cache_seed" }}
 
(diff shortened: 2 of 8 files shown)


src/cider/nrepl/middleware/macroexpand.clj
```diff
@@ -113,10 +113,15 @@
...
   [{:keys [code expander ns] :as msg}]
-  (->> (let [expander-fn (resolve-expander-clj expander)]
-         (binding [*ns* (find-ns ns)]
-           (expander-fn (read-string code))))
-       (walk/prewalk (post-expansion-walker-clj msg))))
+  (with-bindings {clojure.lang.Compiler/LOADER
+                  (if (instance? clojure.lang.Var$Unbound
+                                 @clojure.lang.Compiler/LOADER)
+                    (clojure.lang.DynamicClassLoader. (clojure.lang.RT/baseLoader))
+                    @clojure.lang.Compiler/LOADER)}
```
vemv, line 120: This is the fix

Looks pretty boring but I could use a second pair of 👀s here. @plexus perhaps?

I'd assume this is a reasonable way to instantiate a DCL. I'd guess that it's not a leaky pattern?

bbatsov: Looks quite ugly, though, even if it's pretty simple. You might also add a note that's something specific to tools.deps, as I doubt many people will figure out why we need such gymnastics.

vemv: Definitely can add a comment.

Not 100% sure if exclusive to t.deps though; in the end that should be one of the rawest ways of running Clojure. It can also be framed as 'Lein was doing something for us that we implicitly relied on'.

Draft:

```
Bind LOADER, which can be unbound under certain code paths, particularly when using tools.deps.
```

bbatsov: Sounds reasonable to me.

vemv: Amended with the comment

plexus:
Note that in a "raw" Clojure CLI (tools.deps) REPL you don't seem to have this problem:

```
➜ clj 
Clojure 1.10.3
(ns not-user)
nil

(definterface IFoo)    
not_user.IFoo

(macroexpand-1 '(definterface IFoo))
(clojure.core/let [] (clojure.core/gen-interface :name not_user.IFoo :methods []) (clojure.core/import not_user.IFoo))

(macroexpand '(definterface IFoo))
(let* [] (clojure.core/gen-interface :name not_user.IFoo :methods []) (clojure.core/import not_user.IFoo))

(clojure.walk/macroexpand-all '(definterface IFoo))
(let* [] not_user.IFoo (do (clojure.core/import* "not_user.IFoo")))
```

`Compiler/LOADER` is being bound in Compiler.java in multiple places

```java
public static Object eval(Object form, boolean freshLoader) {
	boolean createdLoader = false;
	if(true)//!LOADER.isBound())
		{
		Var.pushThreadBindings(RT.map(LOADER, RT.makeClassLoader()));
		createdLoader = true;
...
```

```java
public static Object load(Reader rdr, String sourcePath, String sourceName) {
	Var.pushThreadBindings(
			RT.mapUniqueKeys(LOADER, RT.makeClassLoader(),
``` 

As I mentioned in https://github.com/clojure-emacs/cider-nrepl/issues/721#issuecomment-986645765 that `unbound` is misleading.

So basically whevener you `eval` something, or `load` a file, it will always be bound. This is something I didn't fully appreciate before. My impression was also that by default `LOADER` is not bound, and the context classloader is used instead (notice that `RT/baseLoader` checks both).

And in that light maybe this is something that is better fixed at the nREPL level, where we already have a session classloader, which is a DCL, and which we can bind to `LOADER` for the scope of handling nREPL messages.

Not 100% sure, would have to have a look at again what we already do or don't do in nrepl. Fixing it here is also an option, but I'm not sure the check for `Var$Unbound` really makes sense, since it can be unbound (no root binding) but still have a thread-local binding, i.e. can be deref'd.

vemv:
Sounding great 🚀  I'll mark this PR as draft to emphasize hammock time.

> I'm not sure the check for Var$Unbound really makes sense, since it can be unbound (no root binding) but still have a thread-local binding, i.e. can be deref'd.

But the exact check is:

`(instance? clojure.lang.Var$Unbound @clojure.lang.Compiler/LOADER)`

i.e. it does deref it in all cases, which would perceive a thread-local binding if it was there?

---

(maybe-obvious note, there's `clojure.core/thread-bound?` which might help?)

plexus:
> i.e. it does deref it in all cases, which would perceive a thread-local binding if it was there?

oh yes, that seems right

vemv:
I'm merging this as the fix doesn't seem problematic. There might be room for moving it to nrepl though, which OTOH would seem potentially more impactful.

PR #733, opened 2021-12-02
fix: remove reflection in forking-printer

hugoduncan:
Remove reflection to improve speed and reduce garbage creation.

- [x] All tests are passing
- [] You've updated the README (if adding/changing middleware)

### DIFF
--- a/src/cider/nrepl/middleware/out.clj
+++ b/src/cider/nrepl/middleware/out.clj
@@ -18,6 +18,8 @@
 
 (declare unsubscribe-session)
 
+(set! *warn-on-reflection* true)
+
 (defonce
   ^{:doc "Store the values of the original output streams so we can refer to them.
 Please do not inline; they must not be recomputed at runtime."}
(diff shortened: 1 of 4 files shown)


src/cider/nrepl/middleware/out.clj
```diff
@@ -18,6 +18,8 @@
 
 (declare unsubscribe-session)
 
+(set! *warn-on-reflection* true)
```
vemv, line 21: We're using Eastwood as a reflection linter (CI-backed), you can remove this while also removing all of this https://github.com/clojure-emacs/cider-nrepl/blob/c3d6b69c98ef4c81529b69fa7624662f2006f545/project.clj#L150-L156

If the fix is indeed OK, Eastwood should be happy with it

vemv (commented):
Looking good, I like the main fix - not sure if I would have thought of `cond`ing the proxy myself :)

src/cider/nrepl/middleware/out.clj
```diff
@@ -50,17 +52,42 @@ Please do not inline; they must not be recomputed at runtime."}
...
+                  ;; as `int` and `char[]` aren't supported by proxy.
                   (write
                     ([x]
-                     (.write (original-output type) x)
                      (with-out-binding [printer messages type]
-                       (.write printer x)))
+                       (cond
+                         (string? x)
+                         (do
+                           (.write ^Writer (original-output type) ^String x)
```
vemv, line 63: Extra request (sorry for the pickiness), please verify that there's coverage for all these branches or add it

Context: touching this area recently resulted in an opaque issue that impacted users.

bbatsov:
I can only add - don't forget to update the changelog. :-) I know this is isn't exactly user-visible, but I know some people would be happy to see there's a bit less reflection warnings.

vemv (approved):
Neat! Thanks for the iteration 🍻

vemv (approved):
Neat! Thanks for the iteration 🍻

PR #736, opened 2021-12-09
chore: remove unnecessary binding of current message

hugoduncan:
The binding is not used within the body expressions, and costs a thread
local binding, producing garbage.  This changes ensures print-stream
produces no memory garbage.

-------

- [x] All tests are passing

Thanks!

### DIFF
--- a/src/cider/nrepl/middleware/out.clj
+++ b/src/cider/nrepl/middleware/out.clj
@@ -34,11 +34,11 @@ Please do not inline; they must not be recomputed at runtime."}
                                                       (case ~type
                                                         :out #'*out*
                                                         :err #'*err*))]
-       (try (binding [ieval/*msg* ~'msg]
-              ~@body)
-            ;; If a channel is faulty, dissoc it.
-            (catch Exception ~'e
-              (unsubscribe-session ~'session))))))
+       (try
+         ~@body
+         ;; If a channel is faulty, dissoc it.
+         (catch Exception ~'e
+           (unsubscribe-session ~'session))))))
 
 (defn forking-printer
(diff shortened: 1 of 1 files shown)


vemv:
> The binding is not used within the body expressions

Perhaps while this is true, the binding is used by the underlying nrepl machinery?

Reading the code, there seems to be intentful work for binding `msg` to a precise thing out of a `msg-seq`. i.e. if something bothered to do that, probably it was for a reason?

If we couldn't determine the original intent, perhaps I'd default to not changing the code - a `binding`'s runtime cost surely is marginal in the context of printing stuff to `*out*` (i.e. IO-heavy workload).

Each printed line might also imply transient garbage, so in terms of memory, the `binding`'s cost also seems marginal.

bbatsov:
If the output is still properly redirected it's fine to remove this.

bbatsov:
I did notice that this code has been there since the first commit of the file. I'm guessing one reason to have it is to avoid the original msg that triggered some output being replaced by a subsequent evaluation.

hugoduncan:
My understanding is:

The `forking-printer` just takes values and forwards them.  There is no execution of user code, so it can not possibly use the message in any.  

`print-stream` invokes the forking printer `flush` every 100ms, so the current code produces garbage every 100ms.

hugoduncan:
I'm also not sure what the intended purpose of binding the local, `v`, is - it doesn't seem to get used.

hugoduncan:
I think my previous patch introduced a bug - the outout to `original-output` should happen outside the loop created by `with-out-binding`.

bbatsov:
Okay, you've convinced me. :-) Thanks!

PR #738, opened 2021-12-10
chore: use single executor service instead of timer per stream

hugoduncan:
Replace the per output-stream Timer instance with a single scheduled
thread pool executor with a single thread.

- Removes the need for a proxy object on every print-stream.

- Executor service is generally preferred over Timer.

- The thread is named so its purpose is apparent.

-------

- [x] All tests are passing

### DIFF
--- a/src/cider/nrepl/middleware/out.clj
+++ b/src/cider/nrepl/middleware/out.clj
@@ -96,6 +96,12 @@ Please do not inline; they must not be recomputed at runtime."}
 
 (let [id-counter (atom 0)]
   (def ^ScheduledExecutorService flush-executor
+    "An executor used to run flush on `print-stream` instances.
+   Using a single thread reduces resource usage, and possibly reduces
+   the interleaving of output from different streams.
+   The executor service will ensure there is always a thread available,
+   should a flush throw an exception and kill a thread.
+   Thread names are generated with id-counter, to make them unique."
     (Executors/newScheduledThreadPool
      1
      (proxy [ThreadFactory] []
--- a/src/cider/nrepl/middleware/out.clj
+++ b/src/cider/nrepl/middleware/out.clj
@@ -94,13 +94,15 @@ Please do not inline; they must not be recomputed at runtime."}
                       (.flush printer))))
                 true))
 
-(def ^ScheduledExecutorService flush-executor
-  (Executors/newScheduledThreadPool
-   1
-   (proxy [ThreadFactory] []
-     (newThread [^Runnable r]
-       (doto (Thread. r "cider-nrepl output flusher")
-         (.setDaemon true))))))
+(let [id-counter (atom 0)]
+  (def ^ScheduledExecutorService flush-executor
+    (Executors/newScheduledThreadPool
+     1
+     (proxy [ThreadFactory] []
+       (newThread [^Runnable r]
+         (doto (Thread.
+                (str "cider-nrepl output flusher " (swap! id-counter inc)))
+           (.setDaemon true)))))))
 
 (defn print-stream
   "Returns a PrintStream suitable for binding as java.lang.System/out or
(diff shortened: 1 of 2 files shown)


src/cider/nrepl/middleware/out.clj
```diff
@@ -90,6 +90,13 @@ Please do not inline; they must not be recomputed at runtime."}
                       (.flush printer))))
                 true))
 
+(def flush-executor (Executors/newScheduledThreadPool
+                     1
+                     (proxy [ThreadFactory] []
+                       (newThread [r]
+                         (doto (Thread. r "cider-nrepl output flusher")
+                           (.setDaemon true))))))
```
vemv, line 99: Perhaps the names should be numbered?

If the thread fails, it will be spawned again. Good to be able to distinguish old from new threads.

src/cider/nrepl/middleware/out.clj
```diff
@@ -100,16 +108,16 @@ Please do not inline; they must not be recomputed at runtime."}
...
   [printer]
   (let [delay 100
-        print-timer (Timer.)
-        print-flusher (proxy [TimerTask] []
-                        (run []
-                          (.flush ^Writer @printer)))]
-    (.scheduleAtFixedRate print-timer print-flusher delay delay)
+        print-flusher (fn [] (.flush ^Writer @printer))
+        flush-future (.scheduleWithFixedDelay
+                      ^ScheduledExecutorService flush-executor
```
vemv, line 113: The type hint can reside at the `def` as well

test/clj/cider/nrepl/middleware/out_test.clj
```diff
@@ -102,3 +102,20 @@
...
+(defn find-thread [thread-name]
+  (->> (Thread/getAllStackTraces)
+       (.keySet)
+       (filter #(= thread-name (.getName ^Thread %)))
+       first))
+
+(deftest print-stream-flush-test
+  (let [out-writer (StringWriter.)
+        printer (o/print-stream (volatile! out-writer))]
+    (.write printer 32)
```
vemv, line 115: `\space` perhaps?

src/cider/nrepl/middleware/out.clj
```diff
@@ -100,16 +108,16 @@ Please do not inline; they must not be recomputed at runtime."}
...
+                      print-flusher
+                      delay delay TimeUnit/MILLISECONDS)]
+
     (PrintStream.
      (proxy [OutputStream] []
        (close []
-         (.cancel print-flusher)
-         (.cancel print-timer)
+         (.cancel flush-future false)
```
vemv, line 120: why `false`? Seems a good thing to explain in a comment

src/cider/nrepl/middleware/out.clj
```diff
@@ -90,6 +91,13 @@ Please do not inline; they must not be recomputed at runtime."}
                       (.flush printer))))
                 true))
 
+(def flush-executor (Executors/newScheduledThreadPool
```
bbatsov, line 94: I think it's good to add some docstring here.

bbatsov:
The changes look good, but your branch has to be rebased on top of the current `master` branch due to merge conflicts.

test/clj/cider/nrepl/middleware/out_test.clj
hugoduncan: Actually, `\space` is a char and not an int.

hugoduncan:
Thanks for all the feedback!  I believe it has all been addressed now.

test/clj/cider/nrepl/middleware/out_test.clj
vemv: One can do `(int \space)` though 😇

Thanks for applying all the feedback regardless

PR #743, opened 2022-01-09
Upgrade Orchard, nrepl

vemv:
Upgrade [Orchard](https://github.com/clojure-emacs/orchard/blob/v0.9.0/CHANGELOG.md#090-2022-01-10), [nrepl](https://github.com/nrepl/nrepl/blob/0.9.0/CHANGELOG.md#090-2021-12-12), tools.namespace, fipp.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 * [#548](https://github.com/clojure-emacs/cider-nrepl/pull/548): Make the content-type middleware extensible via multimethod
 
+## Changes
+
+* Upgrade [Orchard](https://github.com/clojure-emacs/orchard/blob/v0.9.0/CHANGELOG.md#090-2022-01-10).
+
 ## 0.27.4 (2021-12-15)
 
 ### Bugs fixed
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,10 @@
 
 * [#548](https://github.com/clojure-emacs/cider-nrepl/pull/548): Make the content-type middleware extensible via multimethod
 
+## Changes
+
+* Upgrade [Orchard](https://github.com/clojure-emacs/orchard/blob/v0.9.0/CHANGELOG.md#090-2022-01-10), [nrepl](https://github.com/nrepl/nrepl/blob/0.9.0/CHANGELOG.md#090-2021-12-12), tools.namespace, fipp.
```
bbatsov, line 11: Only the orchard update will have some impact to the end users, so I'm not sure if we need to mention the others here. If we do - it'd be nice if they were all bullet points with the same structure (name, version, changelog).

vemv:
Thanks! Amended

bbatsov:
@vemv I guess it might be a good idea to cut 0.28 at this point, unless you have some other short-term improvements planned.

PR #744, opened 2022-01-17
Upgrade Orchard

vemv:
https://github.com/clojure-emacs/orchard/blob/v0.9.1/CHANGELOG.md#091-2022-01-17

⚠️ nrepl also deserves an upgrade in this batch, it's pending on https://github.com/nrepl/nrepl/pull/265


bbatsov:
Thanks!

PR #751, opened 2022-05-14
undef-all: skip unmapping of default imports

yuhan0:
See https://github.com/clojure-emacs/cider/issues/3194

- [x] All tests are passing


bbatsov:
Looks good to me. Just mention this bugfix in the changelog and we're good to go.

PR #754, opened 2022-06-01
Observe .cljfmt.edn files

vemv:
Soft of a draft, does it look good to you? @yuhan0 @bbatsov

- [ ] All tests are passing

Thanks!

### DIFF
--- a/src/cider/nrepl/middleware/format.clj
+++ b/src/cider/nrepl/middleware/format.clj
@@ -32,13 +32,12 @@
   [{:keys [code options ^File config-file]
     :or {config-file default-config-file}}]
   (let [config-file-opts (when (-> config-file .exists)
-                           (-> config-file slurp edn/read-string))
+                           (->> config-file slurp (edn/read-string {:readers {'re re-pattern}})))
         message-opts (some-> options
                              (select-keys [:indents :alias-map])
                              (update :indents generate-user-indents)
                              (update :alias-map #(reduce-kv (fn [m k v] (assoc m (name k) v)) {} %)))
-        opts (deep-merge {:indents fmt/default-indents}
-                         config-file-opts
+        opts (deep-merge config-file-opts
                          message-opts)]
     {:formatted-code (fmt/reformat-string code opts)}))
 
--- a/src/cider/nrepl/middleware/format.clj
+++ b/src/cider/nrepl/middleware/format.clj
@@ -8,9 +8,10 @@
    [clojure.tools.reader.edn :as edn]
    [clojure.tools.reader.reader-types :as readers]
    [clojure.walk :as walk]
-   [nrepl.middleware.print :as print])
+   [nrepl.middleware.print :as print]
+   [orchard.misc :refer [deep-merge]])
   (:import
-   (java.io StringWriter)))
+   (java.io File StringWriter)))
 
 ;;; Code formatting
 (defn- keyword->symbol [kw]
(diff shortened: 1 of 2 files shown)


src/cider/nrepl/middleware/format.clj
```diff
@@ -22,15 +23,23 @@
...
 (defn format-code-reply
-  [{:keys [code options]}]
-  (let [opts (some-> options
-                     (select-keys [:indents :alias-map])
-                     (update :indents generate-user-indents)
-                     (update :alias-map #(reduce-kv (fn [m k v] (assoc m (name k) v)) {} %)))]
+  [{:keys [code options ^File config-file]
+    :or {config-file default-config-file}}]
+  (let [config-file-opts (when (-> config-file .exists)
+                           (-> config-file slurp edn/read-string))
```
vemv, line 35: TODO: honor `#re` tag https://clojurians.slack.com/archives/CPABC1H61/p1654089774726049

vemv: > Looks like it's simply  [...]

Indeed! Done.

> Do we want to use the same logic as clojure-lsp and use read-string when it's a .clj file?

It's not particularly documented in https://github.com/clojure-lsp/clojure-lsp/blob/9f0c724bdc69abbb0b8aa78c8cb7f6256cf1dcbb/lib/src/clojure_lsp/feature/format.clj#L29 so I'd rather avoid that complexity / surface area for bugs.

> Also I assume the config-file key will be sent by clients as a string

No, I'd expect clients to send nothing, here we use an `:or`. The option is there merely for testability.

i.e. I'd be inclined to promote one and only one possible filename.

src/cider/nrepl/middleware/format.clj
```diff
@@ -22,15 +23,22 @@
      (assoc acc
             (keyword->symbol kw)
             (walk/postwalk #(cond-> % (string? %) keyword) rule)))
-   fmt/default-indents
```
vemv, line 25: Removed this, it was preventing regexes from working (e.g. the default `let` would take precedence over a custom `let.*`)

cljfmt doesn't need its defaults to be passed again, probably by doing so one only creates ambiguity

src/cider/nrepl/middleware/format.clj
yuhan0: > No, I'd expect clients to send nothing, here we use an :or. The option is there merely for testability.

Then perhaps the bit with this internal testable argument could be factored out into a private function and not be exposed to the public API? 

I guess it's not a big deal if Cider / Calva can be relied not to expose any defcustom for the config file and always pass nil to the format op, but it just seems unnecessary.

yuhan0:
Also just wondering if this should be designed in mind for extensibility to other formatting backends in the future, similar to the pprint middleware - eg. the more opinionated `zprint` or a hypothetical "Tonsky formatter".

Zprint for instance already recognizes its own .zprintrc files, so passing an additional key for config-file could be confusing.

bbatsov:
I agree that it'd be better to support both zprint and cljfmt as formatting backends, but I don't feel strongly about this. 

Sorry that I had completely forgotten about this PR!

vemv:
Me too :) will get back at it soon.

vemv:
Getting back into this soon...

as a note to self, besides from supporting cider-format, observing .cljfmt.edn would be interesting for dynamically generating `:style/indent` values. This would elliminate any discrepancies between clojure-mode and cider-format's results.

vemv:
I'm reworking this, will look quite different.

aiba:
Did this get abandoned?

I found this issue from https://clojurians.slack.com/archives/C099W16KZ/p1692166428844959?thread_ts=1692163238.351989&cid=C099W16KZ

I'm having the same issue -- `lsp-format-buffer` works differently from when I just press `<return>` or `<tab>` in a clojure file.  It would be really great if there was a way to have consistent formatting.

vemv:
@aiba sorry - Gmail buried this.

I actually have a newer cljfmt branch which might see the light of day around the next stable CIDER release.

PR #756, opened 2022-09-04
Fix inspect tests

r0man:
Update the inspect test data to match the new section headers and the
Datafy support added to Orchard.

- [x] All tests are passing

### DIFF
--- a/test/smoketest/project.clj
+++ b/test/smoketest/project.clj
@@ -1,6 +1,6 @@
 (defproject smoketest "0.1.0-SNAPSHOT"
   :dependencies [[nrepl "1.0.0"]
-                 [cider/cider-nrepl "0.28.6"]]
+                 [cider/cider-nrepl "0.28.5"]]
   :exclusions [org.clojure/clojure]
   :profiles {:1.7 {:dependencies [[org.clojure/clojure "1.7.0"]]}
              :1.8 {:dependencies [[org.clojure/clojure "1.8.0"]]}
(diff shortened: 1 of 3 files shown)


test/smoketest/project.clj
```diff
@@ -1,6 +1,6 @@
 (defproject smoketest "0.1.0-SNAPSHOT"
   :dependencies [[nrepl "1.0.0"]
-                 [cider/cider-nrepl "0.28.6"]]
+                 [cider/cider-nrepl "0.28.5"]]
```
r0man, line 3: I had to set this back to a version that is available on Clojars to make the tests pass. Otherwise they fail because the dependency can't be downloaded. Is this supposed to test the current version?

r0man:
Hi @bbatsov,

here's the PR that fixes the tests. Sorry again for the trouble, I
totally forgot about those tests when I said I think there are no
changes needed to cider-nrepl.

Do you think there's something to do on Cider as well regarding tests?
I did not find any tests related to the inspector there. So, I guess we are fine.

bbatsov:
Thanks for acting on this so quickly. Unfortunately I just went on a trip without a computer, so I'll leave it to @vemv to cut a new release. There's no 0.28.6 as the tests failed, so I guess he can delete and recycle the tag. I was in a hurry for my trip, so I didn't have time to test locally the tests, which clearly was a mistake.

r0man:
Alright, have a nice trip!

Issue #766, opened 2023-01-30
Complete local bindings for ClojureScript files


iarenaza:
## Expected behavior

When editing a ClojureScript file, CIDER-based completion offers local binding names (the function name and argument names if inside the function, the let-like block bindings if inside a let-like block, etc). Just like it happens when editing a Clojure file.

## Actual behavior

It doesn't offer any local binding names as completion candidates. The root of the problem is that `cider.nrepl.middleware.complete/cljs-sources` only includes `::suitable-sources/cljs-source` as a possible source. And suitable doesn't perform any local binding analysis.

But given that the local binding analysis done in `compliment.sources.local-bindings` namespace doesn't perform any evaluation or execution of the context form (thus it's independent of the actual host platform differences), we can use that same source for ClojureScript local bindings completion. I have tested it locally and it works as expected, without any side effects.

I have a pull request that I will attach in a few minutes implementing the suggested feature.

## Steps to reproduce the problem

Create a ClojureScript REPL using you favorite technology (figwheel-main, shadow-cljs, etc), and connect to it. Enable completion in CIDER following the documentation.  Make sure you don't have `lsp` completion enabled (this could mess with CIDER completion).

Once connected, open a .cljs (or .cljc) file and type the following form. Place the cursor where the '|' character is in the form below:

```clj
(let [foobar "foobar"
      foobaz "foobaz"]
  |
)
```

Start typing 'fooba'. The autocompletion feature should offer you both `foobar` and `foobaz` as completion candidates (and tell you their kind, that they are variables in this case). But it doesn't (it does in this is a .clj file, and you have an active Clojure REPL connection).

### CIDER version

1.6.0

### cider-nrepl version

0.29.0

### nREPL version

1.0.0

### Java version

openjdk version "17.0.4" 2022-07-19
OpenJDK Runtime Environment Temurin-17.0.4+8 (build 17.0.4+8)
OpenJDK 64-Bit Server VM Temurin-17.0.4+8 (build 17.0.4+8, mixed mode, sharing)

### Operating system

Debian GNU/Linux Bullseye

PR #767, opened 2023-01-30
Complete local bindings for ClojureScript files

iarenaza:
When editing a ClojureScript file, cider-nrepl doesn't offer any local binding names (the function name and argument names if inside the function, the let-like block bindings if inside a let-like block, etc.) as completion candidates, like it does when editing a Clojure file.

The root of the problem is that
`cider.nrepl.middleware.complete/cljs-sources` only includes `::suitable-sources/cljs-source` as a possible source. And `suitable` doesn't perform any local binding analysis.

Given that the local binding analysis done in
`compliment.sources.local-bindings` namespace doesn't perform any evaluation or execution of the context form (thus, it is independent of the actual host platform differences), we can use that same source for ClojureScript local bindings completion.

[Closes: #766]

- [X] The commits are consistent with our [contribution guidelines](../blob/master/.github/CONTRIBUTING.md)
- [X] All tests are passing

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## master (unreleased)
 
+### New features
+
+* [#766](https://github.com/clojure-emacs/cider-nrepl/issues/766) Complete local bindings for ClojureScript files.
+
 ## 0.29.0 (2022-12-05)
 
 * [#758](https://github.com/clojure-emacs/cider-nrepl/pull/758) Add nREPL op to parse stacktraces into data (`analyze-stacktrace`), rename `stacktrace` op to `analyze-last-stacktrace` and deprecate `stacktrace`.
--- a/src/cider/nrepl/middleware/complete.clj
+++ b/src/cider/nrepl/middleware/complete.clj
@@ -36,7 +36,13 @@
 
 (def cljs-sources
   "A list of ClojureScript completion sources for compliment."
-  [::suitable-sources/cljs-source])
+  [::suitable-sources/cljs-source
+   ;; The local binding analysis done by
+   ;; :compliment.sources.local-bindings/local-bindings doesn't perform any
+   ;; evaluation or execution of the context form. Thus, it is independent of
+   ;; the actual host platform differences. Given that, we can use that same
+   ;; source for ClojureScript completion.
+   :compliment.sources.local-bindings/local-bindings])
 
 (defn complete
   [{:keys [ns prefix symbol context extra-metadata enhanced-cljs-completion?] :as msg}]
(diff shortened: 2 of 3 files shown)


test/cljs/cider/nrepl/middleware/cljs_complete_test.clj
```diff
@@ -64,7 +64,15 @@
                                      :ns "cljs.user"
                                      :prefix "js/Ob"})
           candidates (:completions response)]
-      (is (empty? candidates)))))
+      (is (empty? candidates))))
+  (testing "local bindings"
+    (let [response (session/message {:op "complete"
+                                     :ns "cljs.user"
+                                     :prefix "ba"
+                                     :context "(defn foo [bar] (let [baz :baz] (str __prefix__)))"
```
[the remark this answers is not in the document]
iarenaza: Sure, will do!

bbatsov:
> Given that the local binding analysis done in
compliment.sources.local-bindings namespace doesn't perform any evaluation or execution of the context form (thus, it is independent of the actual host platform differences), we can use that same source for ClojureScript local bindings completion.

Good catch, I had missed this myself. It might be good to add some comment about this in the code, as at a glance it seems weird to mix ClojureScript and Clojure completion sources.

iarenaza:
> It might be good to add some comment about this in the code, as at a glance it seems weird to mix ClojureScript and Clojure completion sources.

Do you mean in the definition of `cider.nrepl.middleware.complete/cljs-sources` or in the `compliment.sources.local-bindings` namespace? I guess it's the former, but just to make sure :smile:

bbatsov:
The former, although it probably won't hurt to expand the compliment source documentation at some point.

src/cider/nrepl/middleware/complete.clj
```diff
@@ -36,7 +36,8 @@
 
 (def cljs-sources
   "A list of ClojureScript completion sources for compliment."
-  [::suitable-sources/cljs-source])
+  [::suitable-sources/cljs-source
+   :compliment.sources.local-bindings/local-bindings])
```
bbatsov, line 40: Basically when I saw this line my first reaction was "how can this possibly work?". 😆

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## master (unreleased)
 
+* [#766](https://github.com/clojure-emacs/cider-nrepl/issues/766) Complete local bindings for ClojureScript files.
```
bbatsov, line 5: This should be under "New features".

iarenaza: Done!

src/cider/nrepl/middleware/complete.clj
iarenaza: Comment added (and PR updated)

bbatsov:
Thanks!

bbatsov:
@iarenaza I've cut a new release including your change (0.30) and I've updated CIDER's `master` to use it.

iarenaza:
> @iarenaza I've cut a new release including your change (0.30) and I've updated CIDER's `master` to use it.

Great!

PR #775, opened 2023-06-26
Fix info tests

r0man:
Orchard changed how arglists-str are formatted for special forms in this PR:

https://github.com/clojure-emacs/orchard/pull/166

It looks like this broke the info tests in cider-nrepl. Assuming the new behaviour is the correct one, I fixed the tests accordingly.

- [-] You've added tests to cover your change(s)
- [x] All tests are passing
- [-] You've updated the README (if adding/changing middleware)

Thanks!


test/clj/cider/nrepl/middleware/info_test.clj
```diff
@@ -144,7 +144,7 @@
             (pr-str response))
         (is (= (:class response) "cider.nrepl.test.TestClass"))
         (is (= (:member response) "doSomething"))
-        (is (= (:arglists-str response) "[int int java.lang.String]"))
+        (is (= (:arglists-str response) "[a b c]"))
```
bbatsov, line 147: Hmm, I wonder what changed this as it doesn't seem related to the previous PR.

bbatsov:
Looking at the test failures the type info for the Java methods has not disappeared.

test/clj/cider/nrepl/middleware/info_test.clj
r0man: Well, this is a strange one. The tests do pass now in the REPL, but fail when run via `lein test`. Any ideas?

Issue #777, opened 2023-07-15
Indentation inference for defmacros


vemv:
(IDK if this had been talked about before)

Given an arbitrary macro that lacks `:style/indent` metadata, e.g. `(defmacro fact [& forms]`, I'd find it most practical that CIDER inferred a sensible `indent` value for it.

A sensible rule of thumb is:

> indent = <number of macro args before `&`>

For instance, `[a b & c]` has two arguments before `&`, so the inferred indent should be `2`.

There occasionally exist macros without `&`, e.g.:

```
(defmacro foo [opts body]
```

For those, we can devise a similar rule of thumb:

> indent = <number of macro args before an argument named `body`>

(or `forms`, `tail`, similar words)

So that foo would have an indent of `1`.

I understand that this is not quite 'pure' in approach. But my experience over the years is that very few people, even among CIDER users, will remember to add the metadata manually.

Ultimately, users can of course always add metadata, which would override any inference.

WDYT?

Cheers - V

bbatsov:
Yeah, that'd be fine. That's similar to how SLIME used to work out indentation when it wasn't specified - by relying on the the names of some params (I think `body` and `forms`). In general I'm all for adding some sensible inference to simplify the lives of users.

PR #778, opened 2023-07-17
Infer `:style/indent` metadata when suitable

vemv:
A working proposal for #777.

I saw that `style/indent` is ultimately analysed here https://github.com/clojure-emacs/cider/blob/02802a8cd5a6ac0b7c48685d892d49bf7059b966/cider-mode.el#L589-L590

That var info is in turn returned by the `track-state` middleware. So we are quite free to add extra metadata there, i.e. it will not be visible by users if they enter `(meta  #'foo)` into their repls.

LMK what you think. After that I can move most code to Orchard.

Cheers - V

### DIFF
--- a/src/cider/nrepl/middleware/format.clj
+++ b/src/cider/nrepl/middleware/format.clj
@@ -4,13 +4,60 @@
   (:require
    [cider.nrepl.middleware.util.error-handling :refer [with-safe-transport]]
    [cljfmt.core :as fmt]
+   [clojure.set :as set]
    [clojure.string :as str]
    [clojure.tools.reader.edn :as edn]
    [clojure.tools.reader.reader-types :as readers]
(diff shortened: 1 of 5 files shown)


src/cider/nrepl/middleware/format.clj
```diff
@@ -4,13 +4,60 @@
...
+(defn symbolize [x]
+  (case x
+    & '&
+    '_))
+
+(defn structure= [candidate reference]
+  (set/subset? (->> candidate (map (partial mapv symbolize)) set)
+               (->> reference (map (partial mapv symbolize)) set)))
+
+(defn compute-style-indent [^String macro-name arglists]
```
bbatsov, line 26: I'm guessing this function should have a very detailed docstring.

bbatsov:
Looks good to me overall.

PR #780, opened 2023-07-21
Only include `:diffs` for diffable values

vemv:
This will result in a more focused user experience: the diffs will only be presented when they likely have utility.

Else they can easily become noise.

Note that cider-test.el already has conditional handling for possibly-absent `:diff` values.

Cheers - V

### DIFF
--- a/src/cider/nrepl/middleware/test/extensions.clj
+++ b/src/cider/nrepl/middleware/test/extensions.clj
@@ -8,6 +8,39 @@
    [clojure.pprint :as pp]
    [clojure.test :as test :refer [assert-expr]]))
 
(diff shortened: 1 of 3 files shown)


src/cider/nrepl/middleware/test/extensions.clj
```diff
@@ -8,6 +8,30 @@
    [clojure.pprint :as pp]
    [clojure.test :as test :refer [assert-expr]]))
 
+(defn- diffable-objects? [a b]
```
bbatsov, line 11: I'd still add a docstring here, regardless this is private.

bbatsov:
Looks good overall.

PR #781, opened 2023-07-21
Fail faster on CI

vemv:
This reduces the feedback cycle from 7m to 1.5m.

Cheers - V

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -158,6 +158,9 @@ jobs:
           cache_version: << parameters.clojure_version >>|<< parameters.jdk_version >>
           steps:
             - run:
+                name: Running tests # Redundant relative to `make test`, however it makes the CI jobs fail faster, given how much time mranderson takes.
+                command: make quick-test
+            - run:
                 name: Running tests with inlined deps
                 command: make test smoketest
             - run:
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -17,6 +17,8 @@ test: clean .inline-deps test/resources/cider/nrepl/clojuredocs/export.edn
 quick-test: clean
 	lein with-profile -user,-dev,+$(CLOJURE_VERSION),+test test
 
+fast-test: quick-test
```
bbatsov, line 20: I guess that's a bit of a misnomer, as it won't be faster if the tests pass. Perhaps `fail-fast-test`?

.circleci/config.yml
```diff
@@ -157,6 +157,9 @@ jobs:
       - with_cache:
           cache_version: << parameters.clojure_version >>|<< parameters.jdk_version >>
           steps:
+            - run:
```
bbatsov, line 160: I'm a bit puzzled what does this replace. I see it added, but I don't see anything changed/removed.

Makefile
bbatsov: You named the PR "Fail faster on CI" which got me confused. 

Then my comment would be the following - occasionally there are MrAnderson specific issues, which we won't catch if we don't run the CI using it.

.circleci/config.yml
bbatsov: Ah, I finally understand what you mean. 

Might be good to add some comment here explaining all of this.

PR #783, opened 2023-07-21
`middleware.test`: offer fail-fast functionality

### LINKED ISSUE #709: `middleware.test`: offer fail-fast functionality
It would be sweet to offer fail-fast functionality. Under it:

* The first failing failing test var fails that ns; and
* The first failing ns fails the test run.

Relevant code:

https://github.com/clojure-emacs/cider-nrepl/blob/a740583c3aa8b582f3097611787a276775131d32/src/cider/nrepl/middleware/test.clj#L238-L239

https://github.com/clojure-emacs/cider-nrepl/blob/a740583c3aa8b582f3097611787a276775131d32/src/cider/nrepl/middleware/test.clj#L260-L263

Some inspiration:

```clj
(defn run-tests
  "Like clojure.test/run-tests, but fails fast."
  [& namespaces]
  (let [summary (-> (->> namespaces
                         (reduce (bound-fn [r n]
                                   (let [{:keys [fail error]
                                          :as v} (clojure.test/test-ns n)
                                         failed? (some pos? [fail error])
                                         ret (merge-with + r v)]
                                     (cond-> ret
                                       failed? reduced)))
                                 clojure.test/*initial-report-counters*))
                    (assoc :type :summary))]
    (clojure.test/do-report summary)
    summary))
```

vemv:
Fixes #709

Introduces fail-fast via a `"fail-fast"` msg param. The default is to not fail-fast, since I cannot guess the desired behavior for all cider-nrepl clients.

For cider-test.el however, I intend to `(defcustom cider-test-fail-fast t)`.

Cheers - V

### DIFF
--- a/doc/modules/ROOT/pages/nrepl-api/ops.adoc
+++ b/doc/modules/ROOT/pages/nrepl-api/ops.adoc
@@ -971,7 +971,7 @@ Optional parameters::
 
 Returns::
 * `:elapsed-time` a report of the elapsed time spent running all the given namespaces. The structure is ``:elapsed-time {:ms <integer> :humanized <string>}``.
-* `:fail-fast` If true, the tests will be considered complete after the first test has failed or errored.
+* `:fail-fast` If equals to the string "true", the tests will be considered complete after the first test has failed or errored.
 * `:ns-elapsed-time` a report of the elapsed time spent running each namespace. The structure is ``:ns-elapsed-time {<ns as keyword> {:ms <integer> :humanized <string>}}``.
 * `:results` Misc information about the test result. The structure is ``:results {<ns as keyword> {<test var as keyword> [{,,, :elapsed-time {:ms <integer> :humanized <string>}}]}}``
 * `:status` Either done or indication of an error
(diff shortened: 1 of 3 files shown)


doc/modules/ROOT/pages/nrepl-api/ops.adoc
```diff
@@ -971,6 +971,7 @@ Optional parameters::
 
 Returns::
 * `:elapsed-time` a report of the elapsed time spent running all the given namespaces. The structure is ``:elapsed-time {:ms <integer> :humanized <string>}``.
+* `:fail-fast` If true, the tests will be considered complete after the first test has failed or errored.
```
bbatsov, line 974: Just keep in mind there's no boolean type in bencode, so `true` is probably not the best way to describe this. But I get the idea. :-)

bbatsov:
Looks good to me overall.

doc/modules/ROOT/pages/nrepl-api/ops.adoc
bbatsov: Maybe a small change like `When true or the string "true"`...

bbatsov:
Potentially with other transports (e.g. EDN) one can have real booleans, but so far I think we always used the string "true". I was too lazy to check the rest of the codebase. Anyways, no particularly important right now.

PR #784, opened 2023-07-21
Refine timing for deftests with multiple `is` clauses

vemv:
The finest possible granularity for time measuring is var granularity.

However cider-test.el reports test at testing-context granularity, not var granularity.

Therefore this PR:

* Only associates the timing to testing contexts if they were the only one within that var
* Adds a new `:var-elapsed-time` map, which could be useful for future/different UIs.

Cheers - V

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -3,7 +3,10 @@ Before submitting a PR make sure the following things have been done:
 - [ ] The commits are consistent with our [contribution guidelines](../blob/master/.github/CONTRIBUTING.md)
 - [ ] You've added tests to cover your change(s)
 - [ ] All tests are passing
-- [ ] You've updated the README and/or the `cider.nrepl` ns (if adding/changing middleware)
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -3,7 +3,7 @@ Before submitting a PR make sure the following things have been done:
 - [ ] The commits are consistent with our [contribution guidelines](../blob/master/.github/CONTRIBUTING.md)
 - [ ] You've added tests to cover your change(s)
 - [ ] All tests are passing
-- [ ] You've updated the README (if adding/changing middleware)
+- [ ] You've updated the README and/or the `cider.nrepl` ns (if adding/changing middleware)
```
bbatsov, line 6: I'm guessing no one will forget to update the ns, otherwise their middleware won't work. :-) But we can mention here re-gening the docs with `lein docs`.

bbatsov: Guess so. :D

PR #787, opened 2023-07-27
Infer `:style/indent` metadata

vemv:
Implements https://github.com/clojure-emacs/cider-nrepl/issues/777

Most of the work was shipped in https://github.com/clojure-emacs/orchard/pull/169

### DIFF
--- a/src/cider/nrepl/middleware/track_state.clj
+++ b/src/cider/nrepl/middleware/track_state.clj
@@ -25,7 +25,9 @@
 
 ;;; Auxiliary
 
-(defn- inferrable-indent? [m]
+(defn- inferrable-indent?
+  "Does metadata map `m` lack, need a `:style/indent` value, and is a suitable candidate for it?"
+  [m]
   (and (:macro m)
        (:arglists m)
        (not (:style/indent m))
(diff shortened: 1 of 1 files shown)


src/cider/nrepl/middleware/track_state.clj
```diff
@@ -28,7 +29,21 @@
...
+                                         (-> m :ns ns-name)
+
+                                         (symbol? (:ns m)) ;; possible edge cases
+                                         (:ns m)
+
+                                         ;; cljs analyzer info
+                                         :else
+                                         (some-> m :name namespace symbol))
+                                       str)]
+         (not (or (string/starts-with? namespace-name "clojure.")
```
bbatsov, line 44: Why do we need special handling for the built-in macros?

src/cider/nrepl/middleware/track_state.clj
```diff
@@ -5,38 +5,68 @@
...
+   (nrepl.transport Transport)))
 
 (def clojure-core (try (find-ns 'clojure.core)
                        (catch Exception _e nil)))
 
 ;;; Auxiliary
 
-(defn- var-meta-with-fn
-  "Like clojure.core/meta but adds {:fn true} for functions and macros.
+(defn- inferrable-indent? [m]
```
bbatsov, line 28: If we end up keeping this function (I'm not sure special handling for built-in namespaces is a good idea), let's add a docstring for it.

bbatsov:
I get that clojure-mode has hardcoded indents for Clojure built-ins, but what about other clients? That's my main concern here. Probably that's something that should be handled in CIDER, not here.

bbatsov:
But some follow the standard naming for params (e.g. `body`, `forms`)? 

> Which is worth emphasizing, does not change anything when compared to traditional cider-nrepl behavior (which always lacked indent inference)

I get this. I'm just trying to understand what makes those macros special. :-)

bbatsov:
I don't want to open extra work for you. I guess for the needs of CIDER this makes sense and we can think down the road if we want to expand the inference. I can live with the PR in its current form.

Issue #788, opened 2023-07-27
Metadata inference for 'proxy' vars


vemv:
The following pattern can be sometimes found in Clojure:

```clj
(def json-request-format json/json-request-format)
```

Where

* the `def` lacks a docstring,
* the def value is a single symbol,
  * therefore its value is known at compile-time
* said symbol can be resolved to a var,
* and that var has a docstring.

...For those cases, the `track-state` middleware could add :doc metadata for CIDER usage.

Note that this is unobstrusive (non-mutative): users querying `(-> #'x meta :doc)` would not see our infered value.

We already are doing something similar in https://github.com/clojure-emacs/cider-nrepl/pull/787 (for :style/indent)

So, this pattern can be used for:

* :doc
* :arglists
* :tag
* :style/indent (preempting any inference)
* possibly a few, safe, others.

This can be done for clj and cljs alike.

PR #790, opened 2023-07-28
Metadata inference for 'proxy' vars

vemv:
Draft for https://github.com/clojure-emacs/cider-nrepl/issues/788. It seems in a good shape to me. I'd say its use case can be understood very easily given its deftest.

What's pending:

* cljs support (easy)
* I'll want to QA this locally
* would certainly appreciate an early review!

Cheers - V

### DIFF
--- a/src/cider/nrepl/middleware/track_state.clj
+++ b/src/cider/nrepl/middleware/track_state.clj
@@ -5,9 +5,11 @@
    [cider.nrepl.middleware.util :as util]
    [cider.nrepl.middleware.util.cljs :as cljs]
    [cider.nrepl.middleware.util.meta :as um]
+   [cider.nrepl.utils.reader :as utils.reader]
    [clojure.java.io :as io]
    [clojure.string :as string]
    [clojure.tools.namespace.find :as ns-find]
+   [com.climate.claypoole :as claypoole]
    [nrepl.misc :refer [response-for]]
    [nrepl.transport :as transport]
    [orchard.cljs.analysis :as cljs-ana]
@@ -15,7 +17,7 @@
    [orchard.java.classpath :as cp]
    [orchard.misc :as misc])
   (:import
-   (clojure.lang MultiFn Namespace)
+   (clojure.lang MultiFn Namespace Var)
    (java.net SocketException)
    (java.util.jar JarFile)
    (nrepl.transport Transport)))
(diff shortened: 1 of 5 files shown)


src/cider/nrepl/middleware/track_state.clj
```diff
@@ -281,21 +345,32 @@
...
-                                         (if (jar-ns-fn name)
-                                           acc ;; Remove all jar namespaces.
-                                           (assoc! acc name (ns-as-map ns all-namespaces)))))
-                                     all-namespaces)
+         ns-as-map-fn (memoize ns-as-map)
+         project-ns-map (claypoole/with-shutdown! [pool (claypoole/threadpool (if (System/getenv "CI")
+                                                                                1
+                                                                                4))]
+                          (->> all-namespaces
+                               (claypoole/pmap pool
```
vemv, line 353: The parallelism is necessary because we're reading files, for reading forms.

4 seems a good value (it's an IO bound problem, but we don't want to create a lot of threads).

src/cider/nrepl/middleware/track_state.clj
```diff
@@ -281,21 +345,32 @@
          all-namespaces (if cljs
                           (vals (cljs-ana/all-ns cljs))
                           (all-ns))
-         project-ns-map (fast-reduce (fn [acc ns]
-                                       (let [name (ns-name-fn ns)]
-                                         (if (jar-ns-fn name)
-                                           acc ;; Remove all jar namespaces.
-                                           (assoc! acc name (ns-as-map ns all-namespaces)))))
-                                     all-namespaces)
+         ns-as-map-fn (memoize ns-as-map)
```
vemv, line 348: By memoizing ns-as-map which ultimately calls the new, IO-costly stuff, we reuse that IO work a few times.

The memoization is not global - its lifespan is that of the nrepl request.

src/cider/nrepl/utils/reader.clj
```diff
@@ -0,0 +1,31 @@
+(ns cider.nrepl.utils.reader
+  (:require
+   [clojure.java.io :as io]
+   [clojure.tools.reader :as reader]
+   [clojure.tools.reader.reader-types :as reader-types])
+  (:import
+   (clojure.lang Namespace)))
+
+(defn read-form-at
```
vemv, line 9: I could add a dedicated test...

src/cider/nrepl/middleware/track_state.clj
```diff
@@ -281,21 +345,32 @@
...
+         project-ns-map (claypoole/with-shutdown! [pool (claypoole/threadpool (if (System/getenv "CI")
```
vemv, line 349: I had to do this, else tests in CI would time out.

I strongly believe it only boils down to Circle servers being slow - just 1 vCPU - adding threads makes it worse.

As a consequence, builds will be slower.

It shouldn't be a practical problem for users, given we'll use 4 threads for them.

There's also great care in avoiding these IO reads for macros that do not need.

vemv: CI slowness is no more after my second commit.

vemv:
I tried this out, and while the core impl works well, it should really live in middleware.info, not track-state.

PR #791, opened 2023-07-30
`track-state`: Include `cljs.core` ns info

vemv:
Part of https://github.com/clojure-emacs/cider/issues/3385

Cheers - V

### DIFF
--- a/src/cider/nrepl/middleware/track_state.clj
+++ b/src/cider/nrepl/middleware/track_state.clj
@@ -219,7 +219,7 @@
   (cond
     (and cljs (not (contains? old-ns-map 'cljs.core)))
     (assoc project-ns-map 'cljs.core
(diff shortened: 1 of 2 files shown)


src/cider/nrepl/middleware/track_state.clj
```diff
@@ -219,7 +219,7 @@
   (cond
     (and cljs (not (contains? old-ns-map 'cljs.core)))
     (assoc project-ns-map 'cljs.core
-           (ns-as-map (cljs-ana/find-ns cljs "cljs.core")
+           (ns-as-map (cljs-ana/find-ns cljs 'cljs.core)
```
vemv, line 222: All keys in this map are symbols (I verified so locally for clj and cljs).

bbatsov (commented):
Looks good to me. I wonder why no one had noticed this.

Issue #796, opened 2023-08-04
xref: return absolute paths when possible


vemv:
Although https://github.com/clojure-emacs/cider/pull/3349 is a good idea (I'm building on top of it), its technique of translating relative, unprefixed paths like `"orchard/cljs/meta.cljc"` sadly is not workable, because it only _sometimes_ works.

`cider-find-file`'s docstring says:

> Return a buffer visiting the file URL if it exists, or nil otherwise.
If URL has a scheme prefix, it must represent a fully-qualified file path
or an entry within a zip/jar archive.  If AVFS (archive virtual file
system; see online docs) is mounted the archive entry is opened inside the
AVFS directory, otherwise the entry is archived into a temporary read-only
buffer.  If URL doesn't contain a scheme prefix and is an absolute path, it
is treated as such.  **Finally, if URL is relative, it is expanded within each
of the open Clojure buffers till an existing file ending with URL has been
found**.

Because it only translates relative paths for _open_ buffers, it will often return nil.

Therefore, whenever possible, `xref` middleware ops have to return `file:<absolute-path>`.

`cider-find-file`'s translation still seems desirable for some cases, mainly TRAMP and Docker.

PR #797, opened 2023-08-05
`fn-refs` and `fn-deps` middleware: offer `file-url` field

vemv:
Fixes https://github.com/clojure-emacs/cider-nrepl/issues/796

Cheers - V

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 * [#796](https://github.com/clojure-emacs/cider-nrepl/issues/796): `fn-refs` and `fn-deps` middleware: add new `:file-url` field.
   * these are absolute and prefixed by `file:` or `file:jar:`, whenever possible.
 * `fn-refs` and `fn-deps` middleware: sort the results by file, line and column.
-* Bump `orchard` to [1.4.1](https://github.com/clojure-emacs/orchard/blob/v0.14.1/CHANGELOG.md#0141-2023-08-05).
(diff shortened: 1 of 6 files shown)


bbatsov:
I think cider is the only client of this functionality, so you don't really need the `file-format` option IMO and can simplify things a bit.

vemv:
I agree and was on the fence about this one.

But I think it's a good thing to normalize "retro-compat" techniques - if we don't have the habit of using them, we'll forget about them when it matters?

Also, not impossible to have an odd cider/cider-nrepl combo. And it's good to leave the door open to alternative emacs-side xref impls.

e.g. I'm working on a https://github.com/clojure-emacs/cider/pull/3349 spin-off, but what if we decide to change the approach again a few months from now?

bbatsov:
> But I think it's a good thing to normalize "retro-compat" techniques - if we don't have the habit of using them, we'll forget about them when it matters?

That's my bigger concern - I've never been fond of API versioning, so this `v2` bothers me a lot. (I'm fan of accretive changes, meaning I'd rather add new fields in the response than make it possible for the client to pick between versions) For me it's much simpler here to change this than to spend a lot of time discussing what's the best way to do `retro-compat`.

vemv:
Alright! Adding a new field is a good choice as well.

vemv:
Ready.

vemv:
Added an extra commit to sort the results, especially for the line. Otherwise I was finding cider.el side that that for the same given file, results could be presented in an erratic order: line 50, 20, 80, etc.

bbatsov (approved):

bbatsov:
Looks good me.

vemv:
Thanks!

Issue #710, opened 2021-09-14
Namespace the ops


bbatsov:
Right now `cider-nrepl` doesn't namespace its ops, which in the past became problematic when we decided to add a completion op to nREPL. Normally it would have been named `complete`, but we had to go with a different name, as `cider-nrepl` already had `complete`.

It'd be nice to give all the ops names like `cider/info`, `cider/complete`, etc. This has to happen in several steps, though:

* first we need to add the new names alongside the old names
* update the docs to promote the newer ops (and maybe send some notifications to clients when using the legacy names)
* give the clients a few years to update
* drop the old "global" names (optional step)

In general new ops shouldn't use the global namespace.  The same applies to `refactor-nrepl`.

bbatsov:
Yeah, I think that makes sense, as we can have a collision in the name of responses as well. As requests and responses are normally matched on their IDs that's not a big problem for most clients, but the IDs are optional so who knows what someone might be using. 

I definitely regret not thinking more about the proper namespacing of ops earlier in the dev process. :D

r0man:
Ok, I will prefix the responses as well then.

PR #799, opened 2023-08-07
Make the `track-state` middleware invokeable directly

vemv:
> Related: https://github.com/clojure-emacs/cider/issues/3393

There's a good chance that this is necessary, given https://github.com/thheller/shadow-cljs/issues/1138

Even if not, I found an independently good use case for it:

* Assume I `nano src/conduit/events.cljs` and hit save
  * CIDER over Emacs has no way of knowing of this save - no nrepl message was triggered for our connections
* Then, in the CIDER repl output, I will see `[:app] Build completed. (263 files, 2 compiled, 0 warnings, 0.18s)` over stderr
  * So, I can inspect stderr within a `nrepl-make-response-handler` call, grep for such a string, and conditionally `nrepl-send-request '("op" "cider/state") ,,,` from there.

I have fully tested out this solution, it works well for its two foreseen use cases (working around the shadow-cljs issue, and responding to external events).

Cheers - V


bbatsov:
I think the change about the namespace should be done separately.

CHANGELOG.md
```diff
@@ -2,6 +2,13 @@
 
 ## master (unreleased)
 
+### New features
+
+* Make the `track-state` middleware invokeable directly, by requesting the new `"cider/state"` op.
```
bbatsov, line 7: I'd name this `get-state` or even better - `get-ns-state`, so it's clearer what it does. It wasn't that important when it couldn't be triggered, but now it should be some verb in line with most ops.

doc/modules/ROOT/pages/nrepl-api/ops.adoc
```diff
@@ -1302,7 +1288,7 @@ Returns::
 
 
 
-=== `log-add-appender`
+=== `cider/log-add-appender`
```
bbatsov, line 1291: I'm guessing you were doing some experiments with this.

CHANGELOG.md
vemv: I chose it for consistency with this key https://github.com/clojure-emacs/cider/blob/eedbab26684cb1ce01a940b5b0da9ff78aeb4eac/cider-repl.el#L238 , however I'd agree that the alternatives are clearer

doc/modules/ROOT/pages/nrepl-api/ops.adoc
vemv: `cider.nrepl.middleware.log` uses ns-qualified ops, so the doc benefits from this tweak

vemv (commented):
> I think the change about the namespace should be done separately.

The doc one?

bbatsov:
> I think the change about the namespace should be done separately.

My bad - I didn't read the diff carefully and I thought you were also adding a solution for #710 when I saw the update to the docs. :-)

CHANGELOG.md
bbatsov: Yeah, I get this, but I don't think such a name would be appropriate for something one can invoke. The response will obviously stay the same.

vemv:
Thanks!

PR #808, opened 2023-08-27
`"ns-path"` op: also include a `:url` attribute in the response

vemv:
Previously, `ns-path` could return responses that were often of no use in clojurescript.

For instance, `"conduit/events.cljs"` cannot be open (lacks a src/test component) except if a matching Emacs buffer happened to be open already.

So this PR adds a `:url` attribute, as documented.

Works:

```
(-->
  id         "35"
  op         "ns-path"
  session    "3c26ce70-2c20-4b3a-9d14-017c7497f663"
  time-stamp "2023-08-27 15:57:14.251984000"
  ns         "conduit.events"
)
(<--
  id         "35"
  session    "3c26ce70-2c20-4b3a-9d14-017c7497f663"
  time-stamp "2023-08-27 15:57:14.253630000"
  path       "conduit/events.cljs"
  status     ("done")
  url        "file:/Users/vemv/conduit/src/conduit/events.cljs"
)
```

### DIFF
--- a/src/cider/nrepl/middleware/ns.clj
+++ b/src/cider/nrepl/middleware/ns.clj
@@ -7,6 +7,7 @@
    [cider.nrepl.middleware.util.error-handling :refer [with-safe-transport]]
    [cider.nrepl.middleware.util.meta :as um]
    [orchard.cljs.analysis :as cljs-analysis]
+   [clojure.java.io :as io]
    [orchard.info :as info]
    [orchard.misc :as misc]
    [orchard.namespace :as ns]
@@ -100,7 +101,13 @@
   {:ns-vars-with-meta (ns-vars-with-meta msg)})
 
 (defn- ns-path-reply [msg]
-  {:path (ns-path msg)})
+  (let [v (ns-path msg)
+        cljs? (cljs/grab-cljs-env msg)]
+    {:path v
+     :url (if cljs?
+            (or (some-> v io/resource str)
+                v)
+            v)}))
 
 (defn- ns-load-all-reply
   [_msg]
--- a/test/cljs/cider/nrepl/middleware/cljs_ns_test.clj
+++ b/test/cljs/cider/nrepl/middleware/cljs_ns_test.clj
@@ -47,9 +47,11 @@
              {:arglists "(quote ([e]))"}))))
 
   (testing "ns-path op"
-    (let [{:keys [^String path]} (session/message {:op "ns-path"
-                                                   :ns "cljs.core"})]
-      (is (.endsWith path "cljs/core.cljs")))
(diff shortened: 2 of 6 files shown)


src/cider/nrepl/middleware/ns.clj
```diff
@@ -100,7 +101,13 @@
   {:ns-vars-with-meta (ns-vars-with-meta msg)})
 
 (defn- ns-path-reply [msg]
-  {:path (ns-path msg)})
+  (let [v (ns-path msg)
+        cljs? (cljs/grab-cljs-env msg)]
+    {:path v
+     :url (if cljs?
```
bbatsov, line 107: Shouldn't we return a URL for Clojure as well?

test/cljs/cider/nrepl/middleware/cljs_ns_test.clj
```diff
@@ -47,9 +47,11 @@
              {:arglists "(quote ([e]))"}))))
 
   (testing "ns-path op"
-    (let [{:keys [^String path]} (session/message {:op "ns-path"
-                                                   :ns "cljs.core"})]
-      (is (.endsWith path "cljs/core.cljs")))
+    (let [{:keys [^String path ^String url]} (session/message {:op "ns-path"
```
bbatsov, line 50: Probably you want to update the Clojure tests for this op as well. Seems they are in a different file.

src/cider/nrepl/middleware/ns.clj
vemv: For Clojure they already are URLs

test/cljs/cider/nrepl/middleware/cljs_ns_test.clj
vemv: See `test/clj/cider/nrepl/middleware/ns_test.clj` in PR diff

src/cider/nrepl/middleware/ns.clj
bbatsov: Hmm, so why not just put the URL for ClojureScript into the `file` key then? For me it just looks weird when the return structure differ based on the runtime. Alternatively you can stick the Clojure `path` under `url` as well, so at least it'd be consistent.

vemv: The return structure is the same for both runtimes: `:path`, `:url`.

`:path` remains unchanged.

`:url` equals `:path` on the JVM, because `:path` always has been a URL for JVM Clojure.

`:url` is different from `:path` in Clojurescript. Ideally, `:path` would have been a URL to begin with.

This is the best backwards-compatible change I could think of.

bbatsov: Ah, I didn't read this carefully.

PR #814, opened 2023-09-20
`info` and `eldoc` ops: also return `:doc-fragments` when available

vemv:
Commits:

* Bump Orchard
* Modernize clj-kondo setup
* Use mranderson more cleanly
* Add enrich-classpath to the Makefile
* README: document Make tasks
* Adapt `format-test`
* `info` and `eldoc` ops: also return `:doc-fragments` when available

The last commit, of course, is the most important one. Fortunately it's very thin, as it simply forwards attributes from Orchard.

Cheers - V

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -216,10 +216,11 @@ workflows:
                 name: Running cljfmt
                 command: |
                   make cljfmt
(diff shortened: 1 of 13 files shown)


bbatsov (approved):
Looks good to me overall.

vemv:
Thanks!

Issue #812, opened 2023-09-16
Improve `eldoc-reply` with Compliment's type inference


vemv:
Currently, eldoc requests go like this:

```lisp
(-->
  id         "376"
  op         "eldoc"
  session    "6ce622ae-b747-47e0-a302-efb8b43c3142"
  time-stamp "2023-09-16 16:57:49.180643000"
  ns         "my.ns"
  sym        ".root"
)
(<--
  id         "376"
  session    "6ce622ae-b747-47e0-a302-efb8b43c3142"
  time-stamp "2023-09-16 16:57:49.216261000"
  class      ("java.lang.reflect.Field" "java.lang.reflect.Method")
  docstring  nil
  eldoc      (("this"))
  member     "root"
  status     ("done")
  type       "function"
)
```

...That is, it requests for the doc of `.root` without any other context. Which means, we can easily return more info than we need, which is an ambiguous and verbose experience for users.

[Compliment](https://github.com/alexander-yakushev/compliment) excels at deriving a specific type out a provided context.

So, clients could pass an optional context, just like we do for the `"complete"` op:

```diff
+  context                   "(let [^Thread t nil] (__prefix__ t))"
```

Then, cider-nrepl, if it detected that a `context` was a specified, could query Compliment for the relevant type, before hitting Orchard's `info`.

PR #815, opened 2023-09-20
`info` and `eldoc` ops: accept a Compliment-style `context` parameter

vemv:
Fixes https://github.com/clojure-emacs/cider-nrepl/issues/812

Tested out locally - works as intended!

Cheers - V

### DIFF
--- a/src/cider/nrepl/middleware/info.clj
+++ b/src/cider/nrepl/middleware/info.clj
@@ -59,7 +59,10 @@
           blacklist
           util/transform-value))))
 
-(defn extract-class-from-compliment [ns-str context]
+(defn- extract-class-from-compliment
+  "Given a Compliment-style `context`, returns the inferred class name
+  of the object placed at __prefix__."
+  [ns-str context]
   (when (and (seq ns-str)
              (seq context))
     (try
--- a/src/cider/nrepl/middleware/info.clj
+++ b/src/cider/nrepl/middleware/info.clj
@@ -1,5 +1,7 @@
 (ns cider.nrepl.middleware.info
   (:require
+   [compliment.context]
+   [compliment.sources.class-members]
    [cider.nrepl.middleware.util :as util]
    [cider.nrepl.middleware.util.cljs :as cljs]
    [cider.nrepl.middleware.util.error-handling :refer [with-safe-transport]]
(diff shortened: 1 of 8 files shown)


src/cider/nrepl/middleware/info.clj
```diff
@@ -68,10 +105,9 @@
                              {:env env
                               :dialect :cljs}))]
     (cond
+      java? (info/info-java class (or member sym))
       (and ns sym) (info/info* info-params)
-      (and class member) (info/info-java class member)
-      :else (throw (Exception.
```
vemv, line 73: The `throw` wasn't compatible with the new logic. nil is more flexible and will result in a graceful op-specific reply (e.g. status `no-info` or `no-eldoc`)

vemv:
@alexander-yakushev - feel free to check out this PR. You may be happy to see Compliment used in a new way!

Maybe later as things get more time-proven, we may want to have an explicit API just for this kind of purpose.

src/cider/nrepl/middleware/info.clj
```diff
@@ -57,9 +59,44 @@
           blacklist
           util/transform-value))))
 
+(defn extract-class-from-compliment [ns-str context]
```
bbatsov, line 62: I docstring will be nice here, especially given this is a public function. (I know it's clear what it does, but still)

src/cider/nrepl/middleware/info.clj
```diff
@@ -57,9 +59,44 @@
...
+        ;; We can always be analyzing a broken context.
+        nil))))
+
 (defn info
-  [{:keys [ns sym symbol class member] :as msg}]
-  (let [[ns sym class member] (map misc/as-sym [ns (or sym symbol) class member])
+  [{:keys [ns sym class member context]
+    legacy-sym :symbol
+    :as msg}]
+  (let [sym (or (not-empty legacy-sym)
```
bbatsov, line 78: Why do you need `not-empty` here?

src/cider/nrepl/middleware/info.clj
```diff
@@ -57,9 +59,44 @@
...
+                (not-empty sym))
+        class (try
+                (or (when (and (seq class)
```
bbatsov, line 81: Is this nil-punning here? I still think it doesn't read very well. :D

bbatsov:
Other than the slightly confusing usage of `not-empty` and `seq` here and there it seems OK overall.

src/cider/nrepl/middleware/info.clj
vemv: For discarding empty strings

I don't necessarily expect them, but I've found it's a good habit to handle them.

src/cider/nrepl/middleware/info.clj
vemv: My understanding is that you'd prefer...?

```clj
(when (and (seq class) (seq ns))
  (find-ns (symbol ns)))
```

can do!

vemv: Not sure of what the alternative would be

(I had written a suggestion, but it wasn't accurate)

bbatsov: I'm fine either way - I just don't like much mixing the use of `seq` and `not-empty`. Rich liked `seq`, but `not-empty` was other later on for those who felt `seq` wasn't very intention revealing.

bbatsov: (my main point was the one I made earlier - that probably checking for empty strings is redundant, although I understand your reasoning as well)

src/cider/nrepl/middleware/info.clj
vemv: 🤝

In an ideal world we'd have something like Spec checking at the edges

src/cider/nrepl/middleware/info.clj
vemv: A nice convention is to exclusively use `seq` as a condition, and `not-empty` as an object one intends to further use

PR #820, opened 2023-10-02
Exercise enrich-classpath and all Orchard Parser variants in CI

vemv:
* Exercise enrich-classpath in CI
* Exercise each of the three Orchard Java parsers
  * This is accomplished by either of
    * excluding enrich-classpath;
    * including it partially (removing the `--add-opens` flags it adds);
    * or including it fully.

It is valuable to check that our usual tests don't fail in presence of Enrich. Just a couple tests had to be adapted (for the better: with enrich in, the assertion is richer).

It's basically the same setup I introduced in https://github.com/clojure-emacs/orchard/pull/192 , with some variations.

Cheers - V

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -148,6 +148,9 @@ jobs:
       clojure_version:
         description: Version of Clojure to test against
         type: string
+      parser_target:
+        description: The Orchard Java parser to be exercised
+        type: string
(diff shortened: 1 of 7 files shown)


bbatsov:
Looks good to me.

bbatsov (approved):

PR #822, opened 2023-10-02
Proxy clojure.test/report messages to parent reporter

djblue:

- [ ] All tests are passing

Thanks!

### DIFF
--- a/src/cider/nrepl/middleware/test.clj
+++ b/src/cider/nrepl/middleware/test.clj
@@ -353,17 +353,18 @@
    (test-ns ns vars false))
 
   ([ns vars fail-fast?]
-   (binding [test/report report]
-     (test/do-report {:type :begin-test-ns, :ns ns})
-     (let [time-info (atom nil)]
-       (timing time-info
-         (if-let [test-hook (ns-resolve ns 'test-ns-hook)]
-           (test-hook)
-           (test-vars ns vars fail-fast?)))
-       (test/do-report {:type :end-test-ns
-                        :ns ns
-                        :ns-elapsed-time @time-info})
-       @current-report))))
+   (let [parent test/report]
+     (binding [test/report (fn [msg] (parent msg) (report msg))]
+       (test/do-report {:type :begin-test-ns, :ns ns})
+       (let [time-info (atom nil)]
+         (timing time-info
+           (if-let [test-hook (ns-resolve ns 'test-ns-hook)]
+             (test-hook)
+             (test-vars ns vars fail-fast?)))
+         (test/do-report {:type :end-test-ns
+                          :ns ns
+                          :ns-elapsed-time @time-info})
+         @current-report)))))
 
 (defn test-var-query
   "Call `test-ns` for each var found via var-query."


vemv:
The changes look reasonable, however could you expand on the problem being solved?

Also, have you tried them locally? It might be as simple as evaluating things on the fly, otherwise please check out the README for a recipe.

Finally, we'd appreciate some test coverage.

As a note to self, I'll want to test this locally and check that the change causes no unexpected behavior.

Thanks - V

seancorfield:
The "problem" here is that CIDER's test report handling doesn't compose with any other middleware in the stack that wants to handle test reports because CIDER just "eats" the test reporting and no other middleware gets a bite.

In particular, this problem means that Portal middleware cannot do test reporting when used with CIDER middleware.

This PR has come about because of the discussion you started on Slack in #calva as we are trying to get the CIDER and Portal middlewares to "play nice" together.

vemv:
I spent some time trying to debug this, no luck.

I'd still need a greater level of detail as for what are we seeking to accomplish. Please describe it from the user point of view.

I'm saying this because other approaches seem possible, e.g.

* instead of hitting the original `report`, modify our defmethods (defined in the same ns) to so something additional on demand, like `tap>`ping messages perhaps?
* Capture the `test-*` ops in this fashion https://github.com/djblue/portal/blob/b939f751aedc6bdc9cd83f2d9771e33048cef052/src/portal/nrepl.clj#L100
  * nrepl middleware can not only replace an op handler, but also invoke it while adding some arbitrary aspect.

Both bullet points can be used in conjuction, e.g. the Portal middleware can tell our ns to perform the tapping, or whatever else is needed.

In the meantime I'll close this PR because each commit triggers quite the CI matrix. Please develop locally with `lein test :only cider.nrepl.middleware.test-test` (or Makefile tasks, per readme) until it's passing, and feel free to continue the conversation over here!

djblue:
@vemv Thanks for looking into it.

I think the main thing I was trying to accomplish is enabling both middleware to capture and process test output independently without having users configure them to work together. I was trying to use the same pattern in the [portal.nrepl](https://github.com/djblue/portal/blob/master/src/portal/nrepl.clj#L96-L97) middleware. I feel like this approach is still worth a few more tries.

The main thing I discovered in my debugging is that calling any form of `print` in middleware will cause a test failure, it just so happens that the `clojure.test/report` default implementation will print its results. This wasn't an issues I saw outside the test environment.

Are there rules around where middleware can call print?

Also, sorry I wasn't aware each commit would cause so much work for ci 😱 Thanks for the lein tip.

vemv:
> I think the main thing I was trying to accomplish is enabling both middleware to capture and process test output independently without having users configure them to work together. I was trying to use the same pattern in the [portal.nrepl](https://github.com/djblue/portal/blob/master/src/portal/nrepl.clj#L96-L97) middleware. I feel like this approach is still worth a few more tries.

One possible drawback is that existing users (e.g. emacs + cider-nrepl) would see test output twice? e.g. once through the CIDER UI, and then again over stdout.

That, would be of course a non-starter.

The LOC you linked to appears to `swap!` some values. Wouldn't that be quite similar to what I mentioned? (`tap>`ping values)

> The main thing I discovered in my debugging is that calling any form of print in middleware will cause a test failure,

I can check it out, but I doubt that is the case. Perhaps we're facing a pretty extreme edge case, since we're hitting clojure.test while we are running the test suite (`cider.nrepl.middleware.test-test`) which also uses clojure.test. e.g. there's a _double_ usage of clojure.test: one as part of the SUT, and one as the tooling.

> Also, sorry I wasn't aware each commit would cause so much work for ci 😱 Thanks for the lein tip.

It's no issue :) of course, it's expected that we'll CI-build incoming PRs, but I was fearing a wave of wip commits.

djblue:
@vemv after playing with things for a bit, I think the easiest thing would be for the Portal middleware to treat `cider.nrepl.middleware.test/report` the same as `clojure.test/report`. The only thing currently missing is [`:dynamic true`](https://github.com/djblue/cider-nrepl/commit/fcf5cce08123aab382a05aa17018ae14c6958edf). Portal would intercept messages, accumulate them, then call the existing implementation.

How does that sound to you?

vemv:
It sounds good to me.

Would you be able to play with it for about 1-2 weeks to make sure that it's the right approach / all that is needed?

That way, we can confidently enable this "API" (we wouldn't want to retract it later as it would be a breaking change).

Local installation is described here https://github.com/djblue/cider-nrepl/blob/fcf5cce08123aab382a05aa17018ae14c6958edf/README.md?plain=1#L83-L84

As a suggestion, you may want to check Portal-side that the var is dynamic. It's not rare for users to run an older cider-nrepl.

Cheers - V

Issue #770, opened 2023-03-06
new middleware op to tap-current-value


behrica:
In order to continue to support:
https://github.com/clojure-emacs/cider/issues/3311

I think we need a new middleware function, which could the be called later from the cider inspect elisp code to be written.

I think it should be very similar to this function
https://github.com/clojure-emacs/cider-nrepl/blob/d219ce610d9030108c273fab933538fb3b24d8be/src/cider/nrepl/middleware/inspect.clj#L105

and call the new function in orchard `inspect/tap-current-value`
introduced with commit: https://github.com/clojure-emacs/orchard/commit/625a78a4e2d72edf090543d88d1fbd00e4af7f14

Does this seems reasonable to go ahead in supporting https://github.com/clojure-emacs/cider/issues/3311 ?

behrica:
Would you be interested in a PR for this ?

behrica:
I tried to figure out, how I can test my new middleware op without corresponding elisp functions.

I managed to do this:

```
(with-open [conn (nrepl/connect :port 42857)]
     (-> (nrepl/client conn 1000)    ; message receive timeout required
         (nrepl/message {:op "eval" :code "(+ 2 3)"})
         nrepl/response-values))
```

How can I use the **inspector** middleware in this way ?
Are there any prerequisites (middleware ops to be called to "init" the inspector ) ?
Something like this (with an existing op, does not work):




```
(with-open [conn (nrepl/connect :port 42857)]
     (-> (nrepl/client conn 1000)    ; message receive timeout required
         (nrepl/message {:op "inspect-def-current-value" :ns "user" :var-name "test-xxx"})
         ))

```

I suppose I need to call som middleware ops before this ?
I looked at the documentation of all midleware ops:
https://docs.cider.mx/cider-nrepl/nrepl-api/ops.html
        
but did not see anything which looked like "starting the inspector"

behrica:
Otherwise asked:

Which middleware op(s) do I need to call, so that a following op
`inspect-def-current-value` would have a notion of "current value" ?

behrica:
Knowing this, I could then test my op: `inspect-tap-current-value`

behrica:
I did as you suggested and did both small changes to `cider-nrepl` and `cider` at the same time localy and verified that it works.
I will do a PR first here.

PR #824, opened 2023-10-15
Add new op `inspect-tap-current-value`

behrica:
adresses #770

### DIFF
--- a/test/clj/cider/nrepl/middleware/inspect_test.clj
+++ b/test/clj/cider/nrepl/middleware/inspect_test.clj
@@ -117,8 +117,11 @@
 
 (defn inspect-tap-current-value-test-fixture [f]
   (add-tap set-inspect-tap-current-value-test-atom-fn)
+  (Thread/sleep 5000)
   (f)
+  (Thread/sleep 5000)
   (reset! inspect-tap-current-value-test-atom nil)
+  (Thread/sleep 5000)
   (remove-tap set-inspect-tap-current-value-test-atom-fn))
 
 (use-fixtures :each session/session-fixture inspect-tap-current-value-test-fixture)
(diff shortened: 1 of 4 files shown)


test/clj/cider/nrepl/middleware/inspect_test.clj
```diff
@@ -505,3 +518,17 @@
...
+  (testing "inspect-tap-current-value taps the current inspector value"
+    (do
+      (session/message {:op   "eval"
+                        :code "(def x (+ 3 4)))"})
+      (session/message {:op "eval"
+                        :inspect "true"
+                        :code    "x"})
+      (session/message {:op  "inspect-push"
+                        :idx 1})
+      (session/message {:op  "inspect-tap-current-value"}))
```
vemv, line 532: I'd recommend to wait untiil `(= "7"  @inspect-tap-current-value-test-atom)`, up to 10 seconds, then perform the `is`.

What is probably happening is that the clojure taps aren't instantaneous since they have some threading stuff behind the scenes.

(In Orchard we also have some test flakiness related to the inspector/tap functionality. I never got to debug it)

...There's the possibility that we'll still have flakiness after the improvement, but not as pronounced as right now.

behrica: that fixed it

test/clj/cider/nrepl/middleware/inspect_test.clj
```diff
@@ -505,3 +515,17 @@
...
+    (Thread/sleep 10000)
```
vemv, line 530: Please sleep in chunks of 50ms, up 1o 10000

behrica: Not sure what you mean. Just wait using a loop ?

Or testing in a loop, with 50 ms waiting in between and fail if limit reached ?

vemv:
I added some feedback directly in https://github.com/clojure-emacs/cider-nrepl/commit/bdc065d555365623c202e19a6f4e20c50631654e preserving attribution.

I'll add the op to cider.el soon.

Thanks for your contributions!

Issue #828, opened 2023-11-01
Haystack slow on first hits


vemv:
I don't know exactly why, but the first time Haystack is used to analyze exceptions, it takes quite a few seconds. About 10 or so.

Later it becomes instantaneous.

It's long enough for one to believe it's not working.

Let's see if there's some low-hanging fruit. If not, probably performing related `require`s eaglerly would be a good compromise.

vemv:
```
cider.nrepl.middleware.stacktrace> (try
                                     (/ 3 0)
                                     (catch Exception e
                                       (time (analyzer/analyze e))))
"Elapsed time: 12399.759643 msecs"
```

This happens even after requiring`cider.nrepl.middleware.stacktrace` / Haystack in advance.

Afterwards it's 30ms

PR #829, opened 2023-11-01
Warmup Orchard caches for exceptions in advance

vemv:
Fixes https://github.com/clojure-emacs/cider-nrepl/issues/828

I verified it works as intended, shaving 12s for the first time `*cider-error*` pops up.

### DIFF
--- a/src/cider/nrepl.clj
+++ b/src/cider/nrepl.clj
@@ -11,6 +11,7 @@
    [cider.nrepl.middleware :as mw]
    [cider.nrepl.middleware.util.cljs :as cljs]
    [cider.nrepl.print-method] ;; we load this namespace, so it's directly available to clients
+   [haystack.analyzer :as analyzer]
    [nrepl.middleware :refer [set-descriptor!]]
    [nrepl.middleware.caught :refer [wrap-caught]]
    [nrepl.middleware.print :refer [wrap-print wrap-print-optional-arguments]]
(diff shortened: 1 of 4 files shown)


src/cider/nrepl.clj
```diff
@@ -19,15 +20,28 @@
...
-;; (This was our behavior for many years, then had to be disabled for test suite reasons in Orchard 0.15.0 to 0.17.0 / cider-nrepl 0.38.0 to 0.41.0, and now it's restored again)
+;; (This was our behavior for many years,
+;; then had to be disabled for test suite reasons in Orchard 0.15.0 to 0.17.0 / cider-nrepl 0.38.0 to 0.41.0, and now it's restored again)
 ;; Note that this can only be done cider-nrepl side, unlike before when it was done in Orchard itself.
 (def initializer
   (future
+    ;; 1.- Warmup the overall cache for core Java and Clojure stuff
     @orchard.java/cache-initializer
+    ;; 2.- Also cache members involved in a typical exception (which includes stuff from nrepl, clojure.repl, etc)
+    (try
```
bbatsov, line 38: I still don't get what happens in Haystack that takes so much time, though, so we'd need this warm up code.

bbatsov: (I'm also guessing we can just load explicitly whatever's not getting loaded before the first usage, instead of having to trigger an actual exception. )

vemv: > I still don't get what happens in Haystack that takes so much time, though, so we'd need this warm up code.

It's its N `orchard/info*` paths. When Enrich is available, we are able to call the (JDK-official) Java source parser which happens to be slow - about 1s per class.

> (I'm also guessing we can just load explicitly whatever's not getting loaded before the first usage, instead of having to trigger an actual exception. )

I'd be afraid of that fixed list becoming outdated at some point, which would mean we'd have a performance regression (for first hits, that is)

bbatsov: OK, fair enough. 

> Java source parser which happens to be slow - about 1s per class.

That seems way too much to me. Perhaps it can be optimized down the road.

vemv: I agree. Last time I had a look it was essentially an issue in this official JDK offering.

I might make some noise in some mailing list, but it would feel pretty futile.

PR #831, opened 2023-11-02
Introduce `inspect-last-stacktrace-exception` op

vemv:
* Avoid problematic newline in docs
  * See https://docs.cider.mx/cider-nrepl/nrepl-api/ops.html#info
* Introduce `inspect-last-stacktrace-exception` op
  * Part of https://github.com/clojure-emacs/cider/issues/3565
* `lein docs`: Add missing `shutdown-agents` call

Cheers - V

### DIFF
--- a/src/cider/nrepl.clj
+++ b/src/cider/nrepl.clj
@@ -245,7 +245,6 @@ Depending on the type of the return value of the evaluation this middleware may 
    "ns"      "The current namespace"
    "context" "A Compliment completion context, just like the ones already passed for the \"complete\" op,
 with the difference that the symbol at point should be entirely replaced by \"__prefix__\".
-
 For Java interop queries, it helps inferring the precise type of the object the `:sym` or `:member` refers to,
 making the results more accurate (and less numerous)."
    "class"  "A Java class. If `:ns` is passed, it will be used for fully-qualifiying the class, if necessary."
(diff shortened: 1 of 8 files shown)


src/cider/nrepl.clj
```diff
@@ -231,7 +231,6 @@ Depending on the type of the return value of the evaluation this middleware may
    "ns"      "The current namespace"
    "context" "A Compliment completion context, just like the ones already passed for the \"complete\" op,
 with the difference that the symbol at point should be entirely replaced by \"__prefix__\".
-
```
bbatsov, line 234: That's problematic about it?

vemv: https://docs.cider.mx/cider-nrepl/nrepl-api/ops.html#info

There are these broken bullet points

image

bbatsov:
So, to be clear that's supposed to analyze the top-most exception for the last error's staketrace, right? I'm asking mostly because the name is a bit unclear.

vemv:
It inspects (with the Inspector) the last exception that has done through `*cider-error*`. `:index` indicates which exception of it you want to see, the top-level (0), or one of its causes (1-N)

The idea is that within `*cider-error*`, each region of these will trigger these calls for a given `:index` when hitting <kbd>p</kbd>:

image

bbatsov:
My bad - I meant to write "inspect", not "analyze". Just wanted to make sure which part of the stacktrace is considered `last`. I'm wondering if there's a better term/name for this, as `last-stacktrace-exception` definitely confused me a bit.

vemv:
`last` simply means "the latest exception", i.e. `*e`.

Part of the confusion stems from `analyze-last-stacktrace` which I think should have been named `analyze-latest-stacktrace`.

...Thinking about it, we can remove "stacktrace" from `inspect-last-stacktrace-exception`, to `inspect-last-exception`. In the end, this op doesn't really handle stacktraces.

It becomes less evident that this op is closely related to `analyze-last-stacktrace`, though.

bbatsov:
I think `inspect-last-exception` is the best name we can pick at this point.

vemv:
SGTM!

PR #834, opened 2023-11-05
Fix `"inspect-refresh"` op

vemv:
It broke at some point, possibly long ago, as it was showing the same inspector instead of a refreshed one.

Cheers - V

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -123,7 +123,8 @@
                                    [leiningen-core "2.9.10" :exclusions [org.clojure/clojure
                                                                          commons-codec
                                                                          com.google.code.findbugs/jsr305]]
-                                   [cider/piggieback "0.5.3"]]}
+                                   [cider/piggieback "0.5.3"]
+                                   [nubank/matcher-combinators "3.8.8"]]}
 
              ;; Running the tests with enrich-classpath doing its thing isn't compatible with `lein test`
              ;; (because there's no such thing as "run `lein test` using this specific classpath"),
--- a/test/clj/cider/nrepl/middleware/test_test.clj
+++ b/test/clj/cider/nrepl/middleware/test_test.clj
@@ -1,13 +1,15 @@
 (ns cider.nrepl.middleware.test-test
   (:require
    [cider.nrepl.middleware.test :as test]
-   ;; Ensure tested tests are loaded:
-   cider.nrepl.middleware.test-filter-tests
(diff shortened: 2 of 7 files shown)


vemv:
Hmmm an unrelated test broke:

```
FAIL in (print-object-test) (test_test.clj:263)
uses println for matcher-combinators results, otherwise invokes pprint
println is chosen, as indicated by strings printed without quotes
```

strangely enough we were testing for matcher-combinators results, but we had't that library in. Probably this indicates a bug from our side.

project.clj
```diff
@@ -123,7 +123,8 @@
                                    [leiningen-core "2.9.10" :exclusions [org.clojure/clojure
                                                                          commons-codec
                                                                          com.google.code.findbugs/jsr305]]
-                                   [cider/piggieback "0.5.3"]]}
+                                   [cider/piggieback "0.5.3"]
+                                   [nubank/matcher-combinators "3.8.8"]]}
```
bbatsov, line 127: Isn't it a bit of an overkill to introduce a new dep (even if it's just for tests) for a single usage?

vemv: We were lucky to introduce it, since it revealed a bug in our matcher-combinators integration

And answering to the q, Inspector resutls can be quite intrincate so the alternative was quite verbose.

This library also simplified a bespoke mechanism in Orchard tests (the deleted `t/assert-expr 'match?`) https://github.com/clojure-emacs/orchard/commit/9cb236c875773cd90f79aa78e163512fdd8dabb4

test/clj/cider/nrepl/middleware/test_test.clj
```diff
@@ -261,7 +263,9 @@ The `988` value reflects that it times things correctly for a slow test.")
 (deftest print-object-test
   (testing "uses println for matcher-combinators results, otherwise invokes pprint"
     (is (= "{no quotes}\n"
-           (#'test/print-object (with-meta {"no" "quotes"} {:type :matcher-combinators.clj-test/mismatch})))
+           (#'test/print-object (matcher-combinators.clj-test/tagged-for-pretty-printing
+                                 '(not (match? 1 2))
+                                 {:matcher-combinators.result/value {"no" "quotes"}})))
```
vemv, line 268: Updated this test to match this library's actual structure for a typical test failure.

vemv (commented):
Build will pass now after fixing the semi-related broken test

PR #844, opened 2024-01-14
Forcibly exit the JVM on older Clojure versions

vemv:
See https://github.com/nrepl/nrepl/issues/308

### DIFF
--- a/src/cider/nrepl.clj
+++ b/src/cider/nrepl.clj
@@ -22,6 +22,11 @@
    [orchard.namespace]
    [orchard.java]))
 
+(when (and (< (-> *clojure-version* :major long) 2)
+           (< (-> *clojure-version* :minor long) 9))
+  (.println System/err "cider-nrepl cannot be run with older Clojure versions. Exiting.")
+  (System/exit 1))
+
 ;; Perform the underlying dynamic `require`s asap, and also not within a separate thread
 ;; (note the `future` used in `#'initializer`),
 ;; since `require` is not thread-safe:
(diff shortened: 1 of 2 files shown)


src/cider/nrepl.clj
```diff
@@ -22,6 +22,11 @@
    [orchard.namespace]
    [orchard.java]))
 
+(when (and (< (-> *clojure-version* :major long) 2)
```
bbatsov, line 25: I'd probably move the required version to a constant (a map would be best I guess), so it's clearer what we're doing.

src/cider/nrepl.clj
```diff
@@ -22,6 +22,11 @@
...
+           (< (-> *clojure-version* :minor long) 9))
+  (.println System/err "cider-nrepl cannot be run with older Clojure versions. Exiting.")
```
bbatsov, line 27: Might be good to mention the required Clojure version here.

vemv:
Good suggestions, thanks!

src/cider/nrepl.clj
```diff
@@ -22,10 +22,24 @@
...
+
+(when (or (< (-> *clojure-version* :major long)
+             (-> min-clojure-verion :major long))
+
+          (and (= (-> *clojure-version* :major long)
+                  (-> min-clojure-verion :major long))
+               (< (-> *clojure-version* :minor long)
+                  (-> min-clojure-verion :minor long))))
+  (try
+    (.println System/err (format "cider-nrepl cannot be run with older Clojure versions (found: %s). Exiting."
```
bbatsov, line 39: I still think it'd be good to mention here the required version as well, so it's clearer for the users what they need.

vemv: Sure thing! https://github.com/clojure-emacs/cider-nrepl/commit/4497c166919d84fa6a6833f2baaf35cfe0687d14

Issue #848, opened 2024-02-08
`make repl` fails to start a REPL


metayan:
## Expected behavior
REPL should appear

## Actual behavior
```
Error: Could not find or load main class clojure.main
Caused by: java.lang.ClassNotFoundException: clojure.main
make: *** [Makefile:171: lein-repl] Error 1
```

## Steps to reproduce the problem
```
make repl
```

### cider-nrepl version
Current master
v0.45.0
v0.44.0

### Java version
```
openjdk version "19.0.1" 2022-10-18
OpenJDK Runtime Environment Homebrew (build 19.0.1)
OpenJDK 64-Bit Server VM Homebrew (build 19.0.1, mixed mode, sharing)
```
```
openjdk version "17.0.10" 2024-01-16
OpenJDK Runtime Environment (build 17.0.10+7-Debian-1deb12u1)
OpenJDK 64-Bit Server VM (build 17.0.10+7-Debian-1deb12u1, mixed mode, sharing)
```

### Operating system
`macOS 10.15.7`
`Linux Devuan Daedalus`

### Leiningen & Clojure versions
`Leiningen 2.11.1 on Java 19.0.1 OpenJDK 64-Bit Server VM`
`Clojure CLI version 1.11.1.1435`

vemv:
Hi!

Thank you for using CIDER (and presumably, enrich-classpath?).

There's not enough information in this issue for it to be actionable. Plese attach stacktraces if any. The issue should also be reproducible in a project that we can clone.

Feel free to reopen the issue as you gather that info.

metayan:
> There's not enough information in this issue for it to be actionable. Plese attach stacktraces if any. The issue should also be reproducible in a project that we can clone.

Probably a misunderstanding.
I'm running `make repl` in the `cider-nrepl` project directory - as is described in the README:

https://github.com/clojure-emacs/cider-nrepl/blob/9aea519b87430c3a26cd146417dd84c33bb14676/README.md?plain=1#L72-L73

I'm actually trying to investigate an issue in `cider`.

vemv:
Sorry about the misunderstanding - I thought this issue had been created in https://github.com/clojure-emacs/cider

Either way, `make repl` works for me.

But `lein repl` should work as well. Features related to `enrich-classpath` won't be present, however, most times one doesn't need to debug that.

metayan:
> `make repl` works for me.

Can you please reveal which version of Clojure/Lein/Java it works on?

[...]

PR #850, opened 2024-02-23
Support clj-reload workflow

### LINKED ISSUE #849: Support for custom `cider-ns-refresh-fn`
I'd like to use https://tonsky.me/blog/clj-reload/ instead of `clojure.tools.namespace` to refresh my code. The main thing I like it in is how `defonce` just works, and I don't have to fiddle around with `clojure.tools.namespace` in my source code.

I see there's `cider-ns-refresh-before-fn` and `cider-ns-refresh-after-fn` in [the docs](https://docs.cider.mx/cider/usage/misc_features.html#reloading-code), but there's no configurable `cider-ns-refresh-fn`  at least in those docs. So that seemed like a nice hook point.

I'd be interested in making a PR for this approach, or some other approach you feel is better. Thanks for all the good work!

filipesilva:
Fix #849

- [x] All tests are passing

Thanks!

### DIFF
--- a/src/cider/nrepl.clj
+++ b/src/cider/nrepl.clj
@@ -635,6 +635,25 @@ if applicable, and re-render the updated value."
              "refresh-clear"
              {:doc "Clears the state of the refresh middleware. This can help recover from a failed load or a circular dependency error."}}})
 
+(def-wrapper wrap-reload cider.nrepl.middleware.reload/handle-reload
+  {:doc "Reload middleware."
+   :requires #{"clone" #'wrap-print}
+   :handles {"cider.clj-reload/reload"
+             {:doc "Reloads all changed files in dependency order,
+using the io.github.tonsky/clj-reload library. It is bundled with cider-nrepl.
+If that dependency is already in present your project and clj-reload.core/init has been invoked beforehand,
+those configured directories will be honored."
+              :returns {"progress" "Description of current namespace being unloaded/loaded."
+                        "status" "`:ok` if reloading was successful; otherwise `:error`."
+                        "error" "A sequence of all causes of the thrown exception when `status` is `:error`."}}
+             "cider.clj-reload/reload-all"
+             {:doc "Reloads all files in dependency order."
+              :returns {"reloading" "Description of current namespace being unloaded/loaded."
+                        "status" "`:ok` if reloading was successful; otherwise `:error`."
+                        "error" "A sequence of all causes of the thrown exception when `status` is `:error`."}}
+             "cider.clj-reload/reload-clear"
+             {:doc "Clears the state of clj-reload. This can help recover from a failed load or a circular dependency error."}}})
+
 (def-wrapper wrap-resource cider.nrepl.middleware.resource/handle-resource
   {:doc "Middleware that provides the path to resource."
    :handles {"resource"
--- a/src/cider/nrepl/middleware/reload.clj
+++ b/src/cider/nrepl/middleware/reload.clj
@@ -1,0 +1,67 @@
+(ns cider.nrepl.middleware.reload
+  "Reload changed namespaces.
+  Alternative to cider.nrepl.middleware.refresh, using clj-reload instead
+  of tools.namespace."
+  (:require
+   [clj-reload.core :as reload]
+   [clojure.main :refer [repl-caught]]
+   [clojure.string :as str]
+   [haystack.analyzer :as analyzer]
+   [nrepl.middleware.interruptible-eval :refer [*msg*]]
+   [nrepl.middleware.print :as print]
+   [nrepl.misc :refer [response-for]]
+   [nrepl.transport :as transport]
+   [orchard.misc :as misc]))
+
+(defn- user-reload
+  "Resolve clj-reload.core/<sym> from the user project or return fallback."
+  [sym fallback]
+  (or (some-> (symbol "clj-reload.core" (str sym)) ;; Don't use mrandorsenized version
+              resolve)
+      fallback))
+
+(defn- init
+  "Initialize clj-reload with dirs.
+  Only used for test, but necessary because of mranderson."
+  [dirs]
+  (reload/init {:dirs dirs}))
+
+(defn respond
+  [{:keys [transport] :as msg} response]
+  (transport/send transport (response-for msg response)))
+
+(defn operation
+  [msg]
+  (let [opts   {:log-fn (fn [& args]
+                          (respond msg {:progress (str/join " " args)}))}
+        reload (user-reload 'reload reload/reload)
+        unload (user-reload 'unload reload/unload)]
+    (cond
+      (:all msg)   (reload (assoc opts :all true))
+      (:clear msg) (unload opts)
+      :else        (reload opts))))
+
+(defn- reload-reply
+  [{:keys [::print/print-fn transport session id] :as msg}]
+  (let [{:keys [exec]} (meta session)]
+    (exec id
+          (fn []
+            (try
+              (operation msg)
+              (respond msg {:status :ok})
+              (catch Throwable error
+                (respond msg {:status :error
+                              ;; TODO assoc :file, :line info if available
+                              :error  (analyzer/analyze error print-fn)})
+                (binding [*msg* msg
+                          *err* (print/replying-PrintWriter :err msg {})]
+                  (repl-caught error)))))
+
+          (fn [] (respond msg {:status :done})))))
+
+(defn handle-reload [handler msg]
+  (case (:op msg)
+    "cider.clj-reload/reload" (reload-reply msg)
+    "cider.clj-reload/reload-all" (reload-reply (assoc msg :all true))
+    "cider.clj-reload/reload-clear" (reload-reply (assoc msg :clear true))
+    (handler msg)))
--- a/README.md
+++ b/README.md
@@ -136,6 +136,7 @@ Let's also acknowledge some of the projects leveraged by cider-nrepl:
 * [clj-suitable][] - for ClojureScript code completion using runtime inspection
 * [tools.trace][] - for tracing
 * [tools.namespace][] - for namespace reloading
+* [clj-reload][] - for namespace reloading
 * [cljfmt][] - for code formatting
 
 ## License
@@ -149,6 +150,7 @@ Distributed under the Eclipse Public License, the same as Clojure.
 [clj-suitable]: https://github.com/clojure-emacs/clj-suitable
 [tools.trace]: https://github.com/clojure/tools.trace
 [tools.namespace]: https://github.com/clojure/tools.namespace
+[clj-reload]: https://github.com/tonsky/clj-reload
 [cljfmt]: https://github.com/weavejester/cljfmt
 [vim-replant]: https://github.com/SevereOverfl0w/vim-replant
 [vim-fireplace]: https://github.com/tpope/vim-fireplace
--- a/doc/modules/ROOT/pages/nrepl-api/ops.adoc
+++ b/doc/modules/ROOT/pages/nrepl-api/ops.adoc
@@ -1783,3 +1783,54 @@ Returns::
 * `:status` done
 * `:cider/log-update-consumer` The consumer that was updated.
 
+
+
+=== `cider.clj-reload/reload`
+
+Reloads all changed files in dependency order,
+using the io.github.tonsky/clj-reload library. It is bundled with cider-nrepl.
+If that dependency is already in present your project and clj-reload.core/init has been invoked beforehand,
+those configured directories will be honored.
+
+Required parameters::
+{blank}
+
+Optional parameters::
+{blank}
+
+Returns::
+* `:error` A sequence of all causes of the thrown exception when ``status`` is ``:error``.
+* `:progress` Description of current namespace being unloaded/loaded.
+* `:status` ``:ok`` if reloading was successful; otherwise ``:error``.
+
+
+
+=== `cider.clj-reload/reload-all`
+
+Reloads all files in dependency order.
+
+Required parameters::
+{blank}
+
+Optional parameters::
+{blank}
+
+Returns::
+* `:error` A sequence of all causes of the thrown exception when ``status`` is ``:error``.
+* `:reloading` Description of current namespace being unloaded/loaded.
+* `:status` ``:ok`` if reloading was successful; otherwise ``:error``.
+
+
+
+=== `cider.clj-reload/reload-clear`
+
+Clears the state of clj-reload. This can help recover from a failed load or a circular dependency error.
+
+Required parameters::
+{blank}
+
+Optional parameters::
+{blank}
+
+Returns::
+{blank}
--- a/doc/modules/ROOT/pages/usage.adoc
+++ b/doc/modules/ROOT/pages/usage.adoc
@@ -47,6 +47,7 @@ under `:repl-options`.
                   cider.nrepl/wrap-spec
                   cider.nrepl/wrap-profile
                   cider.nrepl/wrap-refresh
+                  cider.nrepl/wrap-reload
                   cider.nrepl/wrap-resource
                   cider.nrepl/wrap-stacktrace
                   cider.nrepl/wrap-test
@@ -161,6 +162,7 @@ That's how CIDER's nREPL handler is created:
     cider.nrepl/wrap-slurp
     cider.nrepl/wrap-profile
     cider.nrepl/wrap-refresh
+    cider.nrepl/wrap-reload
     cider.nrepl/wrap-resource
     cider.nrepl/wrap-spec
     cider.nrepl/wrap-stacktrace
--- a/project.clj
+++ b/project.clj
@@ -19,6 +19,7 @@
                  ~(with-meta '[org.clojure/tools.namespace "1.3.0"]
                     ;; :cognitest uses tools.namespace, so we cannot inline it while running tests.
                     {:inline-dep (not= "true" (System/getenv "SKIP_INLINING_TEST_DEPS"))})
+                 ^:inline-dep [io.github.tonsky/clj-reload "0.4.0"]
                  ^:inline-dep [org.clojure/tools.trace "0.7.11"]
                  ^:inline-dep [org.clojure/tools.reader "1.3.6"]
                  [mx.cider/logjam "0.3.0"]]
@@ -157,6 +158,7 @@
                                                               cider.nrepl/wrap-out
                                                               cider.nrepl/wrap-profile
                                                               cider.nrepl/wrap-refresh
+                                                              cider.nrepl/wrap-reload
                                                               cider.nrepl/wrap-resource
                                                               cider.nrepl/wrap-slurp
                                                               cider.nrepl/wrap-spec
--- a/src/cider/nrepl/middleware.clj
+++ b/src/cider/nrepl/middleware.clj
@@ -23,6 +23,7 @@
     cider.nrepl/wrap-slurp
     cider.nrepl/wrap-profile
     cider.nrepl/wrap-refresh
+    cider.nrepl/wrap-reload
     cider.nrepl/wrap-resource
     cider.nrepl/wrap-spec
     cider.nrepl/wrap-stacktrace
--- a/test/clj/cider/nrepl/middleware/reload_test.clj
+++ b/test/clj/cider/nrepl/middleware/reload_test.clj
@@ -1,0 +1,46 @@
+(ns cider.nrepl.middleware.reload-test
+  (:require
+   [cider.nrepl.middleware.reload :as rl]
+   [cider.nrepl.test-session :as session]
+   [clojure.test :refer :all]))
+
+(use-fixtures :each session/session-fixture)
+
+(def ^:private dirs-to-reload
+  ;; Limit the scope of what we reload, because (for example) reloading the
+  ;; cider.nrepl.middleware.test-session ns causes *session* in that ns to be
+  ;; unloaded, which breaks session-fixture, and hence all of the below tests.
+  ["test/clj/cider/nrepl/middleware/util"])
+
+;; Calling init from reload ns to work around mrandersonized version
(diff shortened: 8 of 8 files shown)


src/cider/nrepl.clj
```diff
@@ -632,6 +632,20 @@ if applicable, and re-render the updated value."
              "refresh-clear"
              {:doc "Clears the state of the refresh middleware. This can help recover from a failed load or a circular dependency error."}}})
 
+(def-wrapper wrap-refresh cider.nrepl.middleware.reload/handle-reload
+  {:doc "Reload middleware."
+   :requires #{"clone" #'wrap-print}
+   :handles {"reload"
```
vemv, line 638:
```suggestion
   :handles {"cider.clj-reload/reload"
```

New ops should use ns-qualified naming.

(`cider/reload` would not be not sufficiently clear - we should make it clear that this is a wrapper for a specific library)

src/cider/nrepl/middleware/reload.clj
```diff
@@ -0,0 +1,49 @@
...
+
+(defn respond
+  [{:keys [transport] :as msg} response]
+  (transport/send transport (response-for msg response)))
+
+(defn- refresh-reply
+  [{:keys [::print/print-fn transport session id] :as msg}]
+  (let [{:keys [exec]} (meta session)]
+    (exec id
+          #(try
```
vemv, line 27: Nit - please use `fn`

https://stuartsierra.com/2019/09/11/clojure-donts-short-fn-syntax

src/cider/nrepl/middleware/reload.clj
```diff
@@ -0,0 +1,49 @@
...
+             (let [reload (user-reload)]
+               (when-not reload
+                 (throw (ex-info "clj-reload.core/reload not found" {})))
```
vemv, line 30: Looks like we don't need to be this stringent, following https://github.com/tonsky/clj-reload/commit/0a082e9de248e9b2440999da3ccd4c60780422e5

src/cider/nrepl/middleware/reload.clj
```diff
@@ -0,0 +1,49 @@
...
+               (binding [*msg* msg
+                         *err* (print/replying-PrintWriter :err msg {})]
+                 (repl-caught error))))
+
+          #(respond msg {:status :done}))))
+
+(defn handle-reload [handler msg]
+  (case (:op msg)
+    "reload" (refresh-reply msg)
+    "reload-all" (refresh-reply (assoc msg :all true))
```
vemv, line 47: Is there not a clear op?

src/cider/nrepl/middleware/reload.clj
```diff
@@ -0,0 +1,49 @@
+(ns cider.nrepl.middleware.reload
+  (:require
+   [clojure.main :refer [repl-caught]]
+   [clojure.string :as str]
+   [haystack.analyzer :as stacktrace.analyzer]
```
vemv, line 5:
```suggestion
   [haystack.analyzer :as analyzer]
```

The `stacktrace.analyzer` rename is sort of a legacy thing

src/cider/nrepl/middleware/reload.clj
```diff
@@ -0,0 +1,49 @@
...
+             (let [reload (user-reload)]
+               (when-not reload
+                 (throw (ex-info "clj-reload.core/reload not found" {})))
+               (reload (cond-> {:log-fn (fn [& args]
+                                          (respond msg {:progress (str/join " " args)}))}
+                               (:all msg) (assoc :only :all)))
+               (respond msg {:status :ok}))
+             (catch Throwable error
+               (respond msg {:status :error
+                             :error  (stacktrace.analyzer/analyze error print-fn)})
```
vemv, line 37:
```suggestion
                             ;; TODO assoc :file, :line info if available
                             :error  (stacktrace.analyzer/analyze error print-fn)})
```

I'd like this to eventually return file/line info if we can easily grab it from the analyzer. This way cider.el could automatically open the culprit file.

(Feel free to give it a shot, but simply adding the TODO is perfectly fine)

src/cider/nrepl.clj
```diff
@@ -632,6 +632,20 @@ if applicable, and re-render the updated value."
...
+             {:doc "Reloads all changed files in dependency order."
```
vemv, line 639:
```suggestion
             {:doc "Reloads all changed files in dependency order,
using the io.github.tonsky/clj-reload library. It is bundled with cider-nrepl.

If that dependency is already in present your project and clj-reload.core/init has been invoked beforehand,
those configured directories will be honored."
```

src/cider/nrepl/middleware/reload.clj
```diff
@@ -0,0 +1,49 @@
...
+   [nrepl.middleware.print :as print]
+   [nrepl.misc :refer [response-for]]
+   [nrepl.transport :as transport]
+   [orchard.misc :as misc]))
+
+(defn- user-reload
+  "clj-reload.core/reload from the user project.
+  Must be configured via clj-reload.core/init before being called."
+  []
+  (some-> (symbol "clj-reload.core" "reload")
```
vemv, line 16:
```suggestion
  (some-> (symbol "clj-reload.core" "reload") ;; Don't use mrandorsenized version
```

vemv (commented):
Looking good, thanks!

Some suggestions

src/cider/nrepl/middleware/reload.clj
filipesilva: Oh that's right! We can use the project one if available, and fall back to the bundled one if not.

src/cider/nrepl/middleware/reload.clj
filipesilva: Hm no, I can't see it in https://github.com/tonsky/clj-reload/blob/main/src/clj_reload/core.clj

src/cider/nrepl/middleware/reload.clj
filipesilva: Added the TODO for now, but may circle back later if I can manage it.

filipesilva:
@vemv running `PARSER_TARGET=parser-next make test` fails with:
```
Running tests in #{"/Users/filipesilva/work/cider-nrepl/test/spec" "/Users/filipesilva/work/cider-nrepl/test/clj" "/Users/filipesilva/work/cider-nrepl/test/common" "/Users/filipesilva/work/cider-nrepl/test/cljs"}
Execution error (FileNotFoundException) at cider.nrepl.middleware.reload-test/eval21172$loading (reload_test.clj:1).
Could not locate clj_reload/core__init.class, clj_reload/core.clj or clj_reload/core.cljc on classpath. Please check that namespaces with dashes use underscores in the Clojure file name.
```

But `PARSER_TARGET=parser-next make fast-test` does not fail. Do you know why?

vemv:
Thanks for applying the feedback!

I think it's a known issue with mranderson - see:

https://github.com/clojure-emacs/cider-nrepl/blob/3824d72f9ba1a14f9ce9910500db0885f6b39c23/test/clj/cider/nrepl/middleware/refresh_test.clj#L131-L136

You could perform the require dynamically, only running the deftest if `(try (require 'clj-reload.core) true (catch Exception _ false)) ;; only run the deftest if inlining hasn't been performed`

filipesilva:
@vemv all the tests in `cider.nrepl.middleware.reload-test` depend on the init, so disabling it means it wouldn't run at all in `make test`... WDYT of using init from the `cider.nrepl.middleware.reload` ns like I have in this last commit?

filipesilva:
That last commit seems to have sorted the cli-reload errors. But now a see two errors that I really don't understand....

```
FAIL in (instrument-function-call-test) (instrument_test.clj:109)
expected: #{[(System/currentTimeMillis) [3 1 1]] [(Thread/sleep 1000) [3 2]] [(- (bp (System/currentTimeMillis) {:coor [3 3 1]} (System/currentTimeMillis)) (bp start-time {:coor [3 3 2]} start-time)) [3 3]] [(System/currentTimeMillis) [3 3 1]] [(def test-fn (fn* ([] (bp (let* [start-time (bp (System/currentTimeMillis) {:coor [3 1 1]} (System/currentTimeMillis))] (bp (Thread/sleep 1000) {:coor [3 2]} (Thread/sleep 1000)) (bp (- (bp (System/currentTimeMillis) {:coor [3 3 1]} (System/currentTimeMillis)) (bp start-time {:coor [3 3 2]} start-time)) {:coor [3 3]} (- (System/currentTimeMillis) start-time))) {:coor [3]} (let [start-time (System/currentTimeMillis)] (Thread/sleep 1000) (- (System/currentTimeMillis) start-time)))))) []] [start-time [3 3 2]] [(let* [start-time (bp (System/currentTimeMillis) {:coor [3 1 1]} (System/currentTimeMillis))] (bp (Thread/sleep 1000) {:coor [3 2]} (Thread/sleep 1000)) (bp (- (bp (System/currentTimeMillis) {:coor [3 3 1]} (System/currentTimeMillis)) (bp start-time {:coor [3 3 2]} start-time)) {:coor [3 3]} (- (System/currentTimeMillis) start-time))) [3]]}
  actual: #{[(- (bp (. System currentTimeMillis) {:coor [3 3 1]} (System/currentTimeMillis)) (bp start-time {:coor [3 3 2]} start-time)) [3 3]] [(let* [start-time (bp (. System currentTimeMillis) {:coor [3 1 1]} (System/currentTimeMillis))] (bp (. Thread sleep 1000) {:coor [3 2]} (Thread/sleep 1000)) (bp (- (bp (. System currentTimeMillis) {:coor [3 3 1]} (System/currentTimeMillis)) (bp start-time {:coor [3 3 2]} start-time)) {:coor [3 3]} (- (System/currentTimeMillis) start-time))) [3]] [(. System currentTimeMillis) [3 1 1]] [(. Thread sleep 1000) [3 2]] [start-time [3 3 2]] [(. System currentTimeMillis) [3 3 1]] [(def test-fn (fn* ([] (bp (let* [start-time (bp (. System currentTimeMillis) {:coor [3 1 1]} (System/currentTimeMillis))] (bp (. Thread sleep 1000) {:coor [3 2]} (Thread/sleep 1000)) (bp (- (bp (. System currentTimeMillis) {:coor [3 3 1]} (System/currentTimeMillis)) (bp start-time {:coor [3 3 2]} start-time)) {:coor [3 3]} (- (System/currentTimeMillis) start-time))) {:coor [3]} (let [start-time (System/currentTimeMillis)] (Thread/sleep 1000) (- (System/currentTimeMillis) start-time)))))) []]}
```

```
FAIL in (inspect-var-integration-test) (inspect_test.clj:264)
rendering a var
expected: (= var-result (value (session/message {:op "eval", :inspect "true", :code "#'*assert*"})))
  actual: (not (= ("Class" ": " (:value "clojure.lang.Var" 0) (:newline) "Value: " (:value "true" 1) (:newline) (:newline) "--- Meta Information:" (:newline) "  " (:value ":ns" 2) " = " (:value "clojure.core" 3) (:newline) "  " (:value ":name" 4) " = " (:value "*assert*" 5) (:newline) (:newline) "--- Datafy:" (:newline) "  " "0" ". " (:value "true" 6) (:newline)) ("Class" ": " (:value "clojure.lang.Var" 0) (:newline) "Value: " (:value "true" 1) (:newline) (:newline) "--- Meta Information:" (:newline) "  " (:value ":added" 2) " = " (:value "\"1.0\"" 3) (:newline) "  " (:value ":ns" 4) " = " (:value "clojure.core" 5) (:newline) "  " (:value ":name" 6) " = " (:value "*assert*" 7) (:newline) "  " (:value ":doc" 8) " = " (:value "\"When set to logical false, 'assert' will omit assertion checks in\\n  compiled code. Defaults to true.\"" 9) (:newline) (:newline) "--- Datafy:" (:newline) "  " "0" ". " (:value "true" 10) (:newline))))
```

I didn't touch or I think interact with these ns's in any way, do you have any idea of what's happening?

filipesilva:
Should I do any changes to the README, as the PR template mentions?

I'd like to test locally a bit to make sure everything is fine, but hitting a failure... when I run
```
# Install the project in your local ~/.m2 directory, using mranderson (recommended)
# The JVM flag is a temporary workaround.
export LEIN_JVM_OPTS="-Dmranderson.internal.no-parallelism=true"
PROJECT_VERSION=0.45.0-reload make install
```
it fails with this message:
```
...
prefixing #{"difflib"} in target/class-deps.jar with cidernrepl0450reload
deleting directories with class files in target/srcdeps...
   difflib  deleted
unzipping repackaged class-deps.jar into target/srcdeps
touch .inline-deps
rm -f .no-mranderson
touch .no-pedantic
lein with-profile -user,-dev,+1.11,+plugin.mranderson/config install
Error: Could not find or load main class clojure.main
Caused by: java.lang.ClassNotFoundException: clojure.main
Compilation failed: Subprocess failed (exit code: 1)
Error encountered performing task 'install' with profile(s): 'base,system,provided,1.11,config'
Compilation failed: Subprocess failed (exit code: 1)
make: *** [install] Error 1
```

Am I doing something wrong? If there's a better way to test locally let me know.

vemv:
That looks like an occurrence of https://github.com/clojure-emacs/cider-nrepl/issues/848 - please try with a slightly older Lein?

filipesilva:
Yeap that worked! 

... but I'm at a bit of a loss of how to test it now. I mean CIDER won't have this operation, right? What should I do to test it in CIDER?

filipesilva:
Well if there's nothing more that I can test at this point, I don't think there's anything pending on my side. Just pushed the README update.

vemv:
Happy that you got it!

After a successful `make install`,  you could fork cider https://docs.cider.mx/cider/contributing/hacking.html so that the new ops are used insted of the tools.namespace ones.

There should be a few defcustoms e.g. `cider-refresh-op`, `cider-refresh-all-op`.

Should be a small change overall.

It also should be good timing to discuss upstream whether a `clear` function op would make sense for clj-reload. Else we'd have to void that op somewhow, to avoid surprises.

filipesilva:
What's the difference between the `clear` op and the `cider-refresh-all-op`?

I tried to setup CIDER locally but get a native compile error while doing `eldev build :autoloads`. I tried something different though: I went into nrepl.clj, changed `(def-wrapper wrap-refresh cider.nrepl.middleware.refresh/handle-refresh` to use `cider.nrepl.middleware.reload/handle-reload` instead, update `handle-reload` to handle the refresh op, `make install`, then load it in a project manually instead of via jack-in.

I could see `cider-ns-refresh` working and logging messages `Reloading successful`, but the current refresh functionality logs this message anyway. There's a `cider-ns-reload-all` but that doesn't actually seem related to the `cider-ns-refresh`, at least from the doc. I don't think `refresh-all` or `refresh-clear` are currently exposed as CIDER commands.

This is the test I used:
```clojure

(def foo 1)

(defonce bar 5)

(comment
  foo
  bar)
```

I saved, reloaded, and got eval'ed `foo` and `bar` in the comment block. Then altered their values, reloaded again, and saw foo update the value but bar remained. This is a different in `cli-reload` than in `clojure.tools.namespace.repl`: the former keeps `defonce`, but the latter does not. So I think this updated `cider-ns-refresh` is working with `cli-reload`. I double checked this test and with release version of CIDER the refresh does indeed update the value of the `defonce` var.

filipesilva:
What are you thoughts on using the new operation from within CIDER by the way? I guessed that a new command would be added, `cider-ns-reload` would be added, but there's already a `cider-ns-reload`. Maybe `cider-ns-clj-reload`?

vemv:
Hi Filipe!

> What's the difference between the clear op and the cider-refresh-all-op?

In general, `clear` unloads all code, without trying to load it again. This is useful in case something temporarily broke - then one can `clear`, fix the code, load it by hand, ensure it works, then `refresh` to load the rest of the proect.

> I don't think refresh-all or refresh-clear are currently exposed as CIDER commands.

Looks like they are: https://github.com/clojure-emacs/cider/blob/dc58ed1a411e6fe6f0350435aa4c56ec6edf59bf/cider-ns.el#L257-L261

> What are you thoughts on using the new operation from within CIDER by the way? I guessed that a new command would be added

I'd prefer if the defuns/commands remained the same. Instead, the ops would be customizable via defcustoms.

Cheers - V

filipesilva:
Ok so what's next? I don't think there's anything left for me to do in this PR, or is there?

vemv:
I'd like to know if there's an intention from clj-reload's side to offer `clear` functionality.

In the meantime, WIP PR welcome for https://github.com/clojure-emacs/cider - let us know if it's working as intended, using the suggested `defcustom` design.

filipesilva:
I see now all the refresh nrepl operations are within the same cider-ns-refresh, that's what I missed. 

It also makes a lot more sense why you'd like 
`clear` to also exist, that way the semantics of the operation are maintained.

But I don't see an exposed unload operation in clj-reload, it's not part of the API I think.

Maybe we could leave that one as not yet implemented and the interface would still be the same?

filipesilva:
@tonsky what are your thoughts about a clear/unload operation?

tonsky:
> In general, clear unloads all code, without trying to load it again. This is useful in case something temporarily broke - then one can clear, fix the code, load it by hand, ensure it works, then refresh to load the rest of the proect.

This? We can do this, sure. Send a PR or I’ll do it myself next week

filipesilva:
@vemv early draft of the cider changes in https://github.com/clojure-emacs/cider/pull/3624, PTAL if it matches what you had in mind.

src/cider/nrepl/middleware/reload.clj
```diff
@@ -0,0 +1,58 @@
...
+                          *err* (print/replying-PrintWriter :err msg {})]
+                  (repl-caught error)))))
+
+          (fn [] (respond msg {:status :done})))))
+
+(defn handle-reload [handler msg]
+  (case (:op msg)
+    "cider.clj-reload/reload" (refresh-reply msg)
+    "cider.clj-reload/reload-all" (refresh-reply (assoc msg :all true))
+    "cider.clj-reload/reload-clear" (refresh-reply (assoc msg :clear true))
```
filipesilva, line 56: @vemv left support for the clear op here so it exists when called from CIDER, but it's still not functional as per the TODO in this file.

src/cider/nrepl/middleware/reload.clj
```diff
@@ -0,0 +1,58 @@
...
+  [{:keys [transport] :as msg} response]
+  (transport/send transport (response-for msg response)))
+
+(defn- refresh-reply
+  [{:keys [::print/print-fn transport session id] :as msg}]
+  (let [{:keys [exec]} (meta session)]
+    (exec id
+          (fn []
+            (try
+              ;; TODO: clear op
```
filipesilva, line 36: @tonsky I think I'll wait until you implement the clear op, because I felt my init implementation was a rather far from what you ended up merging and unload seems to be more involved than init.

I think everything else is don't though. Today I tested the e2e of the feature in my emacs and it was working as expected. All users need to do is add `(setq cider-ns-refresh-tool 'clj-reload)`.

src/cider/nrepl/middleware/reload.clj
```diff
@@ -0,0 +1,58 @@
+(ns cider.nrepl.middleware.reload
```
bbatsov, line 1: This ns should have some high-level description in the form of a docstring.

src/cider/nrepl/middleware/reload.clj
```diff
@@ -0,0 +1,58 @@
...
+  "Initialize clj-reload with dirs.
+  Only used for test, but necessary because of mranderson."
+  [dirs]
+  (reload/init {:dirs dirs}))
+
+(defn respond
+  [{:keys [transport] :as msg} response]
+  (transport/send transport (response-for msg response)))
+
+(defn- refresh-reply
```
bbatsov, line 30: I guess you forgot to rename this.

bbatsov:
Yeah, it'd be nice to explain both here and in CIDER's docs (later) why are there 2 instances of relatively similar functionality. I'm guessing if `clj-reload` works well down the road we can remove `c.t.n.` as I don't think we need both of them in the long run and I like that `clj-reload` is smaller in terms of scope.

bbatsov:
> tbh that would be one big breaking change :) it's easy to imagine a variety of commercial projects that have set up the t.n boilerplate, so using t.n would be a natural fit.

Yeah, I'm well aware of this. I don't see this happening soon, but I also dislike functionality duplication. (it tends to be confusing for the end users) Now CIDER will have 3 ways to do code reloading, so can understand my desire for this not to be a permanent situation. Anyways, that's not particularly important for this PR.

src/cider/nrepl/middleware/reload.clj
filipesilva: Added

src/cider/nrepl/middleware/reload.clj
filipesilva: Yeap, initial copy paste then didn't look at it again :(

filipesilva:
Added `clear` (unload in clj-reload) support, rebase, and incorporated @bbatsov's feedback.

vemv:
Great work, thanks!

FYI a changelog entry was missing - no biggie, adding it now

cichli:
> In general, `clear` unloads all code, without trying to load it again. This is useful in case something temporarily broke - then one can `clear`, fix the code, load it by hand, ensure it works, then `refresh` to load the rest of the proect.

`clear` just clears the state of the refresh tracker – see [the doc for `clojure.tools.namespace.repl/clear`](https://clojure.github.io/tools.namespace/#clojure.tools.namespace.repl/clear) – it forces the next refresh to load everything, but doesn't unload anything itself. It's for recovering from corrupted states (normally caused by deleted files or circular dependencies). I think unloading is generally more useful and easier to understand, but we could maybe consider aliasing it to `unload` in the new middleware (sorry this reply is too late to suggest renaming it).

bbatsov:
@cichli Might be good to file this as a ticket, so the suggestion to add an alias won't be forgotten.

PR #856, opened 2024-03-10
`cider.nrepl.middleware.reload`: support `:before` / `:after` functions

vemv:
Moves shared code to a new ns.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,8 @@
 
 ## Bugs fixed
 
-* [#854](https://github.com/clojure-emacs/cider-nrepl/pull/854): fix cider.clj-reload/reload-all and improve its test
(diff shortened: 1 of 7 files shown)


bbatsov:
All other utility namespaces for middleware are under `cider.nrepl.middleware.util`, so I think this one should go there as well. (as `cider.nrepl.middleware.util.reload`.

src/cider/nrepl/middleware/reload_utils.clj
```diff
@@ -0,0 +1,84 @@
+(ns cider.nrepl.middleware.reload-utils
```
bbatsov, line 1: Don't forget to add here the usual metadata - docstring and `:added`.

vemv:
Thanks! All done.

PR #859, opened 2024-03-26
Add integration tests for next/previous-sibling ops

p4v4n:


vemv:
Thanks - very happy to get extended coverage!

I'll take a good look once the build is green

p4v4n:
> Thanks - very happy to get extended coverage!
> 
> I'll take a good look once the build is green

Not sure if it can be fixed though as it is failing only for the legacy parser.

I was afraid the additional metadata might cause some issue in the future but there was no way to add tests for the rendered content without them.

vemv:
No issue, will be eyeing

p4v4n:
> No issue, will be eyeing

The previous test failure shows a discrepancy between the parsers.
It didn't feel like a bug I could fix realistically so switched to a safer test case without much metadata on page.

vemv (commented):
I'd be OK with merging the test as-is (in the end it's extra coverage for an area where we had none), although I find these tests could be more expressive and exhaustive

* expressive in that one has to take for granted that `sibling-result` is a "good" result
  * what makes it good?
 * exhaustive in exercising a similar set of cases relative to the Orchard test suite
   * maybe not _as_ exhaustive, since that would seem an excessive duplication, but having 1 one case per op seems to have room for improvement.

Maybe, try coding an `are` with 2 to 4 cases for next-sibling and previous-sibling respectively.

Probably with it we'd gain both expressiveness, exhautiveness and some DRY.

Cheers - V

PR #863, opened 2024-03-31
Observe `cider-doc.edn` Java resource files for user-extensible documentation

### LINKED ISSUE #861: info: support user-extensible documentation providers
## Context

From time to time, there are DSLs that escape Clojure normal var system.

For example:

```clj
(with-foo ;; some DSL-based macro
  BAR BAZ ;; symbols that aren't backed by a var (or local variable)
))
```

While we can't reasonably provide documentation for unqualified symbols that escape the var system, users / lib authors should be able to have that, unobstrusively.

## Proposed solution

Users should be able to have a `dev/cider-doc.edn` resource with content such as:

```clj
{[BAR BAZ] ;; all the symbols for which to apply certain rules
 {:info-provider com.corp/foo-info-provider
  :if            com.corp/foo-context?}

 [quux quuz]
 {:info-provider com.corp/alt-info-provider
  :if            com.corp/alt-context?}}
```

Where: 

* `:info-provider` is a function that returns an 'info' map, as Orchard does
* `:if` is a predicate that takes a Compliment context, returns whether the symbol is relevant for the current context
  * e.g. `BAR` may have different meanings depending on the context, as it's an unqualified symbol.
  * One is free to pass `(constantly true)` if one is confident enough in that the given symbols will be unique enough.
  * Besides from invoking the predicate with a Compliment context, we can bind `*ns*` and `*file*` in case that helps users.

These .edn files (and their backing functions) could be distributed as normal Maven artifacts, so that people can add them to their `:dev` alias.

## Additional context

Our `info` middleware already takes a `context` as input.

https://github.com/clojure-emacs/cider-nrepl/blob/dd3a83a93b5cc406b539a903dde6e2b2889f50f1/src/cider/nrepl/middleware/info.clj#L81-L82

So it would seem easy and unobstrusive to observe `cider-doc.edn` files, handle them, and if they aren't found / do not apply, proceed with the normal code path.

Finally, it's OK to have multiple cider-doc.edn files in the classpath (just like with data_readers.clj - they're merged)

vemv:
> Fixes https://github.com/clojure-emacs/cider-nrepl/issues/861

Just a draft for now, will be trying it out for a while in a real project, but any ideas welcome.

### DIFF
--- a/src/cider/nrepl/middleware/info.clj
+++ b/src/cider/nrepl/middleware/info.clj
@@ -1,11 +1,12 @@
 (ns cider.nrepl.middleware.info
   (:require
-   [compliment.context]
-   [compliment.sources.class-members]
    [cider.nrepl.middleware.util :as util]
    [cider.nrepl.middleware.util.cljs :as cljs]
    [cider.nrepl.middleware.util.error-handling :refer [with-safe-transport]]
+   [clojure.edn :as edn]
    [clojure.string :as str]
+   [compliment.context]
+   [compliment.sources.class-members]
    [orchard.eldoc :as eldoc]
    [orchard.info :as info]
    [orchard.meta :as meta]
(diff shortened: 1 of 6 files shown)


src/cider/nrepl/middleware/info.clj
```diff
@@ -78,9 +79,51 @@
 (def var-meta-allowlist-set
   (set meta/var-meta-allowlist))
 
+(def DSLable?
+  (some-fn simple-symbol? ;; don't allow ns-qualified things, since the Clojure var system takes precedence over DSLs
+           simple-keyword?
+           string?))
+
+(defn cider-doc-edn-configs []
```
vemv, line 87: TODO - use a cache for not computing the configs every time...

test/resources/cider-doc.edn
```diff
@@ -0,0 +1,2 @@
+{[cider-doc-edn-example] {:info-provider cider.nrepl.middleware.info-test/sample-info-provider
+                          :if            cider.nrepl.middleware.info-test/sample-info-provider?}}
```
vemv, line 2: Besides from a function, `:if` could be a set of ns-qualified symbols:

```clj
:if #{clojure.core/future com.corp/my-macro ,,, }
```

which would mean: "this rule only applies if the symbol was found within one of these calls".

That would simplify user configs since, I reckon, most people won't want to do parsing themselves.

PR #868, opened 2024-04-23
Drop support for Clojure 1.9

alexander-yakushev:
I designated the new minimal version as 1.10.3. Or is it better to take 1.10.0? It doesn't matter to us, but can also encourage people to update to the latest patch version if they missed it.


bbatsov:
Looks good to me!

PR #869, opened 2024-04-30
Use new Orchard

alexander-yakushev:
Update Orchard to the latest version. This gets all latest inspector updates and new `orchard.print` namespace.

- [x] Adapt inspector middleware to updated `orchard.inspect`
  - [x] Fix tests
  - [x] Remove deprecated calls
  - [x] Handle nil inspector/empty config
  - [x] Add middleware opts for new printing options
  - [x] Fix some old bugs
- [x] Use new printing in cider debugger

  * [x] Please check out and modify the `cider.nrepl` ns which has all middleware documentation.
  * [x]  Run `lein docs` afterwards, and commit the results.

### DIFF
--- a/src/cider/nrepl.clj
+++ b/src/cider/nrepl.clj
@@ -364,6 +364,11 @@ if applicable, and re-render the updated value."
                :requires {"max-coll-size" "New collection size."
                           "session" "The current session"}
                :returns inspector-returns}
+              "inspect-set-max-nested-depth"
+              {:doc "Set the maximum nested levels to display before truncating."
+               :requires {"max-nested-depth" "New nested depth."
+                          "session" "The current session"}
+               :returns inspector-returns}
               "inspect-clear"
               {:doc "Clears the state state of the inspector."
                :requires {"session" "The current session"}
--- a/src/cider/nrepl.clj
+++ b/src/cider/nrepl.clj
@@ -302,7 +302,8 @@ If specified, the value will be concatenated to that of `orchard.meta/var-meta-a
                :returns {"status" "done"}}}}))
 
 (def inspector-returns (merge {"status" "\"done\""
-                               "value" "The inspector result. Contains a specially-formatted string that can be `read` and then rendered client-side."}
+                               "value" "The inspector result. Contains a specially-formatted string that can be `read` and then rendered client-side."
+                               "path" "Printed representation of current inspector path."}
                               fragments-doc))
 
 (def-wrapper wrap-inspect cider.nrepl.middleware.inspect/handle-inspect
@@ -335,10 +336,6 @@ if applicable, and re-render the updated value."
                :returns inspector-returns}
               "inspect-refresh"
               {:doc "Re-renders the currently inspected value."
-               :requires {"session" "The current session"}
-               :returns inspector-returns}
-              "inspect-get-path"
-              {:doc "Returns the path to the current position in the inspected value."
                :requires {"session" "The current session"}
                :returns inspector-returns}
               "inspect-next-page"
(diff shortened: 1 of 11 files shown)


src/cider/nrepl/middleware/inspect.clj
```diff
@@ -165,12 +175,13 @@
       "inspect-next-sibling" next-sibling-reply
       "inspect-previous-sibling" previous-sibling-reply
       "inspect-refresh" refresh-reply
-      "inspect-get-path" get-path-reply
       "inspect-next-page" next-page-reply
       "inspect-prev-page" prev-page-reply
       "inspect-set-page-size" set-page-size-reply
       "inspect-set-max-atom-length" set-max-atom-length-reply
+      "inspect-set-max-value-length" set-max-value-length-reply
```
vemv, line 182: While we can keep this, it would be also much welcome to have an options map so that users can simply have a defcustom.

This way we have fewer calls / moving pieces.

Existing example:

https://github.com/clojure-emacs/cider/blob/1cd6ab7b7cb4a7e079a273600c4d28307c3aba40/cider-client.el#L315-L322

alexander-yakushev: I agree that feels excessive to be able to customize all of those on the fly. Being able to set it up in one place would be better. At the same time, I don't know if a single set of options will fit everything. For example, it feels like the debugger can use stricter limitations than inspector (because debugger shows values inline with the code and inspector has a dedicated buffer).

I also think that different options have different utility in being configurable. E.g., customizing `page-size` on the fly is awesome. Being able to change `coll-length` and `coll-depth` on the fly in the inspector can be useful – if you need to see just a couple extra items or levels. But changing `atom-length` and `total-length` is of questionable benefit to me.

vemv: Yeah I don't oppose this op, and similarly, not everything has to be exposed as an option.

Simply trying to make sure that we can get the most useful options easily conveyed from cider.el to the Inspector.

alexander-yakushev: Here's what I would do:

1. Expose all five of these parameters as defcustoms in CIDER. Let the user customize these defaults as they wish.
2. Keep page-size and coll-length dynamic ops as is. Add a new coll-depth op, bind it to `C` (shift-c) in CIDER inspector to not hijack another letter for a pretty useless command (`c` is currently for changing coll-length).
3. Drop atom-length op from cider-nrepl. Make atom-length and value-length only changeable through defcustoms.

Thoughts?

vemv: I'd do 1 and 2. 3, I'd try really hard to avoid breaking changes.

Perhaps https://github.com/clojure-emacs/cider/discussions/2960 would be a good read - starting from that discussion (and increasingly since) we've been trying to commit to avoiding breakage.

alexander-yakushev:
> Re [clojure-emacs/orchard#227 (comment)](https://github.com/clojure-emacs/orchard/issues/227#issuecomment-2085593375) , maybe we can set that default here in advance

While there is a default of 50000 (soon to be 10000) in `orchard.inspect`, we will probably also have a default for `defcustom` in CIDER, also 10000, so it can be handled there.

alexander-yakushev:
Ready now.

src/cider/nrepl.clj
```diff
@@ -364,6 +361,11 @@ if applicable, and re-render the updated value."
                :requires {"max-coll-size" "New collection size."
                           "session" "The current session"}
                :returns inspector-returns}
+              "inspect-set-max-nested-depth"
```
vemv, line 364: Good to specifically add to the changelog

vemv (approved):
LGTM - great work!

vemv:
I've installed this. First impression is of course good.

The inspector is my "repl" so to speak so there will be a fair bit of usage over the next days.

PR #871, opened 2024-05-03
[inspector] Pass :spacious option to Orchard inspector

alexander-yakushev:
I feel stupid, does bencode support passing boolean false? :)

### DIFF
--- a/src/cider/nrepl/middleware/inspect.clj
+++ b/src/cider/nrepl/middleware/inspect.clj
@@ -65,7 +65,12 @@
 (defn inspect-reply*
   [msg value]
   (let [config (select-keys msg [:page-size :max-atom-length :max-coll-size
-                                 :max-value-length :max-nested-depth])
+                                 :max-value-length :max-nested-depth :spacious])
+        config (if (contains? config :spacious)
+                 (update config :spacious #(case %
+                                             ("false" "nil") false
+                                             true))
+                 config)
         inspector (swap-inspector! msg #(inspect/start (merge % config) value))]
     (warmup-javadoc-cache (class (:value inspector)))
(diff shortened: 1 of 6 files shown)


bbatsov:
It doesn’t support Boolean values.

src/cider/nrepl/middleware/inspect.clj
```diff
@@ -65,7 +65,10 @@
 (defn inspect-reply*
   [msg value]
   (let [config (select-keys msg [:page-size :max-atom-length :max-coll-size
-                                 :max-value-length :max-nested-depth])
+                                 :max-value-length :max-nested-depth :spacious])
+        config (if (contains? config :spacious)
+                 (update config :spacious #(if (= % "false") false %))
```
vemv, line 70: Seems worthwhile to also handle "true"

(I'm aware that `"true"` also can act as `true` but that kind of impurity is better coerced asap in the edges of our logic)

vemv (approved):

vemv:
I wasn't able to get it working - are you?

This is my CIDER request:

```
(-->
  id              "24"
  op              "eval"
  session         "1c9d380b-7a19-40c3-b6d8-9c4af1fbfb2d"
  time-stamp      "2024-05-03 17:41:50.089558000"
  code            "*1"
  inspect         "true"
  max-atom-length 150
  max-coll-size   5
  ns              "zzzzzzzzzz"
  page-size       32
  spacious        "false"
)
```

alexander-yakushev:
Yes, it works for me with a local install.

image

vemv:
My bad, I had Orchard 0.24 in my cp for other reasons.

PR #873, opened 2024-05-05
`middleware.refresh`: elide a lock for the `"refresh-clear"` op

vemv:
> Closes https://github.com/clojure-emacs/cider/issues/3652

### DIFF
--- a/src/cider/nrepl/middleware/refresh.clj
+++ b/src/cider/nrepl/middleware/refresh.clj
@@ -76,12 +76,18 @@
    msg]
   (reload-utils/after-reply error msg))
 
+(def ^{:added "0.48.0"} user-indicated-clear?
+  "Whether the user indicated that a `clear` is desired."
+  (atom false))
+
 (defn- refresh-reply
   [{:keys [dirs transport session id] :as msg}]
   (let [{:keys [exec]} (meta session)]
     (exec id
           (fn []
             (locking refresh-tracker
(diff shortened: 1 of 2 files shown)


src/cider/nrepl/middleware/refresh.clj
```diff
@@ -76,12 +76,18 @@
    msg]
   (reload-utils/after-reply error msg))
 
+(def ^{:added "0.48.0"} user-indicated-clear?
```
bbatsov, line 79: I don't think this name is good, as it's too specific and it's possible that some op might be triggered in a different way as opposed to a direct request from an IDE. (unlikely, but possible) Should this be named just `reset-in-progress` or something like this?

src/cider/nrepl/middleware/refresh.clj
```diff
@@ -107,8 +116,10 @@
   (let [{:keys [exec]} (meta session)]
     (exec id
           (fn []
-            (locking refresh-tracker
-              (vreset! refresh-tracker (track/tracker))))
+            ;; This used to be a `locking`-based call to our `clear` implementation
+            ;; Now it merely "enqueues" a clearing, because the "refresh-clear" op is called synchronously by CIDER,
```
bbatsov, line 120: It's best to avoid CIDER specifics here, as the middleware aims to be client-neutral outside of Emacs-specific features like Eldoc.

PR #879, opened 2024-05-30
[inspect] Introduce support for :view-mode config and inspect-toggle-view-mode op

alexander-yakushev:
1. Adding `:view-mode` as a supported config options passed from CIDER to Orchard. This part is straightforward.
2. Given that I wanted the view-mode to be a toggle, the best place for toggling logic is here. Adding it on CIDER side would require CIDER to know what is the current view-mode of the inspector which is extra work. cider-nrepl has that information though, so it's easy to do it here.

### DIFF
--- a/src/cider/nrepl.clj
+++ b/src/cider/nrepl.clj
@@ -340,7 +340,12 @@ if applicable, and re-render the updated value."
                :optional {"page-size" "New page size"
                           "max-atom-length" "New max length of single rendered value"
                           "max-coll-size" "New max size of rendered collection"
-                          "max-nested-depth" "New max nested depth of rendered collection"}
+                          "max-nested-depth" "New max nested depth of rendered collection"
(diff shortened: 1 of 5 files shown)


src/cider/nrepl.clj
```diff
@@ -340,7 +340,12 @@ if applicable, and re-render the updated value."
                :optional {"page-size" "New page size"
                           "max-atom-length" "New max length of single rendered value"
                           "max-coll-size" "New max size of rendered collection"
-                          "max-nested-depth" "New max nested depth of rendered collection"}
+                          "max-nested-depth" "New max nested depth of rendered collection"
+                          "view-mode" "Mode of viewing the value - either \"normal\" or \"object\""}
+               :returns inspector-returns}
+              "inspect-toggle-view-mode"
+              {:doc "Toggles the viewing mode of the inspector."
```
vemv, line 347: Good place to list those and briefly explain what they are

vemv (approved):
lgtm!

PR #882, opened 2024-06-02
Cleanup the docs

alexander-yakushev:
Notable changes:
- Removed mentions of Boot as part of the soft-deprecation process (@bbatsov).
- Trimmed the compatibility table to leave only versions where compatibilities have changed.

Ignore the 0.49.0 commit, I will rebase it when that PR is merged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## master (unreleased)
 
+## 0.49.0 (2024-06-02)
+
 ### Changes
 
 * [#877](https://github.com/clojure-emacs/cider-nrepl/pull/877): `inspect-refresh` middleware is now capable of setting all config options that `orchard.inspect` supports.
--- a/Makefile
+++ b/Makefile
@@ -109,7 +109,7 @@ light-kondo: clean
 
 lint: kondo cljfmt eastwood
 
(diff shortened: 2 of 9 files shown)


bbatsov:
> Trimmed the compatibility table to leave only versions where compatibilities have changed.

I'm fine with this, but let's add some `NOTE:` that states this explicitly, so users don't wonder why some versions are absent from the table.

alexander-yakushev:
Done. I feel like the matrix is still not complete because there is a huge gap between 0.25.0 and 0.44.0 but I'm not sure how to reconstruct the compatibilities of those in-between versions (and if it's worth doing).

bbatsov:
> Done. I feel like the matrix is still not complete because there is a huge gap between 0.25.0 and 0.44.0 but I'm not sure how to reconstruct the compatibilities of those in-between versions (and if it's worth doing).

It's not worth doing IMO. Probably we should cut some `cider-nrepl` 1.0 soon and just dispense will all the docs for the prior releases (same as I did for CIDER in the past). There's always work to be done on `cider-nrepl`, but the main API has been reasonably stable for a while now. Probably the only thing we really need to do before it is to namespace the ops as aliases to the existing names. (#710)

PR #883, opened 2024-06-13
Decrease minimal Clojure version in Leiningen plugin to 1.10.0

alexander-yakushev:
Turns out, it was a bad idea to set the minimal Clojure version to `1.10.3` - a patch version that was released much later than the original 1.10.0. Since we don't really care as long as it's 1.10, I propose to reduce the requirement.
Context: https://clojurians.slack.com/archives/C0617A8PQ/p1718210427715849


bbatsov:
Fine by me, just mention this in the changelog and we're good to go.

alexander-yakushev:
Done

PR #884, opened 2024-06-13
Don't apply indentation inference for `:style/indent nil` metadata

vemv:
> Fixes https://github.com/clojure-emacs/cider/issues/3684

### DIFF
--- a/src/cider/nrepl/middleware/track_state.clj
+++ b/src/cider/nrepl/middleware/track_state.clj
@@ -31,8 +31,8 @@
   [m]
   (and (:macro m)
        (:arglists m)
-       (not (:style/indent m))
-       (not (:indent m))
+       (not (contains? m :style/indent))
+       (not (contains? m :indent))
        (if-let [namespace-name (some-> (cond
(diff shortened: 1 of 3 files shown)


src/cider/nrepl/middleware/track_state.clj
```diff
@@ -31,8 +31,8 @@
   [m]
   (and (:macro m)
        (:arglists m)
-       (not (:style/indent m))
-       (not (:indent m))
+       (not (contains? m :style/indent))
+       (not (contains? m :indent))
```
alexander-yakushev, line 35: Is this a backward-compability wart? I've never seen anybody use `:indent`.

alexander-yakushev (commented):
Looks good!

src/cider/nrepl/middleware/track_state.clj
vemv: Me neither, but from memory, it's always been there internally

vemv:
Thanks for the eyeing!

PR #887, opened 2024-07-13
Fully evaluate lazy seqs when macroexpanding

### LINKED ISSUE #886: Macroexpansion with lazy seqs occurs in the wrong ns
## Expected behavior
Cider-nrepl's `macroexpansion` op should produce the same output as manually calling `macroexpand`.

## Actual behavior
If the macro definition contains lazily evaluated seqs, their contents will be evaluated with the wrong dynamic *ns* bindings (in the `user` ns), which can result in incorrect / misleading expansions.

```clj
(ns repro)

(defmacro lazy []
  (cons (ns-name *ns*)
        (lazy-seq [(ns-name *ns*)])))

(macroexpand '(lazy))
;; => (repro repro)

(lazy) ;; <- M-x cider-macroexpand: (repro user)
```

### cider-nrepl version
0.49.1

### Java version

21.0.1

### Operating system

MacOS 14

yuhan0:
Fixes #886

Note: I also added a `doall` in what seemed like the right spot for the cljs expander, but wasn't able to easily test whether it had any effect. There's no existing test coverage for that code path so I'm fine with removing it in case it breaks anything.

---

- [x] All tests are passing

Thanks!

### DIFF
--- a/src/cider/nrepl/middleware/macroexpand.clj
+++ b/src/cider/nrepl/middleware/macroexpand.clj
@@ -108,6 +108,13 @@
     "tidy" (tidy-walker-clj msg)
     (throw (IllegalArgumentException. (format "Unrecognized value for display-namespaces: %s" display-namespaces)))))
 
+(defn- doall-seq
+  "Ensure that macroexpansions containing lazy seqs are fully evaluated."
+  [expansion]
+  (if (seqable? expansion)
+    (doall expansion)
+    expansion))
+
 (defn- expand-clj
   "Returns the macroexpansion of the given Clojure form :code, performed in the
   context of the given :ns, using the provided :expander and :display-namespaces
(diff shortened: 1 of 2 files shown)


vemv (approved):
Thanks!

LGTM

yuhan0:
Oh.. I only ran `lein test`, good thing the CI caught this bug.  Somehow I couldn't replicate the test failure locally using `make tools-deps-test` but hopefully the following commit should fix it.

alexander-yakushev:
Unfortunately, it is not enough to call `doall` to force all internal parts of the constructed sexp. See https://clojure-goes-fast.com/blog/clojures-deadly-sin/#no-way-to-force-everything.

Try making a test case where the lazy part of the macro is somewhere deeper in the top-level expression. I think you'll have to `print-str` or something like that.

alexander-yakushev:
This PR is superseded by #894 which should handle the lazy expansion problem fully besides other things.

alexander-yakushev:
@yuhan0 Thanks for reporting the issue and proposing a fix! This led to uncovering a deeper problem with macroexpansion middleware.

PR #889, opened 2024-08-13
[complete] Bump Compliment to 0.6.0

alexander-yakushev:
This version brings new 1.12 completions for class members like `Thread/.interrupt` or `Thread/new`.


bbatsov:
Thanks!

PR #892, opened 2024-08-13
Exclude tools.cli from dependencies

alexander-yakushev:
Tools.cli is brought by cljfmt. In cljfmt, it's only used in `cljfmt.main` namespace to parse args in the main entrypoint. We don't touch that namespace in cider-nrepl, so it should be safe to drop this dependency.

https://github.com/search?q=repo%3Aweavejester%2Fcljfmt%20tools.cli&type=code

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,6 +39,8 @@ dump-version:
 .inline-deps: project.clj clean
 	rm -f .no-mranderson
 	lein with-profile -user,-dev inline-deps
+# Remove cljfmt.main because it depends on tools.cli which we explicitly removed.
+	rm -f target/srcdeps/cider/nrepl/inlined/deps/cljfmt/*/cljfmt/main.clj
 	touch $@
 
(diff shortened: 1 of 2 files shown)


alexander-yakushev:
Well, sh*t 😆.

alexander-yakushev:
Currently, cider-nrepl performs AOT compilation in two places:
- It AOT-compiles some inlined dependencies. I'm not sure why those and not the others. I think this can be disabled in the future.
- It AOTs all namespaces as part of the smoketest routine.

Anyway, I've added a hack that removes `cljfmt/main.clj` file after running through MrAnderson. This should take care of the problem in both cases.

bbatsov:
Looks a bit ugly, but I guess that will do. 

> It AOT-compiles some inlined dependencies. I'm not sure why those and not the others. I think this can be disabled in the future.

Same.

PR #893, opened 2024-08-14
[trace] Replace clojure.tools.trace with orchard.trace

alexander-yakushev:
Nothing changes on cider-nrepl API side. DId a small blunder of not publicly exposing `traced-nses` in some form from `orchard.trace`, will fix that in the next iteration. 

---


alexander-yakushev:
My bad, `0.26.3` should have been `0.27.0`, but here we are.

bbatsov:
Looks good!

> My bad, 0.26.3 should have been 0.27.0, but here we are.

Not a big deal, but if you want you can always cut one more release (0.27). For me it's fine either way.

PR #894, opened 2024-08-21
[macroexpand] Make macroexpand mw delegate the expansion to eval

alexander-yakushev:
This PR is a quite significant do-over of macroexpansion middleware.

The current implementation of CLJ macroexpander performs the expansion in the "middleware context". Because of that, the middleware has to carefully recreate an eval-like context (bind `*ns*`, ensure `LOADER` is bound), and problems like #886 still fall through. But we already have the eval-like context set up in the interruptible-eval middleware. This approach is already used by the `inspect` middleware for the initial inspector start command – it is a flavor of `eval` message that is intercepted by inspect mw on its way out.

I propose to do the same for macroexpansion – let the `eval` mw perform the expansion with all the correct dynamic bindings and context, just like if the user typed `(macroexpand 'form)` at the REPL. Then, intercept the resulting form and perform the "middleware work" (pretty printing, tidying up symbols) in the "middleware context".

---

- [x] All tests are passing

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -62,9 +62,9 @@ commands:
         default: "1"
     steps:
       - run:
-          name: Install wget, git, and node
+          name: Install node
           command: |
-            apt-get update && apt-get install -y git nodejs wget
+            apt-get update && apt-get install -y nodejs
       - run:
           name: Generate Cache Checksum
           command: |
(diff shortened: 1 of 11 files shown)


alexander-yakushev:
Removed tools-deps-testing .This test only exists to ensure LOADER is bound during macroexpansion. By deletaging the expansion to eval middleware, LOADER becomes the responsibility of nREPL which always ensures it is bound.

Besides, since there is no longer a direct function to macroexpand a form, recreating this test would require properly setting up a client-server integration test. I don't feel this is necessary anymore.

bbatsov:
I like the direction of this PR - admittedly I had totally forgot how we had implemented the macroexpansion in the past, and your proposal seems better in every way. (I'm not sure how you've managed to come across this, but I'm glad you did!) A simpler test setup is welcome as well.

bbatsov:
Impressive speedup! Great work! 🚀

PR #895, opened 2024-08-23
Bump Orchard to 0.27.0

alexander-yakushev:
Removed "Boot testing" section from tests – this looks like some leftover because it doesn't seem to me like it tests anything Boot-specific (or touches Boot at all).

Removed boot jars from `:test` profile – looks like this was meant for Eastwood to check `cider.tasks`.

`cider.tasks` for adding cider-nrepl to `boot repl` task still remains.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -55,7 +55,7 @@ commands:
       files:
         description: Files to consider when creating the cache key
         type: string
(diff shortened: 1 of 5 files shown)


bbatsov (approved):
Looks good to me!

PR #896, opened 2024-08-23
Redo testing

alexander-yakushev:
This PR gives the same treatment to CI/testing as https://github.com/clojure-emacs/orchard/pull/288.

1. Pull in JDK sources which were missing so that Java parsing capabilities can be tested.
2. Test parsers on JDK21 where we have the sources.
3. Simplify testing setup, remove dependency on enrich-classpath for testing, remove the necessity for Cognitest.
4. Move `smoketest` runs back into every `test` job since it already has sources MrAndersoned. Just don't clean them beforehand, so the smoketest is fast then.

---

- [x] All tests are passing

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -62,9 +62,9 @@ commands:
         default: "1"
     steps:
       - run:
-          name: Install node
+          name: Install packages
           command: |
-            apt-get update && apt-get install -y nodejs
+            apt-get update && apt-get install -y wget nodejs unzip zip
       - run:
           name: Generate Cache Checksum
           command: |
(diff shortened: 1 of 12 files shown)


alexander-yakushev:
The changes significantly cut the number of CI jobs which means Circle doesn't choke us anymore and the pipeline finishes faster.

bbatsov:
> The changes significantly cut the number of CI jobs which means Circle doesn't choke us anymore and the pipeline finishes faster (~7 minutes).

Great work!

bbatsov (approved):

theronic:
Which version of cider-nrepl includes this fix, please?

alexander-yakushev:
Sorry, which particular fix do you refer to? This PR is about CI, it doesn't impact cider-nrepl artifact much.

PR #897, opened 2024-08-27
0.50.0

alexander-yakushev:


README.md
```diff
@@ -81,13 +81,13 @@ PARSER_TARGET=parser-next make quick-test
 # Install the project in your local ~/.m2 directory, using mranderson (recommended)
 # The JVM flag is a temporary workaround.
 export LEIN_JVM_OPTS="-Dmranderson.internal.no-parallelism=true"
-PROJECT_VERSION=0.49.3 make install
+PROJECT_VERSION=99.99 make install
```
bbatsov, line 84: Probably something like `X.Y.Z` will be a bit better here to express a generic version.

bbatsov (approved):

README.md
alexander-yakushev: Since this is a `make install` example, to install locally, an exact version like this makes sense because it doesn't matter the exact number as long as it's something that doesn't look like a release version.

In the `git tag` example, I changed to `X.Y.Z`.

alexander-yakushev: Actually, OK, let it be the same for constistency.

PR #899, opened 2024-08-28
[test] Use Clojuredocs export.edn that is bundled with Orchard

alexander-yakushev:
Since Orchard already contains it, this saves us the trouble to download it again.


bbatsov (approved):
Fine by me. I think originally we didn't bundle the EDN file with Orchard an we probably forgot to update this. (we started to bundle it because the users were complaining that the first time they used the functionality they had to wait a lot)

PR #900, opened 2024-09-03
[macroexpand] Ensure print mw doesn't interfere with expansion

alexander-yakushev:
Context: https://clojurians.slack.com/archives/C0617A8PQ/p1725361078656679

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## master (unreleased)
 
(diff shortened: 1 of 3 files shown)


bbatsov:
Pretty curious bug! The fix seems good to me.

bbatsov (approved):

alexander-yakushev:
Yeah, it's obvious in retrospect that we should defend against this.

PR #906, opened 2024-12-06
Include generated input with each test.chuck `checking` block error

zjroth:
CIDER includes generated input in reports when the generated input is made available, which happens when using `test.chuck` and its `checking` block.  Currently, generated input is made available only to the first test condition within a `checking` block.  Unfortunately, this means that the generated input is not reported anywhere in CIDER unless the first test condition fails.  As a result, the generated input is (in my experience) typically not reported in CIDER.

This PR addresses that issue by modifying the `cider.nrepl.middleware.test/current-report` atom selectively based on whether a `test.chuck` testing context is available, which results in the presence of a non-`nil` value in the `:gen-input` key of the report precisely when the report corresponds to an error generated in a `checking` block.

cc: @daemianmack

### DIFF
--- a/src/cider/nrepl/middleware/test.clj
+++ b/src/cider/nrepl/middleware/test.clj
@@ -171,19 +171,26 @@
   [_m]
   (swap! current-report update-in [:summary :var] inc))
 
+(defn- in-checking-block?
+  "Determine whether the report being generated is for a test.chuck `checking` block."
+  [m]
+  (boolean (:com.gfredericks.test.chuck.clojure-test/testing-contexts m)))
+
 (defn- report-final-status
   [{:keys [type] :as m}]
   (let [ns (ns-name (get m :ns (:testing-ns @current-report)))
-        v (last test/*testing-vars*)]
+        v (last test/*testing-vars*)
+        gen-input (when (in-checking-block? m)
+                    (:gen-input @current-report))]
     (swap! current-report
            #(-> %
                 (update-in [:summary :test] inc)
                 (update-in [:summary type] (fnil inc 0))
+                (assoc :gen-input gen-input)
                 (update-in [:results ns (or (:name (meta v))
                                             fallback-var-name)]
                            (fnil conj [])
-                           (test-result ns v m))
-                (assoc :gen-input nil)))))
+                           (test-result ns v m))))))
 
 (defmethod report :end-test-var
   [{:keys [var-elapsed-time]


alexander-yakushev:
I don't think I ever used `test.chuck` (just the regular `test.check`). Could you please make a small reproducer to make it possible to validate the changes? A cider-nrepl test case would be nice too, but we can start with a reproducer and see if a proper test can be written.

bbatsov:
@zjroth ping :-)

zjroth:
Thanks for the response, @alexander-yakushev, and for the ping, @bbatsov.  Apologies for the delay in this unfortunately busy season for me.

> Could you please make a small reproducer to make it possible to validate the changes? A cider-nrepl test case would be nice too, but we can start with a reproducer and see if a proper test can be written.

I'm not familiar with the term "reproducer" as used here, but it seems to be something like a minimal example of code for which the issue arises.  Is that correct?

zjroth:
As an example of how this situation appears to me, consider the following test:
```clojure
;; [clojure.test.check.generators :as gen]
;; [com.gfredericks.test.chuck.clojure-test :as chuck :refer [checking]]
(deftest show-input-in-test-chuck-checking-block
  (checking "Generated value shown in CIDER test report (only first test fails)"
    {:seed 1}
    [x (gen/choose 1 10)]

    (is (even? x))
    (is true))

  (checking "Generated value not shown in CIDER test report (first test succeeds but subsequent test fails)"
    {:seed 3}
    [x (gen/choose 1 10)]

    (is true)
    (is (even? x))))
```

For me, this produces the following CIDER test report results:
```
Fail in show-input-in-test-chuck-checking-block
Generated value shown in CIDER test report (only first test fails)

expected: (even? x)
  actual: (not (even? 3))          
   input: [{x 3}]

Fail in show-input-in-test-chuck-checking-block
Generated value not shown in CIDER test report (first test succeeds but subsequent test fails)

expected: (even? x)          
  actual: (not (even? 7))      
```
Note the missing `input` from the second failure above.  (Note also that the value of `7` shows because `x` is present in the test condition, which may not always be the case.)

With this PR, the test results change to the following:
```
Fail in show-input-in-test-chuck-checking-block
Generated value shown in CIDER test report (only first test fails)

expected: (even? x)
  actual: (not (even? 3))          
   input: [{x 3}]

Fail in show-input-in-test-chuck-checking-block
Generated value not shown in CIDER test report (first test succeeds but subsequent test fails)

expected: (even? x)          
  actual: (not (even? 7))          
   input: [{x 7}]
 ```
Note that each failure now contains an `input` value, which is not dependent on whether the test condition happens to contain the generated value/symbol.

zjroth:
Hi, @bbatsov and @alexander-yakushev.  I'm just checking back in on this.  Is my example above sufficient as a reproducer as requested?  (NOTE: I modified the example above slightly to make it clearer—in a somewhat contrived way—that the `expected` and `actual` values do not always convey the dynamically generated input.)

alexander-yakushev:
Hi! Sorry for such a long delay. I've tried it out and can confirm that it works. Thank you!

PR #908, opened 2025-01-01
Bump Orchard

alexander-yakushev:
Bumping Orchard and cleaning up Makefile/CI to be the same as in Orchard (download and repackage JDK sources for each JDK version, etc.).


bbatsov (approved):
The changes look good to me!

PR #911, opened 2025-01-05
[info] Add support for automatically downloading sources JARs

alexander-yakushev:
Sister PRs: https://github.com/clojure-emacs/orchard/pull/310, https://github.com/clojure-emacs/cider/pull/3769.

Not a lot of changes were needed here. Given how things are implemented on the Orchard side, cider-nrepl only needs to set up a correct `*download-sources-jar-fn*` before calling out to Orchard functions that do Java sources discovery.

We maintain a ConcurrentHashMap (logically, a set, but there is no ConcurrentHashSet, and CHM API for this usecase is cleaner than `(atom #{})`) of Maven coordinates (`groupId/artifactId "version"`) of the dependencies we already tried downloading sources once. If it succeeded, we won't call this function again anyway, if it failed, the CHM will prevent trying for the second time. This debouncer only exists in memory which I suppose makes it easier to retry (by restarting) if the sources failed to download because bad Internet or something, not because the sources JAR does not exist in the repository. If this cache was persistent, retrying will be a bit more involved. Anyway, this may change after user feedback; maybe, there would be frequent annoying cases where certain Java classes wouldn't have sources JAR, and it will be annoying that the first `C-c C-d C-d` on such class is slow after each REPL restart.

We bind `*download-sources-jar-fn*` only in `info` op, not the `eldoc` op. Eldoc is less intentional and much more common, so I don't want to make slow and download things. Counter to that, `info` is invoked by `M-.` or `C-c C-d C-d` where the user explicitly requested something, so it makes sense to attempt downloading then. Until the JAR is downloaded, `eldoc` will simply return results obtained via reflection, after that it will pick up the parsed Javadocs.

---

### DIFF
--- a/src/cider/nrepl.clj
+++ b/src/cider/nrepl.clj
@@ -57,11 +57,9 @@
   then had to be disabled for test suite reasons in Orchard 0.15.0 to 0.17.0 / cider-nrepl 0.38.0 to 0.41.0, and now it's restored again)
   Note that this can only be done cider-nrepl side, unlike before when it was done in Orchard itself."
   []
-  ;; 1.- Warmup the overall cache for core Java and Clojure stuff
-  @orchard.java/cache-initializer
-  ;; 2.- Also cache classes that are `:import`ed throughout the project.
-  ;;     The class list is obtained through `ns` form analysis,
-  ;;      so that we don't depend on whether the namespaces have been loaded yet:
+  ;; Cache classes that are `:import`ed throughout the project. The class list
+  ;; is obtained through `ns` form analysis, so that we don't depend on whether
+  ;; the namespaces have been loaded yet:
   (doseq [ns-form (vals (orchard.namespace/project-ns-forms))
           class-sym (orchard.namespace/ns-form-imports ns-form)
           :when (try
--- a/src/cider/nrepl.clj
+++ b/src/cider/nrepl.clj
@@ -27,7 +27,7 @@
   Having an enforced minimum version can help users and maintainers alike diagnose issues more quickly,
   avoiding problematic code paths in our middleware, and in clients like cider.el."
   {:major 1
-   :minor 9})
+   :minor 10})
 
 ;; Make sure we're running a supported Clojure version
 (when (or (< (-> *clojure-version* :major long)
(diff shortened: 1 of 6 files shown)


src/cider/nrepl.clj
```diff
@@ -57,11 +57,9 @@
   then had to be disabled for test suite reasons in Orchard 0.15.0 to 0.17.0 / cider-nrepl 0.38.0 to 0.41.0, and now it's restored again)
   Note that this can only be done cider-nrepl side, unlike before when it was done in Orchard itself."
   []
-  ;; 1.- Warmup the overall cache for core Java and Clojure stuff
-  @orchard.java/cache-initializer
```
alexander-yakushev, line 62: This initializer was removed in latest Orchard.

src/cider/nrepl.clj
```diff
@@ -37,12 +37,11 @@
...
-    (.println System/err (format "cider-nrepl requires a newer Clojure version (found: %s, minimum required: %s). Exiting."
-                                 *clojure-version*
-                                 min-clojure-verion))
-    (finally
-      (System/exit 1))))
+  (let [msg (format "cider-nrepl requires a newer Clojure version (found: %s, minimum required: %s)."
+                    *clojure-version* min-clojure-verion)]
+    (try
+      (.println System/err msg)
+      (finally (throw (ex-info msg {}))))))
```
alexander-yakushev, line 44: I don't like that a library can System/exit, I think it takes too much open itself. Throwing an exception is better because it can be caught (for whichever reason the user has), it can be properly logged, you get a stacktrace to know who required `cider.nrepl` namespace (e.g. this might be a surprise to the user), etc.

bbatsov: Fine by me - that `System/exit` was always controversial. Ideally we'd have at some point a way for the server to notify clients if they are not meeting some runtime requirements, but if cider-nrepl can work to some extent and the users are aware of this, it's probably a good trade-off.

alexander-yakushev: Just to clarify, if user doesn't try to catch the exception around requiring `cider.nrepl`, then the end result will be the same as in System/exit – the server will fail to start. It's just that throwing an exception gives more options to react to it and properly report.

bbatsov: Yeah, yeah - I get this. My point was mainly that the way the code was previously written it would fail on the version check alone, even if it might actually work in practice and might be fine with some users. (working, but not really supported) I think the purpose of the explicit `exit` here was to prevent such scenarios from happening. 

I agree it's better for the clients to decide how to handle an error from a library that they're using.

bbatsov (approved):
Apart from the failing tests the PR looks good to me.

bbatsov:
> We bind *download-sources-jar-fn* only in info op, not the eldoc op. Eldoc is less intentional and much more common, so I don't want to make slow and download things. Counter to that, info is invoked by M-. or C-c C-d C-d where the user explicitly requested something, so it makes sense to attempt downloading then. Until the JAR is downloaded, eldoc will simply return results obtained via reflection, after that it will pick up the parsed Javadocs.

You might want to mention this somewhere in the comments and perhaps later in CIDER's docs, so people are aware this is intentional.

PR #914, opened 2025-02-25
[inspector] Don't show javadocs when inspecting classes

alexander-yakushev:
I propose to remove the feature that displays parsed Javadoc in the inspector buffer when inspecting classes, methods, and fields. It currently looks like this:

image

While the idea of uniting the inspector, doc browser, and stacktrace viewer in a coherent UX is a sound one, the current implementation is not ideal. 

- The javadoc is shown at the very bottom of the inspector and usually takes a lot of scrolling to reach. 
- The free-form text looks quite out of place in the inspector buffer.
- Compared to a `*cider-doc*` buffer, the inspector view lacks jump-to-definiton and online Javadoc buttons.
- This implementation complicates the inspector on both Clojure and Emacs sides.

I have a vision of tying `cider-doc` and the Inspector together through clickable buttons instead. This will be less invasive and will have a better separation of concerns for these instruments. Meanwhile, we can clean up the current implementation.

Removing cache warmup too, both because it's no longer needed by the inspector, and the arguments from #913 also apply here.

- [x] All tests are passing
- [x] Middleware documentation is up to date. Run `lein docs` afterwards, and commit the results.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * [#913](https://github.com/clojure-emacs/cider-nrepl/pull/913): Disable background warmup of `orchard.java` cache.
 * [#913](https://github.com/clojure-emacs/cider-nrepl/pull/913): Enable background warmup of Compliment cache.
+* [#914](https://github.com/clojure-emacs/cider-nrepl/pull/914): Remove javadoc section from the inspector output.
 
 ## 0.52.1 (2025-02-24)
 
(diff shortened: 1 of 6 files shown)


bbatsov:
I had even forgotten we had such a feature. When was this introduced?

bbatsov (approved):
> I have a vision of tying cider-doc and the Inspector together through clickable buttons instead. This will be less invasive and will have a better separation of concerns for these instruments. Meanwhile, we can clean up the current implementation.

I agree with your arguments for the change and I like the proposed vision.

alexander-yakushev:
> I had even forgotten we had such a feature. When was this introduced?

Looks like it happened here: https://github.com/clojure-emacs/cider/blob/master/CHANGELOG.md#1110-2023-11-07

PR #919, opened 2025-03-11
Haystack to Orchard

alexander-yakushev:
Multiple things going on in this PR:

1. Remove analyze-stacktrace ops (the ones that took an exception printed as string, parsed it, and returned as an exception object that can be presented via `*cider-error*`). The idea is that https://github.com/clojure-emacs/orchard/pull/320 will cover the usecases where user wants to jump to source location from a printed stacktrace, and that is more discoverable than a separate quite obscure command.
2. Replace the usage of `haystack.analyzer` with `orchard.stacktrace`. The latter is almost the same as the former, but with improvements and optimizations. See https://github.com/clojure-emacs/orchard/pull/322.
3. I rewrote a few test namespaces along the way to use matcher-combinators to reduce noise and improve error reporting.

I plan to let it hang here for some time while I dogfood the changes.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## master (unreleased)
 
+* Bump `orchard` to [0.31.0](https://github.com/clojure-emacs/orchard/blob/master/CHANGELOG.md#0310-2025-03-14).
+  - Info: recognize printed Java classes/methods and munged Clojure functions in stacktrace outputs.
+  - Add dedicated renderers for exceptions in inspector and debugger.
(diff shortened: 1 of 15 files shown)


bbatsov (approved):
Everything looks good to me. The test improvements are really nice!

I'm guessing no other editor is using directly analyze ops, as they are somewhat tied to how we've been doing things in CIDER, but we should probably check with Calva at least to make sure that's the case.

alexander-yakushev:
Searching through Calva source code shows that nothing uses `analyze-stacktrace` op.

PR #920, opened 2025-03-13
[util] Add util/respond-to

alexander-yakushev:
There is a very common pattern used everywhere in cider-nrepl:

```clj
(transport/send (:transport msg) (response-for msg ...))
```

This PR commemorates this pattern as a function `respond-to` and replaces all matching usages with it. Perhaps, this is something that could live in nREPL itself, but we'll still have to support older versions of nREPL here, so might as well add it here.

### DIFF
--- a/src/cider/nrepl/middleware/debug.clj
+++ b/src/cider/nrepl/middleware/debug.clj
@@ -3,22 +3,20 @@
   {:author "Artur Malabarba"}
   (:require
    [cider.nrepl.middleware.inspect :refer [swap-inspector!]]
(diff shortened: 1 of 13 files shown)


alexander-yakushev:
Includes commits from a neighbouring PR, will merge after that one.

bbatsov:
@alexander-yakushev Really good idea - I love the symmetry between `response-for` (a subject) and `respond-to` (an action). 

I also thing it'd be prudent to include this in nREPL, even if we won't be able to use it directly for a while.

bbatsov (approved):

PR #922, opened 2025-03-25
Bump compliment to 0.7.0

alexander-yakushev:
A bunch of goodies in the new version.

Some test prettyfication as a bonus.

---


alexander-yakushev:
Wanna ship it quick, it is a nice one.

PR #923, opened 2025-03-26
[complete] Make sort-order customizable

alexander-yakushev:
So, I've merged the new Compliment that brought this: https://github.com/alexander-yakushev/compliment/pull/122. The idea is to improve the order of candidates with information from the backend.

However, I couldn't make it work right away. Turns out, company-mode (and Emacs' own complete-at-point) perform their own client-side post-sorting of candidates – which they do by-name). To get rid of it, there is a metadata parameter that can be assigned to completion backend to avoid client-side sorting. I did that, so now the backend results come as is. And they come in like this, for example:

image

Notice the weird order – a `java.` class first, then two `javax.` classes, then a `java.` class again. It is because Compliment's own default sorting method is `:by-length` – shorter candidates first. This looks messy. This is also not what users are used to by default – because by default company-mode and friends performed their own sorting and that sorting is strictly alphabetical.

The proposal here is to change the order that cider-nrepl returns to `:by-name`. This is inline with previous user experience and will allow us to cleanly transition to backend-driven sorting. 

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## master (unreleased)
 
+* [#923](https://github.com/clojure-emacs/cider-nrepl/pull/923): Complete: make sorting order customizable.
+
 ## 0.53.1 (2025-03-26)
 
 * Bump `compliment` to [0.7.0](https://github.com/alexander-yakushev/compliment/blob/master/CHANGELOG.md#070-2025-03-25).
--- a/src/cider/nrepl/middleware/complete.clj
+++ b/src/cider/nrepl/middleware/complete.clj
@@ -45,11 +45,13 @@
    :compliment.sources.local-bindings/local-bindings])
 
 (defn complete
-  [{:keys [ns prefix symbol context extra-metadata enhanced-cljs-completion?] :as msg}]
+  [{:keys [ns prefix symbol context extra-metadata enhanced-cljs-completion? sort-order]
+    :as msg}]
   ;; TODO: Drop legacy symbol param in version 1.0
   (let [prefix (str (or prefix symbol))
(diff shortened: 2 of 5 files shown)


bbatsov:
That sounds fine to me, but I wonder if this change won't affect negatively other client, that might be relying on the sort order from `cider-nrepl`? Perhaps it'd be better to make the sort order a param for the completions request instead?

alexander-yakushev:
I doubt anybody would prefer that by-length sorting from the look of it, but I'm OK keeping the old default and making it configurable.

bbatsov:
Well, I'm also fine with just keeping the PR as suggested and simplifying our lives, as otherwise we'd have to make a similar configuration changes for nREPL. Just sharing a point of concern, without really knowing if the other clients do any sorting of their own.

bbatsov (approved):

src/cider/nrepl/middleware/complete.clj
```diff
@@ -45,11 +45,13 @@
    :compliment.sources.local-bindings/local-bindings])
 
 (defn complete
-  [{:keys [ns prefix symbol context extra-metadata enhanced-cljs-completion?] :as msg}]
+  [{:keys [ns prefix symbol context extra-metadata enhanced-cljs-completion? sort-order]
```
bbatsov, line 48: This should probably be mentioned in the op's docs.

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## master (unreleased)
 
+* [#923](https://github.com/clojure-emacs/cider-nrepl/pull/923): Sort candidates by name by default.
```
bbatsov, line 5: You'll need to update the changelog entry.

PR #925, opened 2025-04-01
Stop shipping Puget with cider-nrepl

alexander-yakushev:
So, here's another attempt to reduce the surface of cider-nrepl dependency. The proposal is based on this completely unscientific test that checks how many people are using non-standard pretty-printers with CIDER. Out of 15 responses, 2 are using Fipp and 1 is using Puget.

The idea in the PR is to revert to clojure.pprint if Puget could not be loaded. We also print a warning to tell the user what's going on. I'd say this is graceful UX degradation as the results are still being pretty-printed, and there is a path for the user to fix this. One flow I'm worried about is the `cider-jack-in` one where user haven't ever set up a custom profile to put extra dependencies there.

There's no rush with this one, I'll be happy to know what you think.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## master (unreleased)
 
+* [#925](https://github.com/clojure-emacs/cider-nrepl/pull/9250): Stop vendoring Puget dependency.
+* [#917](https://github.com/clojure-emacs/cider-nrepl/pull/917): Sort printed maps in test output.
+
 ## 0.53.2 (2025-03-26)
 
 * [#923](https://github.com/clojure-emacs/cider-nrepl/pull/923): Complete: make sorting order customizable.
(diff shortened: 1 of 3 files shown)


bbatsov:
I'm wondering if won't better to just leave some instructions for the users how to define a custom printer and be done with it. (or push for such printers to be added upstream in the respective libraries) The current implementation aimed to simplify the usage of different printers, but I think the dependency trade-off was not worth it. 

But I'm not opposed to your suggestion, as it will definitely be less disruptive.

bbatsov:
Why don't we give `fipp` the same treatment?

alexander-yakushev:
> I'm wondering if won't better to just leave some instructions for the users how to define a custom printer and be done with it. (or push for such printers to be added upstream in the respective libraries) The current implementation aimed to simplify the usage of different printers, but I think the dependency trade-off was not worth it.

History shows that such cross-project integrations aren't implemented often. I think it's OK to keep the integration points on our side. Truth be told, I don't expect new popular pretty-printers to appear anymore, so there won't be much space for new integrations and collaborations. And since we already have the existing integrations and they don't consume much by themselves, we won't gain much by removing them.

> Why don't we give fipp the same treatment?

I decided to start with Puget as I feel like fipp may be more popular as a simple "better pretty-printer" while Puget is a more sophisticated one. There's a higher chance people are using fipp. I think we can remove fipp in the future after we ship the `pp` printer vendored inside Orchard and make it the default fallback printer for everything.

bbatsov:
One more thing - this will require some updates to CIDER's docs when it's merged.

alexander-yakushev:
Alright, let's merge it but don't poke any more pprint stuff for 1.18, see how this works for the users.

PR #928, opened 2025-04-04
[inspect] Add support for :table view-mode

alexander-yakushev:
Sister PR: https://github.com/clojure-emacs/orchard/pull/331

Not a lot to be done here. We already have `inspector-toggle-view-mode` that does the heavy lifting. The only meaningful thing here is checking if the current object can be rendered as a table.

- If it does, the view mode cycle goes like `normal -> table -> object -> normal`
- Otherwise, it's `normal -> object -> normal`.

---

### DIFF
--- a/src/cider/nrepl/middleware/inspect.clj
+++ b/src/cider/nrepl/middleware/inspect.clj
@@ -77,7 +77,10 @@
     (inspector-response msg (swap-inspector! msg #(inspect/refresh % overrides)))))
 
 (defn- toggle-view-mode [{:keys [view-mode] :as inspector}]
-  (let [toggle-order {:normal :object, :object :normal}
+  ;; The order in which view modes are cycled depends on the inspected object.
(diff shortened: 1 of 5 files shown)


src/cider/nrepl/middleware/inspect.clj
```diff
@@ -77,7 +77,9 @@
     (inspector-response msg (swap-inspector! msg #(inspect/refresh % overrides)))))
 
 (defn- toggle-view-mode [{:keys [view-mode] :as inspector}]
-  (let [toggle-order {:normal :object, :object :normal}
+  (let [toggle-order (if (inspect/supports-table-view-mode? inspector)
```
bbatsov, line 80: I'd add a note here that depends on the underlying data, as I'm not sure it will be obvious from the name of the function alone.

bbatsov (approved):

PR #929, opened 2025-04-08
[profile] Migrate profiling middleware to orchard.profile

alexander-yakushev:
Sister PR: https://github.com/clojure-emacs/orchard/pull/333

- Bumped Orchard
- Removed thunknyc/profile from dependencies
- Dropped several ops from the middleware – things like individual function summaries, tuning max sample count, stuff like that. Only four ops remain: toggle profile for a var, for a whole ns, clear results, and render results via the inspector.
- Renamed the remaining ops to have `cider/` prefix. I briefly tried keeping the semblance of backward compatibility and keep old names too, but since the most important op (summary) is not compatible with its old implementation, it provides no value. GIven that the feature is an obscure one, I don't care much about breakages here.
- Added `:orchard.profile/profiled` as a relevant meta key – we can then outline vars with it in the same way we do with traced vars.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,10 +2,13 @@
 
 ## master (unreleased)
 
+* Bump `orchard` to [0.33.0](https://github.com/clojure-emacs/orchard/blob/master/CHANGELOG.md#0330-2025-04-08).
+* [#929](https://github.com/clojure-emacs/cider-nrepl/pull/929): Migrate profiling middleware to orchard.profile.
+
 ## 0.54.0 (2025-04-05)
 
 * Bump `orchard` to [0.32.1](https://github.com/clojure-emacs/orchard/blob/master/CHANGELOG.md#0320-2025-04-05).
-* [#925](https://github.com/clojure-emacs/cider-nrepl/pull/9250): Stop vendoring Puget dependency.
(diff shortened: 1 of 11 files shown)


bbatsov (approved):
> Renamed the remaining ops to have cider/ prefix. I briefly tried keeping the semblance of backward compatibility and keep old names too, but since the most important op (summary) is not compatible with its old implementation, it provides no value. GIven that the feature is an obscure one, I don't care much about breakages here.

Agreed. I'm pretty sure that was used by CIDER only anyways.

alexander-yakushev:
@bbatsov Are you OK with the `cider/` prefix? I also see there is `cider.clj-reload/*`, so I wonder what's the final idea for the consistent prefixes.

bbatsov:
> @bbatsov Are you OK with the `cider/` prefix? I also see there is `cider.clj-reload/*`, so I wonder what's the final idea for consistent prefixes.

Yeah, the plan is to prefix everything with "cider/". This is the way!

bbatsov:
Here are a bit more details about my plan for adopting a consistent `cider/` op prefix https://github.com/clojure-emacs/cider-nrepl/issues/710 Not a lot of work overall, but somehow I never got to doing it.

PR #930, opened 2025-04-09
[stacktrace] Add support for directly inspecting ex-data

alexander-yakushev:
Changes made to address https://github.com/clojure-emacs/cider/issues/3804.

1. The op `inspect-last-exception` now accepts `ex-data: true` parameter that tells it the user wants to inspect ex-data of the exception directly, not the entire exception. This is used to have separate buttons in `*cider-error*` buffer for the whole exception and for just ex-data.
2. Replaced a separate dynamic variable manually stored in the session with storing the last analyzed exception inside session's metadata – the same approach as used by inspector middleware.

---


bbatsov (approved):
Looks solid to me!

Issue #806, opened 2019-12-03
Track-state functionality grinds CIDER to a halt on a project with many namespaces and heavy :refer :all usage


alexander-yakushev:
We are using CIDER with a really big project (1000+ namespaces). Most of those namespaces have a common "prelude" form that requires and refers many other namespaces. With a project  like this, CIDER consistently slows down the application and the Emacs. We were able to pinpoint the problem to `track-state` middleware, particularly the part that sends the namespaces that were changed since the last time. When the cache is empty, it sends *all* namespaces from cider-nrepl to cider. This actually creates three problems:

1. Clojure side (on cider-nrepl) spins like crazy and spends a lot of time in `cider.nrepl.middleware.track-state/filter-core-and-get-meta` function.

2. The huge result (~50Mb) gets transferred to the client. This takes ages when doing any remote development.

3. Emacs caches the received result and keeps it in memory all the time which makes Emacs GC last a few seconds. Most Emacs operations become ridiculously slow.

We were able to reduce this problem a bit by blacklisting `:doc` and `:arglists` from the data that gets transferred from cider-nrepl to CIDER. Apparently, it doesn't break anything, so it is not obvious why they are there in the first place.

In any case, it looks like this way of tracking changes (transferring huge maps on every change) doesn't scale too well. I could try to suggest a less strenuous way to do this, but I have to know where else the information from track-state is used. So far, I found only that font-locking relies on it.

vemv:
Issue moved to cider-nrepl

alexander-yakushev:
I currently don't have access to the project where this has triggered. But I can't give you the overall shape. So, there are around ~20 "common" namespaces, each having say ~20-30 functions. Then, there is ~1000 "consumer" namespaces, each doing `(require :refer :all)` of every common namespace. Important detail that all those 1000 namespaces are eagerly loaded into the REPL.

I don't quite remember if the slowdown was all the time or just on fresh REPL initialization. I think it was the latter. Once the results for `track-state` were cached, the Clojure side wasn't lagging anymore. However, Emacs side did because of the huge cache and GC issues.

alexander-yakushev:
What I have is the hack that I used to workaround this issue:

```clj
;; Patch track-state middleware to reduce the amount of data transferred.

(let [meta-ns (try-req 'cider.nrepl.middleware.util.meta)
      misc-ns (some-> meta-ns (.lookupAlias 'misc))
[...]

PR #931, opened 2025-04-11
[track-state] Completely redesign track-state middleware

alexander-yakushev:
I can't believe it took me this long to get to this. The current implementation of `track-state` is very far from ideal and I had problems with it for a long time (see #806). In the current large project I'm testing this, every eval op takes ~300-400ms to recalculate the updated state (even if there were no changes to relevant metadata) and this produces ~250-300 MB of garbage. The state recalculation doesn't delay the evaluation result (as the state update happens on a separate thread), but it delays font lock updates which can be annoying. Besides, the allocated garbage produces extra strain if a REPL process is already low on memory. Finally, the cache used by this middleware occupies again ~200 MB of heap, and there is a counterpart cache on the Emacs side which eats memory and causes GC pauses too.

This is a complete redesign. Here are the main points:
- Drop metadata that is not relevant to state tracking. Namely, I removed `:doc` and `:arglists`. This instantly reduced the cache size by a lot.
- The most common relevant metadata for functions is `{:fn "true"}` (not private, not deprecated, not a macro). Reusing this as a literal brings down retained cached size further. This is very significant.
- **New cache design.**
   - For each namespace, we keep a map of `var-symbol -> real-metadata-map`. Because the "relevant metadata map" is always a derivative of a real metadata map, we can skip recomputing the relevant map if the real one hasn't changed, and take the computed map from the cache (from previous run). This map is kept as WeakHashMap to avoid holding onto metadata maps for vars that could have been undefined. The cache WHM is updated if the real metadata of the var has changed since the previous run.
   - For each loaded namespace, we go over its vars and recompute relevant metadata maps. If none of the vars changed its metadata, it means we took all relevant meta maps from cache, and the whole namespace state can remain as is (speeds up the comparison/diff operations later).

The remaining changes are mostly about refactoring and prettyfication. I introduced a couple of dynamic vars to simplify the internal API. Dynvars usually have worse performance, but here it is negligible, and I cache them into local vars where necessary.

Results on that large project:
- ~15ms middleware execution time (it becomes ~20-40ms end to end – time between "done" and "state" messages, but it still feels instant).
- ~3-4 MB allocated.

The public API hasn't changed. There are minor tweaks to be done on CIDER side later, but overall this is backwards-compatible.

---

### DIFF
--- a/src/cider/nrepl/middleware/track_state.clj
+++ b/src/cider/nrepl/middleware/track_state.clj
@@ -1,13 +1,13 @@
 (ns cider.nrepl.middleware.track-state
   "State tracker for client sessions."
-  {:author "Artur Malabarba"}
+  {:author "Artur Malabarba, Oleksandr Yakushev"}
   (:require
    [cider.nrepl.middleware :as mw]
    [cider.nrepl.middleware.util :as util]
    [cider.nrepl.middleware.util.cljs :as cljs]
    [cider.nrepl.middleware.util.meta :as um]
    [clojure.java.io :as io]
-   [clojure.string :as string]
+   [clojure.string :as str]
    [clojure.tools.namespace.find :as ns-find]
    [nrepl.misc :refer [response-for]]
    [nrepl.transport :as transport]
@@ -16,16 +16,13 @@
    [orchard.java.classpath :as cp]
    [orchard.misc :as misc])
   (:import
-   (clojure.lang MultiFn Namespace)
+   (clojure.lang MultiFn Namespace Var)
    (java.io File)
-   (java.net SocketException)
+   (java.util WeakHashMap)
    (java.util.jar JarFile)
    (nrepl.transport Transport)))
 
-(def clojure-core (try (find-ns 'clojure.core)
-                       (catch Exception _e nil)))
-
-;;; Auxiliary
+;;; Shared part
 
 (defn- inferrable-indent?
   "Does metadata map `m` lack, need a `:style/indent` value, and is a suitable candidate for it?"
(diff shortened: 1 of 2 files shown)


alexander-yakushev:
I need to fix the tests, but this can be already reviewed.

bbatsov (approved):
At a glance everything seems solid to me. Amazing work!!! 🔥

src/cider/nrepl/middleware/track_state.clj
```diff
@@ -52,34 +50,106 @@
...
-                              [sym (um/relevant-meta the-meta)]))))))))
+;;; Clojure part
+
+(def ^:private clojure-core (try (find-ns 'clojure.core)
+                                 (catch Exception _e nil)))
+
+;; A "real metadata cache" is a map {ns-symbol WHM{var-symbol meta}} of
+;; actual (not filtered) var metadata. We keep this to know when to recompute
+;; the filtered metadata for a var (if the real meta hasn't changed, no need to
+;; recompute). WHM is used tp avoid unnecessarily holding onto the metadata if
```
bbatsov, line 65: tp -> to

src/cider/nrepl/middleware/track_state.clj
```diff
@@ -52,34 +50,106 @@
...
+        ;; WHM is not thread-safe but we should only access this from a single
+        ;; (session) thread.
+        cached-meta (.get real-metadata-ns-cache var-name)
+        current-meta (meta the-var)]
+    (when-not (identical? cached-meta current-meta)
+      (.put real-metadata-ns-cache var-name current-meta)
+      current-meta)))
+
+;; Note that we aggressively cut down unnecessary keys in this mw as it triggers
+;; on EACH evaluation, and the middleware retains (caches) this data for ALL
```
bbatsov, line 82: ALL loaded project namespaces :-)

PR #932, opened 2025-04-11
Add support for pretty printing values in the inspector.

r0man:
Hello,

this PR adds a new pretty view mode that uses the pretty printer from https://github.com/eerohele/pp via this PR in orchard https://github.com/clojure-emacs/orchard/pull/335. I believe there are no changes needed on the Emacs side.

TODO:
- Tests are failing because I need a new orchard version on ci, but how?

Open questions:
- Where should we put the pretty mode in the toggle list? I would put it after normal mode, 2nd position. Right now it is last, hard to reach In my opinion.
- When drilling down the inspector always switches to normal mode. If I'm in pretty mode I would like to stay there, and I would argue for other modes as well, if possible. right now you have to toggle a ton, to get that behaviour. 
Wdyt?

---

- [x] All tests are passing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 ## 0.55.1 (2025-04-14)
 
 * [#931](https://github.com/clojure-emacs/cider-nrepl/pull/931): Redesign and optimize track-state middleware.
+* [#932](https://github.com/clojure-emacs/cider-nrepl/pull/932) Add support for pretty printing values in the inspector.
 
 ## 0.55.0 (2025-04-10)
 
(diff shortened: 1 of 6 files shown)


alexander-yakushev:
> Tests are failing because I need a new orchard version on ci, but how?

We don't have a solution for that right now. So it's usually "merge and publish Orchard first, then merge cider-nrepl". But you can test locally by installing a local Orchard build and specifying that as a dependency for cider-nrepl.

r0man:
Ok, that's what I remembered as well. I'm using the approach you suggested, tests are passing here, and I also fixed the issue I had with the linter. Thanks!

alexander-yakushev (approved):
All good!

alexander-yakushev:
@r0man Do you want to change anything else or we are good to merge?

PR #933, opened 2025-04-13
Add inspect-print-current-value op

r0man:
This PR adds the `inspect-print-current-value` op. It prints the current value of the inspector with the user provided `::print/print-fn` or a default one.

It is used by CIDER in the `cider-inspector-print-current-value` command, bound to "P" in the inspector, to print it's current value. https://github.com/clojure-emacs/cider/pull/3810

- [x] All tests are mostly passing (there seem to be some unrelated flaky tests)

### DIFF
--- a/src/cider/nrepl/middleware/inspect.clj
+++ b/src/cider/nrepl/middleware/inspect.clj
@@ -1,8 +1,11 @@
 (ns cider.nrepl.middleware.inspect
   (:require
+   [cider.nrepl.middleware.util :refer [respond-to]]
    [cider.nrepl.middleware.util.cljs :as cljs]
    [cider.nrepl.middleware.util.error-handling
     :refer [eval-interceptor-transport with-safe-transport]]
+   [cider.nrepl.pprint :as pprint]
+   [nrepl.middleware.print :as print]
    [nrepl.misc :refer [response-for]]
    [nrepl.transport :as transport]
    [orchard.inspect :as inspect]))
@@ -107,6 +110,15 @@
 (defn def-current-value [msg]
   (inspector-response msg (swap-inspector! msg inspect/def-current-value (symbol (:ns msg)) (:var-name msg))))
 
+(defn print-current-value-reply [{:keys [::print/print-fn session] :as msg}]
+  (let [inspector (-> session meta ::inspector)]
+    (with-open [writer (print/replying-PrintWriter :value msg msg)]
+      (binding [*print-length* (or *print-length* 100)
+                *print-level* (or *print-level* 20)]
+        ((or print-fn pprint/pprint) (:value inspector) writer)
+        (.flush writer)))
+    (respond-to msg :status :done)))
+
 (defn tap-current-value [msg]
   (inspector-response msg (swap-inspector! msg inspect/tap-current-value)))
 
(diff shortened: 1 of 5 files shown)


alexander-yakushev:
I can't say I immediately understand the usecase, to be honest.

1. Inspector is already a "better pretty-print", so there is less value in printing the value to stdout if you are looking at it in the inspector.
2. We'll have a separate pretty-printing view that can be switched on inside the inspector.
3. After pretty-printing the value to stdout, if the user only had the inspector window open, they will have to close it or explicitly navigate to the REPL buffer to see what's had been pprinted. Can this be convenient?

r0man:
@alexander-yakushev Have you seen the screenshot and my explanation here: https://github.com/clojure-emacs/cider/pull/3810 ?

It is not printing to stdout, it opens the value in a popup buffer, like `cider-pprint-eval-last-sexp`. I also mentioned there why I sometimes prefer to look at the pretty printed version of it, vs the one we show in the inspector today.

That said, let's see where we land with the pretty mode before continuing with this.

r0man:
The jdk8 test is failing because of this:
```
ERROR in (continue-non-stop) (debug_integration_test.clj:293)
Uncaught exception, not in assertion.
expected: nil
  actual: clojure.lang.ExceptionInfo: Unexpected timeout waiting for nREPL response. Expected message: {:status ["done"]}
```
I think it's a flaky test, that I have seen failing before.

bbatsov:
One question on my mind is whether we need a dedicate op to print the inspector value. I was thinking it might also makes sense to have an op that fetches the current inspector value and for it to be printed separately. I guess the current suggestion would be more efficient, but it seems oddly specific to me.

r0man:
Hi @bbatsov, I think we need one. Initially I had an op that just returned the rendered value as a string. But I changed it because it caused slowness on large data structures and NREPL timeouts.

alexander-yakushev:
@bbatsov I understand what you are saying. But what's "fetching the current inspector value"? You can't "return" a value from nrepl server to the client, it has to be printed into a string anyway.

src/cider/nrepl/middleware/inspect.clj
```diff
@@ -107,16 +110,30 @@
...
 (defn tap-indexed [msg]
   (inspector-response msg (swap-inspector! msg inspect/tap-indexed (:idx msg))))
 
 (defn handle-inspect [handler {:keys [op inspect] :as msg}]
-  (if (and (= op "eval") inspect)
+  (cond
+    (and (= op "eval") inspect)
     (handle-eval-inspect handler msg)
 
+    (and (= op "inspect-print-current-value"))
```
r0man, line 133: @alexander-yakushev I moved this out of `with-safe-transport` because I saw an exception in the `*nrepl-messages-buffer*`, which I think was caused by the new op sending mutiple lines vs the ops in `with-safe-transport` returing a response value. 

I didn't notice this exception before, because from the user's perspective both seemed to work.

With this change I don't see the exeception anymore. Can you double check that this ok?

alexander-yakushev: Yeah, I think it makes sense. This op is indeed different in this regard. It's fine to leave this as is, just leave a comment why it stands separately.

src/cider/nrepl/middleware/inspect.clj
```diff
@@ -107,16 +110,33 @@
...
 (defn handle-inspect [handler {:keys [op inspect] :as msg}]
-  (if (and (= op "eval") inspect)
+  (cond
+    (and (= op "eval") inspect)
     (handle-eval-inspect handler msg)
 
+    ;; This is outside of `with-safe-transport` because it streams the pretty
+    ;; printed result to the client in multiple messages. It does NOT return any
+    ;; value, which is expected by `with-safe-transport`.
+    (= op "inspect-print-current-value")
```
bbatsov, line 136: @alexander-yakushev Should we prefix this with `cider/` from the beginning or leave this to for another day? (e.g. when we add aliases for the all the existing ops)

alexander-yakushev: I don't mind; might as well add the prefix now indeed.

r0man: I would keep it consistent and do it once we decide to rename all. But I can also do this now (we are talking about this op only right?) if you prefer.

bbatsov: Okay, let's do this later.

PR #938, opened 2025-04-28
[ci] Speed up testing

alexander-yakushev:
A couple of CI improvements. The most significant one is to only run quick test for the complete JDK*Clojure matrix. Quick tests are those that don't andersonize dependencies. Only run the full test with inline dependencies and the smoketest for the latest Clojure and for JDK8 and the latest JDK.


.circleci/config.yml
```diff
@@ -93,17 +93,27 @@ jobs:
...
       - with_cache:
-          cache_version: "lint_v3_1.12"
-          steps: << parameters.steps >>
+          cache_version: "lint_v4"
+          steps:
+            - run:
+                name: Running cljfmt
+                command: |
+                  make cljfmt
+            # TODO: use `make kondo` instead once there's a fix for https://github.com/clj-kondo/clj-kondo/issues/2189
```
bbatsov, line 108: Is this still relevant?

.circleci/config.yml
```diff
@@ -130,7 +130,30 @@ jobs:
 
   test:
     description: |
-      Run tests against given version of JDK and Clojure
+      Run quick tests against given version of JDK and Clojure without andersonizing dependencies
```
bbatsov, line 133: andersonizing -> inlining :-)

Let's not get too specific with the particular tool.

.circleci/config.yml
alexander-yakushev: Good catch, removed.

alexander-yakushev:
Eastwood occasionally triggers on its own, so it's worth at least something. But yeah, it can evaluated in the future.

alexander-yakushev:
I'm delighted to see most jobs completing in 1-2 minutes instead of 4-5. Circle will throttle us less too.

Issue #936, opened 2025-04-24
track-state may spin forever consuming 100% CPU


alexander-yakushev:
From https://clojurians.slack.com/archives/C0617A8PQ/p1745491033246169

> we upgraded cider-nrepl to 0.55.1 and today we had the case, where 11 cpus ran at 100% for around 45 minutes until we killed the process. I ran jstack and it looks like it has to do with latest changes to track-state

```
"clojure-agent-send-off-pool-7" #88 [3077360] prio=5 os_prio=0 cpu=2571688.20ms elapsed=3094.35s allocated=663M defined_classes=6 tid=0x00007fc44400a090 nid=30
0007fc4275fc000]
   java.lang.Thread.State: RUNNABLE
        at java.util.WeakHashMap.get(java.base@21.0.3/WeakHashMap.java:415)
        at cider.nrepl.middleware.track_state$get_metadata_if_changed_QMARK_.invokeStatic(track_state.clj:74)
        at cider.nrepl.middleware.track_state$get_metadata_if_changed_QMARK_.invoke(track_state.clj:69)
        at cider.nrepl.middleware.track_state$compute_var_meta.invokeStatic(track_state.clj:96)
        at cider.nrepl.middleware.track_state$compute_var_meta.invoke(track_state.clj:89)
        at cider.nrepl.middleware.track_state$compute_var_metas_for_namespace$fn__96498.invoke(track_state.clj:128)
        at clojure.lang.PersistentHashMap$NodeSeq.kvreduce(PersistentHashMap.java:1309)
        at clojure.lang.PersistentHashMap$BitmapIndexedNode.kvreduce(PersistentHashMap.java:804)
        at clojure.lang.PersistentHashMap$NodeSeq.kvreduce(PersistentHashMap.java:1314)
        at clojure.lang.PersistentHashMap$BitmapIndexedNode.kvreduce(PersistentHashMap.java:804)
        at clojure.lang.PersistentHashMap$ArrayNode.kvreduce(PersistentHashMap.java:468)
        at clojure.lang.PersistentHashMap.kvreduce(PersistentHashMap.java:238)
        at clojure.core$fn__8555.invokeStatic(core.clj:6987)
        at clojure.core$fn__8555.invoke(core.clj:6967)
        at clojure.core.protocols$fn__8283$G__8278__8292.invoke(protocols.clj:174)
        at clojure.core$reduce_kv.invokeStatic(core.clj:6998)
[...]

PR #939, opened 2025-04-29
[track-state] Synchronize access to WeakHashMap cache

alexander-yakushev:
Addresses #936.

See https://dowbecki.com/WeakHashMap-is-not-thread-safe/.

---


bbatsov (approved):
Looks good to me.

PR #941, opened 2025-05-01
Stop shipping fipp with cider-nrepl

alexander-yakushev:
In the same way as Puget and zprint.

Additionally, default all three to `orchard.pp`, our new pretty-printing implementation that is faster than `clojure.pprint` and is also more consistent with the way we do printing in other parts of CIDER (inspector, debugger).

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## master (unreleased)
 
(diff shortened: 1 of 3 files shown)


bbatsov (approved):
Looks good to me, but I guess we should document this somewhere. And mention it in the changelog as well.

alexander-yakushev:
I will describe it in CIDER docs, it has a section about (pretty-)?printing.

PR #943, opened 2025-05-20
[debug] Reduce instrumentation footprint 

alexander-yakushev:
We care about instrumentation size because when instrumenting large functions, we run into the JVM method size limit (65KB). This PR adds two improvements:

- Inject `DebugSupport/doBreak` instead of `c.n.m.debug/break`. Invoking a method requires less bytecode than calling a function. A method also has overrides for common primitives.
- Generate a more compact code for constructing a locals map.

---


bbatsov (approved):
The changes look good to me.

Issue #946, opened 2025-09-15
Reader tags like #dbg aren't respected when loading an entire file


jpmonettas:
When you eval a single expression containing a `#dbg` tag then it gets instrumented as expected and you can use the debugger on it. But if instead of evaluating a single expression at a time you load the entire  file the functions containing `#dbg` don't get instrumented.

jpmonettas:
The issue seams to be in how `cider.nrepl.middleware.debug/maybe-debug` works.
It reads the code with a fake reader to determine if there is a #dbg  tag.
In the single form eval, the code is `#dbg (defn some-code [...]...)` which makes has-debug? true. But in the load file case the code is :
```
(try (clojure.lang.Compiler/load
      (java.io.StringReader.
       ((clojure.core/deref (clojure.core/deref (var nrepl.middleware.load-file/file-contents)))
        (quote ["... CODE_FILE_PATH ..." G__23022]))) "... CODE_FILE_PATH ..." "... CODE_FILE_PATH ...")
     (finally (clojure.core/swap! (clojure.core/deref (var nrepl.middleware.load-file/file-contents)) clojure.core/dissoc (quote ["... CODE_FILE_PATH ..." G__23022]))))
```
which makes has-debug? false because that code doesn't have a `#dbg` tag.

I'm attaching two FlowStorm screenshots showing these two cases : 

<img width="1811" height="507" alt="Image" src="https://github.com/user-attachments/assets/a131cc8b-f7e1-441c-a380-2613c7c3196b" />
<img width="1368" height="501" alt="Image" src="https://github.com/user-attachments/assets/258fcefa-9075-4e60-8960-8734bc164313" />

jpmonettas:
Had some time so adding a little bit more on this.

For a normal middleware stack we have : 

```clojure
(:stack @nrepl.middleware.dynamic-loader/*state*)

(#'nrepl.middleware/wrap-describe
 #'cider.nrepl/wrap-profile
 #'cider.nrepl/wrap-classpath
 #'cider.nrepl/wrap-version
 #'cider.nrepl/wrap-trace
 #'cider.nrepl/wrap-log
 #'cider.nrepl/wrap-slurp
 #'nrepl.middleware.interruptible-eval/interruptible-eval
 #'nrepl.middleware.session/add-stdin
 #'cider.nrepl/wrap-refresh
 #'cider.nrepl/wrap-stacktrace
 #'refactor-nrepl.middleware/wrap-refactor
[...]

PR #951, opened 2025-09-23
[debug] Correctly process #dbg during load-file

alexander-yakushev:
This is the tock to https://github.com/nrepl/nrepl/pull/391 's tick. These two PRs are enough to fix https://github.com/clojure-emacs/cider-nrepl/issues/946.

With the updated `load-file` in place, it now becomes possible to correctly instrument functions with `#dbg` and other debug-related reader tags when doing `C-c C-k`. I thought it was gonna be trivial from here but woe is me. First of all, I wanted to drop the old approach of double-reading the code. It's ugly. The forms are now only read once by the `interruptible-eval` loop, just like the gods intended. We pass a custom read function for that because:
- We want to also capture the string underlying the form. This string is necessary for the client (Emacs) to verify that the code buffer didn't change after the instrumentation and the debugger can be shown inline or that a new debug-buffer needs to be created.
- We use the metadata on the form as a vessel to pass extra data to the instrumentation machinery.

Re: capturing the form source code. This proved to be PITA. See `comment-trimming-read+string`.

Of course, the debugger has to continue working with older nREPLs down to 1.0.0. That's why all the older code remained, and I only introduced the new path for the new nREPL.

What else can I say? This crap works now, I sincerely hated doing this, gg go next.

---

### DIFF
--- a/test/clj/cider/nrepl/middleware/inspect_test.clj
+++ b/test/clj/cider/nrepl/middleware/inspect_test.clj
@@ -598,18 +598,19 @@
 
 (deftest inspect-print-current-value-test
   (testing "inspect-print-current-value returns the currently inspected value as a printed string"
-    (is (= [(str/join "\n" ["{:a -1,"
-                            " :bb \"111\","
-                            " :ccc (1),"
-                            " :d"
-                            " ({:a 0, :bb \"000\", :ccc ()}"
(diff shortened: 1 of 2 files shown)


bbatsov (approved):
Looks solid to me!

codeasone:
Your efforts are much appreciated @alexander-yakushev 👏🏻

PR #968, opened 2026-03-05
Remove MrAnderson dependency inlining

bbatsov:
I've been frustrated with MrAnderson for a while now, and I think the time has come to just drop it.

It served us well over the years, but it adds a lot of complexity for diminishing returns — a SNAPSHOT plugin dependency, dual CI test matrices, runtime workarounds (dynamic keyword/symbol construction to sidestep namespace rewriting), build hacks (manually deleting files post-inlining), and conditional env vars for tooling compatibility. The three remaining inlined deps (cljfmt, tools.namespace, tools.reader) are all stable Clojure libraries with strong backwards compatibility, and we already un-inlined orchard, compliment, clj-reload, and suitable without issues.

This PR makes cljfmt, tools.namespace, and tools.reader regular dependencies and removes all the MrAnderson machinery — plugin config, `target/srcdeps` build step, `full-test` CI job, `CIDER_NO_MRANDERSON` env var, runtime keyword construction workarounds, and the associated docs.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -115,12 +115,7 @@ jobs:
                   make eastwood
 
   deploy:
-    # JDK 8 is preferred for deployments (https://github.com/benedekfazekas/mranderson/issues/57)
     executor: jdk8
-    environment:
-      # Parallelism is disabled, at least for now
-      # (https://github.com/benedekfazekas/mranderson/issues/56)
-      LEIN_JVM_OPTS: -Dmranderson.internal.no-parallelism=true
     steps:
       - checkout
       - run:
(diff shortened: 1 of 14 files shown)


benedekfazekas:
need to think about this more, but maybe tools is a way forward to separate cider-nrepl and project deps: https://github.com/clojure/clojure/blob/master/changes.md#22-invoke-tool-functions-out-of-process and https://clojure.org/reference/clojure_cli#tools

as this is a clojure 1.12 feature mranderson could be phased out gradually

bbatsov:
Closing this one. I've since become a co-maintainer of MrAnderson and managed to fix the issues that were troubling me with it, so dropping it is no longer necessary.

PR #969, opened 2026-03-10
Fix spurious nREPL warnings about missing piggieback

bbatsov:
nREPL 1.6+ warns when middleware listed in `:requires` isn't present in the middleware stack. Seven cider-nrepl middleware declared piggieback as a hard `:requires` dependency via `requires-piggieback`, triggering this warning in pure Clojure projects where piggieback isn't loaded.

This deprecates `requires-piggieback` (now a no-op) and removes all internal call sites. The three middleware that genuinely need piggieback ordering (`wrap-inspect`, `wrap-out`, `wrap-tracker`) already use `expects-piggieback` and are unaffected.

Related: nrepl/nrepl#424

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 * [#959](https://github.com/clojure-emacs/cider-nrepl/pull/959): Stop shading Compliment and clj-reload.
 * [#710](https://github.com/clojure-emacs/cider-nrepl/issues/710): Make namespaced ops (e.g., `cider/complete`) the default and keep the old names as deprecated aliases.
 * [#965](https://github.com/clojure-emacs/cider-nrepl/issues/965): Stop losing quotes in matcher-combinators test output.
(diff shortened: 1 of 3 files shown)


alexander-yakushev:
It is unclear to me how "piggieback's own descriptor handles its ordering in the middleware stack". I see that it `:expects #{"eval" "load-file"}` but does nothing for middleware like `wrap-complete` and others. Could you please explain your reasoning?

bbatsov:
I was referring to https://github.com/nrepl/piggieback/blob/master/src/cider/piggieback.clj#L29 indeed. 

In general things like completion just use `grab-cljs-env` directly, so it didn't seem to me that the ordering matters much to them. And the tests still pass, which is an indication that that's a correct assumption. 😂

bbatsov:
See also https://github.com/clojure-emacs/cider-nrepl/blob/master/src/cider/nrepl/middleware/complete.clj#L56 for an example of what I meant.

bbatsov:
Closing in favor of #971.

PR #971, opened 2026-03-10
[middleware] Refactor how piggieback dependencies are established

alexander-yakushev:
This PR achieves several things:

### 1

Deprecate `requires-piggieback` and `expects-piggieback`. Instead, apply a much more pinpointed function `maybe-add-piggieback` directly to require/expect sets.

### 2 

Since nREPL 1.6 now complains that the require/middleware sets refer to middleware not being explicitly included (but still begrudgingly includes it), it is now the job for cider-nrepl to add Piggieback to the middleware list if it is present on the classpath. Is it really the resposibility of cider-nrepl? No. But I think too many setups have this in their deps.edn:

    --middleware "[cider.nrepl/cider-middleware]"

and not

    --middleware "[cider.nrepl/cider-middleware, cider.piggieback/wrap-cljs-repl]"

So, in order for their setups to continue to "just work" without warnings, somebody has to be putting Piggieback onto the mw list.

### 3

Remove mentions and support for `piggieback.core` namespace that never came up to be, and will unlikely ever.

---

- [x] You've updated the [changelog](../blob/master/CHANGELOG.md) (if adding/changing user-visible functionality)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 * [#710](https://github.com/clojure-emacs/cider-nrepl/issues/710): Make namespaced ops (e.g., `cider/complete`) the default and keep the old names as deprecated aliases.
 * [#965](https://github.com/clojure-emacs/cider-nrepl/issues/965): Stop losing quotes in matcher-combinators test output.
 * [#972](https://github.com/clojure-emacs/cider-nrepl/issues/972): Stop relying on tools.reader functions in our code.
+* [#971](https://github.com/clojure-emacs/cider-nrepl/issues/971): Middleware: refactor how piggieback dependencies are established.
 
 ## 0.58.0 (2025-10-16)
 
--- a/project.clj
+++ b/project.clj
@@ -137,7 +137,8 @@
                                                               cider.nrepl/wrap-tracker
                                                               cider.nrepl/wrap-undef
                                                               cider.nrepl/wrap-version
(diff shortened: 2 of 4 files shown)


bbatsov:
> Since nREPL 1.6 now complains that the require/middleware sets refer to middleware not being explicitly included (but still begrudgingly includes it), it is now the job for cider-nrepl to add Piggieback to the middleware list if it is present on the classpath. Is it really the resposibility of cider-nrepl?

I strongly feel that's not cider-nrepl's responsibility - CIDER adds the piggieback manually, and I think that's the expectation for every cider-nrepl user/client. 

For me the optimal solution here would be one that doesn't make any assumptions about what someone would want to do in the presence of piggieback on the classpath and everything is just explicit (and as simple as it can be).

bbatsov:
> Remove mentions and support for piggieback.core namespace that never came up to be, and will unlikely ever.

I had long forgotten I intended to dual-publish piggieback with the old and new coordinates. Yeah, that is certainly not happening . 😂

bbatsov:
So, do we agree to remove the implicit addition of the piggieback middleware (when piggieback is on the classpath) and call it a day?

alexander-yakushev:
@bbatsov Yeah, I relieved cider-nrepl of the duty of bringing Piggieback middleware to the server. The PR is now mergeable as is.

bbatsov (approved):

PR #972, opened 2026-03-10
Stop using tools.reader functions

alexander-yakushev:
It looks to me like we don't use `tools.reader` enough to justify keeping it as a dependency. Regular Clojure's `read` and `clojure.edn` should be sufficient.

---

- [x] You've updated the [changelog](../blob/master/CHANGELOG.md) (if adding/changing user-visible functionality)


alexander-yakushev:
Yeah, this idiot forgot that tools.reader also comes as a transitive dependency of cljfmt. Can't get rid of that one.

alexander-yakushev:
Update: even if we can't drop tools.reader yet, it makes sense to stop depending on its functions directly.

bbatsov (approved):

PR #977, opened 2026-04-19
Fixed typo: min-clojure-verion

metayan:
All occurrences of `min-clojure-verion` changed to `min-clojure-version` in `nrepl.clj`.
Only used internally in `nrepl.clj`, so not adding to Changelog.


alexander-yakushev:
Thanks!

PR #982, opened 2026-06-22
Add trace-subscribe/unsubscribe streaming ops

bbatsov:
Step 2 of giving CIDER a dedicated trace buffer (Tier 2 of the enlighten/tracing audit).

Trace output only ever went to `*out*` (the REPL). This adds a streaming subscription, mirroring the log middleware's consumer pattern:

- `cider/trace-subscribe` - a held-open request that registers an `orchard.trace` listener forwarding a structured event (`id`, `phase`, `name`, `depth`, `args`/`value`) for each call and return, and routes trace output to listeners instead of the REPL.
- `cider/trace-unsubscribe` - removes the listener and restores REPL output once the last subscriber is gone.

This lets a client (CIDER, next) render trace output in a navigable buffer.

Depends on the new `orchard.trace` listener API (clojure-emacs/orchard#400). Pinned to `0.43.0-alpha1` for now; should be bumped to `0.43.0` once that's released, before cider-nrepl's next release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## master (unreleased)
 
+* [#982](https://github.com/clojure-emacs/cider-nrepl/pull/982): Add the `cider/trace-subscribe` and `cider/trace-unsubscribe` ops, streaming a structured event for every traced call and return so clients can render trace output in a dedicated buffer instead of the REPL (requires `orchard` 0.43.0).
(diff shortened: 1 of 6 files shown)


alexander-yakushev:
Does anybody need this? Or is CIDER going to use this feature directly?

bbatsov:
It's for CIDER only at this point. It was something I had on my mind for a while and recently I've decided to experiment with some backlogged ideas I had for better macroexpansion, tracing and enlighten for the next CIDER release.

Probably nobody really needs it, but it seemed like an interesting idea to me as tracing can really clutter the REPL.

PR #1029, opened 2026-07-10
Fix a cluster of debugger bugs

### LINKED ISSUE #750: cider-debug-defn-at-point on a defn with a big body results in exception
## Expected behavior
In Emacs `cider-debug-defun-at-point` on [this defn](https://github.com/clj-kondo/clj-kondo/blob/9195cc0f0761a2b673816f5296a95380917ff05b/src/clj_kondo/impl/linters.clj#L202) results in defn marked as being debugged.

## Actual behavior
Results in `*cider-error*` buffer with:
```
  Show: Project-Only All 
  Hide: Clojure Java REPL Tooling Duplicates  (0 frames hidden)

1. Unhandled java.lang.IllegalArgumentException
   No implementation of method: :send of protocol:
   #'nrepl.transport/Transport found for class: nil

          core_deftype.clj:  583  clojure.core/-cache-protocol-fn
          core_deftype.clj:  575  clojure.core/-cache-protocol-fn
             transport.clj:   23  nrepl.transport/eval631/fn/G
                 nrepl.clj:   18  cider.nrepl.middleware.util.nrepl/notify-client
                 nrepl.clj:    8  cider.nrepl.middleware.util.nrepl/notify-client
                 nrepl.clj:   15  cider.nrepl.middleware.util.nrepl/notify-client
                 nrepl.clj:    8  cider.nrepl.middleware.util.nrepl/notify-client
                 debug.clj:  603  cider.nrepl.middleware.debug/instrument-and-eval
                 debug.clj:  595  cider.nrepl.middleware.debug/instrument-and-eval
                  Var.java:  381  clojure.lang.Var/invoke
    interruptible_eval.clj:   87  nrepl.middleware.interruptible-eval/evaluate/fn/fn
                  AFn.java:  152  clojure.lang.AFn/applyToHelper
                  AFn.java:  144  clojure.lang.AFn/applyTo
                  core.clj:  657  clojure.core/apply
                  core.clj: 1965  clojure.core/with-bindings*
                  core.clj: 1965  clojure.core/with-bindings*
               RestFn.java:  425  clojure.lang.RestFn/invoke
    interruptible_eval.clj:   87  nrepl.middleware.interruptible-eval/evaluate/fn
                  main.clj:  243  clojure.main/repl/read-eval-print/fn
                  main.clj:  243  clojure.main/repl/read-eval-print
                  main.clj:  261  clojure.main/repl/fn
                  main.clj:  261  clojure.main/repl
                  main.clj:  177  clojure.main/repl
               RestFn.java: 1523  clojure.lang.RestFn/invoke
    interruptible_eval.clj:   84  nrepl.middleware.interruptible-eval/evaluate
    interruptible_eval.clj:   56  nrepl.middleware.interruptible-eval/evaluate
    interruptible_eval.clj:  152  nrepl.middleware.interruptible-eval/interruptible-eval/fn/fn
                  AFn.java:   22  clojure.lang.AFn/run
               session.clj:  218  nrepl.middleware.session/session-exec/main-loop/fn
               session.clj:  217  nrepl.middleware.session/session-exec/main-loop
                  AFn.java:   22  clojure.lang.AFn/run
               Thread.java:  748  java.lang.Thread/run
```

## Please note
I have narrowed down the cause of the exception, narrowed-down steps are in _a comment_ below, following them will be of less hussle.

Have the same defn locally.
E.g., by having clj-kondo repo locally and checkouting to the exact commit (latest atm):
```
git clone https://github.com/clj-kondo/clj-kondo.git
cd clj-kondo
git checkout 9195cc0f0761a2b673816f5296a95380917ff05b
```

Launch cider-nrepl 0.28.3 (latest atm):
```
clj -A:test -Sdeps '{:deps {cider/cider-nrepl {:mvn/version "0.28.3"} }}' -m nrepl.cmdline --middleware "[cider.nrepl/cider-middleware]"
```

Open Emacs and connect to the nrepl.
E.g., via `cider-connect-clj`.
In my case, cider informs of such environment:
```
;; CIDER 1.3.0 (Ukraine), nREPL 0.9.0
;; Clojure 1.9.0, Java 1.8.0_161
```

Navigate to the defn.
E.g., via `project-find-regexp RET defn lint-var-usage`.
E.g., via `find-file` at `clj-kondo/src/clj_kondo/impl/linters.clj`, line 202.

Put cursor at the defn symbol, `cider-debug-defun-at-point`.
`*cider-error*` is expected to pop up.

### cider-nrepl version

0.28.3

### cider version

1.3.0 (Ukraine)

### nrepl version

0.9.0

### Clojure version

1.9.0

### Java version

1.8.0_161

### Operating system

Guix, linux kernel. Based on [this config](https://github.com/abcdw/nonrde/blob/main/ixy.scm#L18) with some tweaks.

bbatsov:
A pass over the debugger bug cluster - five independent bugs, one focused commit each.

- **#1016 / #1017**: `#dbg`/`#light` on a bare collection literal that closes over a local (e.g. `#dbg [x]`) threw `Unable to resolve symbol: STATE__`. The wrapper that binds the `STATE__` anaphor was skipped for plain collection literals, so their interior breakpoints referenced an unbound var. `#dbg`/`#dbgexn` now force the top-level wrapper (matching `#break`), and `instrument-and-eval` always establishes the bindings (which also covers the `#light` reader path).
- **#1020**: interrupting a test that isn't wrapped in `is` (e.g. `(while true)`) crashed the reporter with `no conversion to symbol`. The escaped throwable is now reported as an error.
- **#1018**: aborting a debug session with no session thread used to `NullPointerException`. The behavior fix shipped in 0.62 (moving off `Thread.stop`); this adds the missing regression test.
- **#750**: a form too large to instrument even without locals now falls back to evaluating it uninstrumented (with a warning), instead of surfacing a raw `Method code too large!` compiler error.
- **#1022**: destructuring a fn argument with more than eight explicit key/val pairs threw `Symbol cannot be cast to Keyword`. Maps larger than eight entries were rebuilt as sorted-maps, which breaks `destructure`'s `:as`/`:or` keyword lookups; they're now kept as plain maps.

Every fix has a regression test that fails without it; the base suite, cljs, and lint are green.

Fixes #1016. Fixes #1017. Fixes #1018. Fixes #1020. Fixes #750. Fixes #1022.

---

- [x] The commits are consistent with the contribution guidelines
- [x] Tests cover each change
- [x] Changelog updated
- [x] Middleware documentation up to date (no op descriptor or doc changes here)


alexander-yakushev:
I don't think the #750 fix commit is needed. I'd rather have the exception upfront telling me the function coudn't be instrumented. Debugging a function is an explicit action (either with C-u C-c C-c or with #dbg). If it can't be done, don't do it, don't make the user dig through the warnings to see why it didn't work.

alexander-yakushev:
Let's also delay the #1022 fix for now – there is a change in 1.13 that raises the PAM->PHM transition limit. Let me look how to make this patch account for that.

